PR #16, opened 2013-10-27
[romprefix] Allow to enforce PnP BIOSes to be detected as non-PnP

stefanct:
Add a define NONPNP_FORCE that controls building of the PnP detection
routine. Together with the existing NONPNP_HOOK_INT19 hook this enables
booting iPXE on PnP BIOSes that can not be persuaded to boot iPXE by
BIOS menu configuration alone (e.g. my abit BX6 2.0).

Signed-off-by: Stefan Tauner stefan.tauner@gmx.at

### DIFF
--- a/src/arch/i386/prefix/romprefix.S
+++ b/src/arch/i386/prefix/romprefix.S
@@ -295,6 +295,7 @@ no_pci3:
 	popl	%edx
 	popl	%ebx
 
+#ifndef NONPNP_FORCE
 	/* Check for PnP BIOS.  Although %es:di should point to the
 	 * PnP BIOS signature on entry, some BIOSes fail to do this.
 	 */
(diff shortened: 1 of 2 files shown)


lastthyme (approved):

mcb30:
It's been a very long time (sorry) but: I would prefer not to include this unless it can be demonstrated that there is no better way to achieve the goal.  Adding `#ifdef MY_PET_FEATURE` is generally something I try hard to avoid.

In this case, I'd need to be convinced that the BIOS in question is detected as PnP by iPXE's `romprefix.S` (i.e. that the existing `NONPNP_HOOK_INT19` is insufficient), and that the BIOS is genuinely too broken to allow for booting from PnP option ROMs.  For example: if the BIOS can be demonstrated to fail to boot from a very standard option ROM (e.g. a NIC containing a known-working Intel PXE ROM) then that would be convincing evidence.

Closing this PR on the assumption that it is unlikely to receive any further attention after such a long delay, but please do reopen if it is still relevant.

PR #20, opened 2014-05-21
ConnectX3 support from Mellanox Flexboot

ya-mouse:
Merge FlexBoot's hermon driver code from http://www.mellanox.com/downloads/Drivers/PXE/gpxe_3.4.225.tgz to support MT27500.


ya-mouse:
Here it is:
http://www.mellanox.com/downloads/Drivers/PXE/FlexBoot-3.4.460_SRC.tar.gz
Looks like a new iPXE codebase. Is it acceptable to merge drivers from FlexBoot to upstream?

robinsmidsrod:
I guess it all depends on licensing. What kind of license is FlexBoot under, and how does it interface with/use iPXE? If it is practically a fork of iPXE with some extra drivers I don't see why not. But licensing is always a tricky subject.

ya-mouse:
Regarding to the License header in Mellanox drivers, it licensed under GPLv2 without exceptions.

ipxe-devel:
Hi,
We will push the code very soon to the flexboot branch in upstream.
We need to add few things before.
thanks

PR #30, opened 2014-10-01
Wrong UUID on (some) Dell systems

42wim:
Seems like some Dell desktops/laptops with older SMBIOS (2.4 in my tests)

have also UUIDs in little-endian, those need to be mangled also to be
correct. (if not, it seems to give issues with SCCM/WDS)

Explicitly check on 0x4c4c4544 for DELL


stappersg:
FWIW:  I think it would have helped if it were several commits.

The four year old commit does much more then
```
       if  ( uuid.canonical.a == 0x4c4c4544 ) {
                uuid_mangle ( &uuid );
       }
```
What the extra changes are, is now lost in the large commit ...

stappersg:
@42wim, or others who can close this merge request, please do.

PR #31, opened 2014-10-01
tftp booting from a windows TFTP server gives issues.

42wim:
PXE sends a leading / that needs to be removed.

e.g. chain tftp://ip.of.server/boot\x64\wdsbnp.com

PXE send TFTP Request /Boot\x64\wdsnbp.com
That first "/" must be removed.

Original fix from
http://forum.ipxe.org/archive/index.php/thread-7242.html


hildred:
This patch is very site specific, not portable, and may have other problems. I recommend against this patch being included in the main codebase.

PR #19, opened 2014-05-15
Restore code to strip an initial '/' from TFTP RRQ.

terryburton:

PR #33, opened 2015-01-07
Reverts changes to src/net/udp/tftp.c made in commit 76675365271291beb9d...

bpenglase:
...daeec10da14f4faa55ecc on 02/27/14.

This allows tftp to load from a Windows/Microsoft TFTP server.
This should also fix pull request #31: https://github.com/ipxe/ipxe/pull/31
Without this, when you try to load from a Windows/Microsoft TFTP server,
you get error 3d1260 "Inappropriate I/O control operation", as there is a leading "/".

### DIFF
--- a/src/net/udp/tftp.c
+++ b/src/net/udp/tftp.c
@@ -323,11 +323,23 @@ void tftp_set_mtftp_port ( unsigned int port ) {
  * @ret rc		Return status code
  */
 static int tftp_send_rrq ( struct tftp_request *tftp ) {
-	const char *path = tftp->uri->path;
 	struct tftp_rrq *rrq;
+	const char *path;	
 	size_t len;
 	struct io_buffer *iobuf;
 	size_t blksize;
+
+	/* Strip initial '/' if present. If we were opened via the
+	* URI interface, then there will be an initial '/', since a
+	* full tftp:// URI provides no way to specify a non-absolute
+	* path. However, many TFTP servers (particularly Windows
+	* TFTP servers) complain about having an initial '/', and it
+	* violates user expectations to have a '/' silently added to
+	* the DHCP-specified filename.
+	*/
+	path = tftp->uri->path;
+	if ( *path == '/' )
+		path++;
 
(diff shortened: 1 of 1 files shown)


bpenglase:
For reference, as I didn't noticed it at first, #19 from terryburton also reverts this as well.

src/net/udp/tftp.c
```diff
@@ -323,12 +323,24 @@ void tftp_set_mtftp_port ( unsigned int port ) {
...
 
+	/* Strip initial '/' if present. If we were opened via the
+	* URI interface, then there will be an initial '/', since a
+	* full tftp:// URI provides no way to specify a non-absolute
+	* path. However, many TFTP servers (particularly Windows
+	* TFTP servers) complain about having an initial '/', and it
+	* violates user expectations to have a '/' silently added to
+	* the DHCP-specified filename.
+	*/
+	path = tftp->uri->path;
```
hildred: how is lines 327 and 340 different from line 326?

I like your comment though.

NiKiZe:
I think #19 is the more correct revert of 76675365 just as @hildred notes.
One must also ask what kind of issues this causes  @mcb30 must have had some reason for the change more then just refactoring? And does the URI tests (added in the same commit) still succeed after this is applied?

bpenglase:
@hildred Good catch, all that is needed is the if statement, as #19 does. Closing this pull request in favor of @terryburton 's #19. Not sure if you meant the comment in the file, or on the commit, but the one in the file is not mine, and was on the original code prior to it being removed.

@NiKiZe Agreed, I missed the extra code. I also thought about what it may break, and I'm wondering if it's got something to do with the IPv6 stuff. I can give it a shot when I get home, as I don't have v6 here at work (yet). Granted, I have limited testing here, as I only have 2 TFTP servers, SCCM and the one that offers iPXE up. While I could pull some things over TFTP, I have everything defaulting over HTTP, mainly for performance. However, everything seems to be working well here.

hildred:
We should also close #31 in favor of #19.

PR #34, opened 2015-03-12
[crypto] New command imgdigest, verify image using digest

bakketun:
The imgdigest works like imgverify, expect that the image is verified
using a digest instead of a signature.

### DIFF
--- a/src/hci/commands/image_trust_cmd.c
+++ b/src/hci/commands/image_trust_cmd.c
@@ -161,6 +161,69 @@ static int imgverify_exec ( int argc, char **argv ) {
 	return rc;
 }
 
+/** "imgdigest" options */
+struct imgdigest_options {
+	/** Download timeout */
+	unsigned long timeout;
+};
+
+/** "imgdigest" option list */
+static struct option_descriptor imgdigest_opts[] = {
+	OPTION_DESC ( "timeout", 't', required_argument,
+		      struct imgdigest_options, timeout, parse_timeout),
+};
+
+/** "imgdigest" command descriptor */
+static struct command_descriptor imgdigest_cmd =
+	COMMAND_DESC ( struct imgdigest_options, imgdigest_opts, 3, 3,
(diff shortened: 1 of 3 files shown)


mcb30:
This seems to be a less useful variant of the existing "imgtrust" command, since it does not allow for the possibility of updating the image in future without also updating the embedded script.

What is the motivation for adding this feature?  What problem does it solve which cannot more flexibly be solved using the existing "imgtrust"?

mcb30:
(My last comment should of course say "imgverify command" instead of "imgtrust command"; sorry).

bakketun:
Example use case:

* DHCP server issues filename = http://10.0.1.1/boot.ipxe
* The subnet is isolated and contains only trusted servers. Thus we can trust the content of boot.ipxe without any signatures.
* boot.ipxe loads images from a different server. We can't trust this data, maybe because the connection is not secure.
* These images should never change. New images are published at different URLs.
* These images are then verified using imgdigest with an embedded digest.
* Thus, the images are essentially included in boot.ipxe, and can be trusted just as much as boot.ipxe it self.

When an image is updated:
* A new image is produced somewhere and it's hash is computed.
* We somehow decides if the new images hash is trustworthy.
* The boot.ipxe is updated with the new URL and hash.

Using imgverify instead:
* We need a signing certificate and private key.
* The private key must be kept secret.
* ipxe must be have some way of knowing if the certificate is trusted, probably by embedding it.
* When an image is changed we need access to the private key in order to create a new signature.
* We don't have to update boot.ipxe, but we do need to store the new signature file somewhere.

Often you might want to use both imgverify and imgdigest. The first the script is verified using imgverify, everything else using imgdigest. This is useful since the calculating the hashed don't require access to a private key. This is can done on a build server. If we don't trust the build server anymore, we can simply stop signing the hashes produced by it.

stappersg:
Ten years silence made the merge request ten years old and currently, 2025-04-16, the oldest open merge request.

PR #36, opened 2015-06-25
[wip][freebsd] Initial support freebsd boot

baloo:
Hi,

As this, this code does not work. I'm sending out this pull request to get feedback, and understand a bit better how memory is managed by ipxe.

The freebsd64 kernel is an elf64 kernel. The entry is an absolute virtual address. And the kernel expects to be placed at a specific address when booting. For that purpose, the freebsd loader relies on MMU to map destination virtual address to the physical address where the kernel has been allocated by the loader.
This mechanism is handled in the amd64_tramp.S and permits 32bit code to load 64bit kernel.

To perform this mapping, we need to allocate a PT4/3/2 pagetables and provide physical address to it. I'm unsure how to use the memory library in ipxe.

my comments are inline and feedback more than welcome :)

Thanks

### DIFF
--- a/src/arch/i386/image/freebsd.c
+++ b/src/arch/i386/image/freebsd.c
@@ -1,0 +1,203 @@
+/*
+ * Copyright (C) 2008 Michael Brown <mbrown@fensystems.co.uk>.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ *
+ * You can also choose to distribute this program under the terms of
+ * the Unmodified Binary Distribution Licence (as given in the file
+ * COPYING.UBDL), provided that you have satisfied its requirements.
+ */
+
+FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
+
+#include <errno.h>
+#include <elf.h>
+#include <stdio.h>
+#include <ipxe/image.h>
+#include <ipxe/elf.h>
+#include <ipxe/features.h>
+#include <ipxe/init.h>
+#include <ipxe/io.h>
+
+typedef uint64_t p4_entry_t;
+typedef uint64_t p3_entry_t;
+typedef uint64_t p2_entry_t;
+extern p4_entry_t PT4[];
+extern p3_entry_t PT3[];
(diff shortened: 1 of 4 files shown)


src/arch/i386/image/freebsd.c
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	kern_end = buffer;
+
+	entry_lo = ehdr.e_entry & 0xffffffff;
+	entry_hi = (ehdr.e_entry >> 32) & 0xffffffff;
+
+	for (i = 0; i < 512; i++) {
+		/* Each slot of the level 4 pages points to the same level 3 page */
+		PT4[i] = (p4_entry_t)VTOP((void *)&PT3[0]);
```
baloo: For this one, I need to get back the physical address of PT3, see https://svnweb.freebsd.org/base/head/sys/boot/i386/libi386/elf64_freebsd.c?view=markup#l98.

vtop is defined https://svnweb.freebsd.org/base/head/sys/boot/i386/btx/lib/btxv86.h?view=markup#l57 and __base https://svnweb.freebsd.org/base/head/sys/boot/i386/btx/lib/btxcsu.S?view=markup#l39

It looks like virtual_to_phys does not behave the same.

ipxe-devel:
On 25/06/15 15:17, Arthur Gautier wrote:
> The freebsd64 kernel is an elf64 kernel. The entry is an absolute
> virtual address. And the kernel expects to be placed at a specific
> address when booting. For that purpose, the freebsd loader relies on MMU
> to map destination virtual address to the physical address where the
> kernel has been allocated by the loader.
> This mechanism is handled in the amd64_tramp.S and permits 32bit code to
> load 64bit kernel.

Main question: does the FreeBSD kernel provide any 32-bit entry point? 
(For comparison, Linux bzImage kernels provide a 32-bit entry point even 
for 64-bit kernels.)

Secondary question: what does the FreeBSD kernel expect in terms of the 
page tables that we set up?  Is this documented somewhere?

The code could potentially be simplified by adding long-mode code and 
data segments to the usual iPXE GDT (defined for the standard BIOS build 
in arch/i386/transitions/librm.S).

It would be nice to have a generic mechanism for calling out to 
long-mode code, looking similar to the existing mechanisms for calling 
real-mode code or flat-physical-addressing code, e.g.:

   __asm__ __volatile__ ( REAL_CODE ( "int $0x16" ) );

   __asm__ __volatile__ ( PHYS_CODE ( "call *%%edi" ) : : "D" ( entry ) );

Michael

DarkHelmet433:
1) No, there is no 32 bit entry point.  It expects to be executed at its compiled address, with paging enabled, in long mode.  This is quite different to how the i386 kernel starts up.

2) It's not particularly fussy - but it does need V==P for the first 1GB or so, and the 1GB that the kernel is executing in.  I kept it simple for our loader and mapped first 1GB of physical memory over the entire address space.  All slots of the 4th level page table point to the same level 3 page, which point to the same level 2 page, which is full of 1:1 mappings for 2MB pages for the first 1GB.  This 12KB is discarded after very early in the bootstrap.

src/arch/i386/image/freebsd.c
DarkHelmet433: Don't worry too much about the specifics of vtop and friends, that is a btx-ism that won't apply the same way in this environment. btx runs with a non-zero segmentation offset.  You'll be looking for whatever the equivalent is here if there's even a difference between virtual and physical addresses.

PR #40, opened 2015-11-17
Add set option to prompt to save the input key as a variable

renini:
Ability to save the input key when using 'prompt' as a variable.
With this you are able to use this in the script further on, for example:

prompt --set keycombo --timeout 5000 Press Ctrl-B for the menu... && goto keypress || goto exit
:keypress
isset ${keycombo} || set keycombo 0x2
iseq ${keycombo} 0x1 && goto ctrla ||
iseq ${keycombo} 0x2 && goto ctrlb ||
iseq ${keycombo} 0x11 && goto ctrlq ||
echo You pressed: ${keycombo}
goto exit

### DIFF
--- a/src/image/script.c
+++ b/src/image/script.c
@@ -366,6 +366,8 @@ struct prompt_options {
 	unsigned int key;
 	/** Timeout */
 	unsigned long timeout;
+        /** Set */
+        struct named_setting set;
 };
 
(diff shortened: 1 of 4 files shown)


renini:
@mcb30 is this good enough to merge with upstream?

PR #41, opened 2015-11-19
add nfs block dev support

cleverca22:
this commit allows booting disk images over nfs with the sanboot command


stappersg:
Breaking the silence after five years  and aiming for this merge request to be closed. Time has shown _this feature is not wanted_.  There should have earlier a message like _it feels wrong to include block device on top of NFS code_.  @cleverca22, or others who can close this MR, please do.

PR #45, opened 2016-03-03
[efi] Remove 'hlt' instruction as it causes us to lose interrupts.

jdieter:
For some reason, running 'hlt' is causing us to lose keypresses because we are (I assume) losing interrupts.  Removing it means that we don't lose any keypresses.

Tested on ASUS Z77 motherboard.

### DIFF
--- a/src/arch/x86/interface/efi/efix86_nap.c
+++ b/src/arch/x86/interface/efi/efix86_nap.c
@@ -46,8 +46,10 @@ static void efix86_cpu_nap ( void ) {
 	 * The EFI shell doesn't seem to bother sleeping the CPU; it
 	 * just sits there idly burning power.
 	 *
+	 * Originally we tried using the 'hlt' instruction, but it causes
+	 * us to lose interrupts.  So we'll just have to fake it, and run
+	 * the cpu at full power.
 	 */
-	__asm__ __volatile__ ( "hlt" );
 }
 
 PROVIDE_NAP ( efix86, cpu_nap, efix86_cpu_nap );


ipxe-devel:
On 03/03/16 09:41, Jonathan Dieter wrote:
> For some reason, running 'hlt' is causing us to lose keypresses because
> we are (I assume) losing interrupts. Removing it means that we don't
> lose any keypresses.
>
> Tested on ASUS Z77 motherboard.

Good catch.  Unfortunately that patch would cause iPXE to spin at 100% 
CPU when idle.  This is particularly bad for VMs.  We'll need to find 
some alternative solution.

Michael

jdieter:
Ok, ignore this.  I've just retested and it didn't fix the problem, so something else is going on.  I'll keep looking.

jdieter:
Ok, it turned out that this *does* fix the problem after I fixed a different bug in commit #44, so I'm reopening (even though we still need to find an alternative solution)

jdieter:
To be more specific, if I comment out cpu_nap() in core/getkey.c, line 54, eveything works fine.  Unfortunately, that seems to be the most common case in the main loop there, which means we'll be spinning at 100% CPU.

jdieter:
I've found a workaround.  I've put the following in config/local/nap.h:

#undef          NAP_EFIX86
#define         NAP_NULL

It doesn't really fix the problem, but I suspect this is a buggy BIOS and there's not much we can do to get around it short of spinning at 100% CPU.  FWIW, it's an **MSI** Z77A-G45 Gaming motherboard with the latest BIOS.  I'm going to go ahead and close this one now.

PR #46, opened 2016-03-10
[efi] Get font height even if GetFontInfo() doesn't work

jdieter:
On my MSI Z77A-G45, GetFontInfo() returns ERR_NOT_FOUND all the time, but we can still get our glyph information.  The only reason we run GetFontInfo() is to get the font height, so this patch finds the font height using the incredibly cunning algorithm of choosing the largest font height of all the glyphs if GetFontInfo() returns an error.


jdieter:
Thanks, that looks great!

PR #48, opened 2016-03-14
Initial support for automatically choosing resolution based on EDID

jdieter:
Ok, here's a more elegant way of automatically choosing the resolution based on the EDID returned by the monitor.  This patch adds an "-a" option to the console command that will automatically choose the best resolution.

edid.c and edid.h are originally from xrandr, but with some formatting changes to make them fit in better with the ipxe code base, and some new interface code.

Compiling with DEBUG=edid:2 will give you a human-readable dump of the EDID when you run "console -a".


jdieter:
Looking at grub's code, it looks like it _might_ be reasonably sane to extract the EDID when booting using a legacy BIOS.  If this looks good, that will be the next thing I look into.

ipxe-devel:
On 14/03/16 12:51, Jonathan Dieter wrote:
> Ok, here's a more elegant way of automatically choosing the resolution
> based on the EDID returned by the monitor. This patch adds an "-a"
> option to the console command that will automatically choose the best
> resolution.
>
> edid.c and edid.h are originally from xrandr, but with some formatting
> changes to make them fit in better with the ipxe code base, and some new
> interface code.
>
> Compiling with DEBUG=edid:2 will give you a human-readable dump of the
> EDID when you run "console -a".

Thanks.  The code appears to extract a whole batch of information from 
EDID that is not required for iPXE.  This includes, for example, the 
aspect ratio: this is a floating point value, and floating point is 
simply not supported inside iPXE.

Could you strip it down to the bare minimum required?  I think the only 
single piece of information that you need is the preferred resolution; 
nothing else needs to be extracted.  It should reduce to less than 
around 200 lines of code in total.

Thanks,

Michael

jdieter:
Ok, I'm working on reducing this.  What about the CRC check?  Should I still fail hard if it fails?

jdieter:
Ok, that's down to 235 lines total.  I could strip it down more, but that would mean stripping out documentation in EdidActive.h.

ipxe-devel:
On Mon, 2016-03-14 at 12:02 -0700, Jonathan Dieter wrote:
> Ok, that's down to 235 lines total. I could strip it down more, but
> that would mean stripping out documentation in EdidActive.h.

Is there anything else you'd like me to with this patch?

Jonathan

stappersg:
Breaking the silence after four years  and aiming for this merge request to be closed. Time has shown _this feature is not wanted_.  There should have earlier a message like _it feels wrong to include screen resolution code_.   @jdieter or others who can close this MR please do.

jdieter:
Sure.  Closing now.

NiKiZe:
Revisiting this, would it be cleaned to extract this information to an edid settings block that could be used for different things?
Some of the data could be really interesting to have there as well - except for maybe timings? ;)

PR #49, opened 2016-03-15
Fix no-PIE workaround for i386 builds

eworm-de:
This workaround did not work for my version of gcc (4.9.2 20150304) as
no option -nopie exists.

We take another way: Let's check whether or not the macro **PIE** is defined
and add -fno-PIE if it is.

### DIFF
--- a/src/arch/i386/Makefile
+++ b/src/arch/i386/Makefile
@@ -69,14 +69,21 @@ CFLAGS		+= -fshort-wchar
 #
 CFLAGS			+= -Ui386
 
-# Some widespread patched versions of gcc include -fPIE -Wl,-pie by
-# default.  Note that gcc will exit *successfully* if it fails to
-# recognise an option that starts with "no", so we have to test for
-# output on stderr instead of checking the exit status.
+# Some widespread patched versions of gcc include -fPIE -Wl,-pie or
+# similar by default or enable hardening via toolchain.
+# First use the preprocessor to check whether or not a workaround
+# is required, then enable supported flags. Note that gcc will exit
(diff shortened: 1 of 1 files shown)


ipxe-devel:
On 15/03/16 13:33, Christian Hesse wrote:
> This workaround did not work for my version of gcc (4.9.2 20150304) as
> no option -nopie exists.
>
> We take another way: Let's check whether or not the macro *PIE* is defined
> and add -fno-PIE if it is.

Thanks.  Your patch seems to completely remove the "-nopie" option for 
all versions of gcc.  I'm pretty sure this is going to break the 
versions that required the workaround in the first place.

Does your version of gcc actually require the workaround?  It's not 
required on mine (also 4.9.2).

Michael

eworm-de:
> Thanks. Your patch seems to completely remove the "-nopie" option for all versions of gcc. I'm pretty sure this is going to break the versions that required the workaround in the first place.

I did not find a version of `gcc` that requires (or even understands) `-nopie`.
Do you have any information what systems do require this?

> Does your version of gcc actually require the workaround? It's not required on mine (also 4.9.2).

By default `gcc` does not enable PIE. This changes when you enable hardening toolchains and/or enable some options in your environment, though.

But that is what my change is about: It uses the preprocessor to check whether or not `gcc` generates PIE code. The workaround is added in first case only.

Thinking about some more changes...

eworm-de:
Reworked this a bit... I hope it covers all cases now.

eworm-de:
Should be fixed with 7c395b0e21806b946fe944a27fc273407f357ea1, closing.

PR #51, opened 2016-03-22
make golan_check_rc_and_cmd_status() static

eworm-de:
This fixes compiler error:
'__bswap_variable_32' is static but used in inline function
'golan_check_rc_and_cmd_status' which is not static


eworm-de:
Fixed with 05027a7a125eee751e7aa2cd93f3a95f683f901a. Thanks!

PR #53, opened 2016-05-09
ath9k: fix buffer overrun for ar9287

eworm-de:
This backport is from linux kernel upstream commit:

> commit 83d6f1f15f8cce844b0a131cbc63e444620e48b5
> Author: Arnd Bergmann <arnd@arndb.de>
> Date:   Mon Mar 14 15:18:36 2016 +0100
>
> ath9k: fix buffer overrun for ar9287
>
> Code that was added back in 2.6.38 has an obvious overflow
> when accessing a static array, and at the time it was added
> only a code comment was put in front of it as a reminder
> to have it reviewed properly.
>
> This has not happened, but gcc-6 now points to the specific
> overflow:
>
> drivers/net/wireless/ath/ath9k/eeprom.c: In function 'ath9k_hw_get_gain_boundaries_pdadcs':
> drivers/net/wireless/ath/ath9k/eeprom.c:483:44: error: array subscript is above array bounds [-Werror=array-bounds]
>      maxPwrT4[i] = data_9287[idxL].pwrPdg[i][4];
>                    ~~~~~~~~~~~~~~~~~~~~~~~~~^~~
>
> It turns out that the correct array length exists in the local
> 'intercepts' variable of this function, so we can just use that
> instead of hardcoding '4', so this patch changes all three
> instances to use that variable. The other two instances were
> already correct, but it's more consistent this way.
>
> Signed-off-by: Arnd Bergmann <arnd@arndb.de>
> Fixes: 940cd2c12ebf ("ath9k_hw: merge the ar9287 version of ath9k_hw_get_gain_boundaries_pdadcs")
> Signed-off-by: David S. Miller <davem@davemloft.net>

### DIFF
--- a/src/drivers/net/ath/ath9k/ath9k_eeprom.c
+++ b/src/drivers/net/ath/ath9k/ath9k_eeprom.c
@@ -368,10 +368,9 @@ void ath9k_hw_get_gain_boundaries_pdadcs(struct ath_hw *ah,
 
 	if (match) {
 		if (AR_SREV_9287(ah)) {
-			/* FIXME: array overrun? */
 			for (i = 0; i < numXpdGains; i++) {
 				minPwrT4[i] = data_9287[idxL].pwrPdg[i][0];
-				maxPwrT4[i] = data_9287[idxL].pwrPdg[i][4];
+				maxPwrT4[i] = data_9287[idxL].pwrPdg[i][intercepts - 1];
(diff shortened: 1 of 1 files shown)


eworm-de:
The commit in upstream linux repository:
[ath9k: fix buffer overrun for ar9287](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=83d6f1f1)

PR #54, opened 2016-05-28
Apple NetBoot EFI protocol for chainloading Mac OS X on Apple computers

mkuron:
When booting iPXE over the network on a Mac via the method described in the forum thread about [Mac Pro (late 2013 model) EFI and iPXE booting](http://forum.ipxe.org/showthread.php?tid=7323&pid=11253#pid11253) (wrapping snponly.efi into a .nbi bundle and serving that via Apple's NetBoot/NetInstall service), one can chainload Linux kernels and EFI binaries just fine.
However, one cannot chainload Mac OS X 10.11 and higher because Apple does not use standard PXE EFI protocols to obtain the DHCP packet from EFI. For Mac OS X 10.10 and lower, this was no problem as the kernel contained a DHCP client (see [bsd/kern/netboot.c](http://opensource.apple.com//source/xnu/xnu-2782.40.9/bsd/kern/netboot.c), line 646) that was used when the DHCP packet could not be obtained from EFI. Mac OS X 10.11 removes that DHCP client and panics if no DHCP packet can be obtained from EFI.

This pull request adds support for the Apple NetBoot EFI protocol to iPXE so that iPXE can pass on its DHCP packet to the operating system.
With this patch applied, Mac OS X 10.11 and higher can be chainloaded like this:
```
set nbiname OSX-10111.nbi
set serverip ${next-server}
set tftpboot tftp://${serverip}
set netbootsp0 /Library/NetBoot/NetBootSP0
set root-path nfs:${serverip}:${netbootsp0}:${nbiname}/NetInstall.dmg
initrd --name x86_64\kernelcache ${tftpboot}/NetBoot/NetBootSP0/${nbiname}/i386/x86_64/kernelcache
kernel ${tftpboot}/NetBoot/NetBootSP0/${nbiname}/i386/booter
imgargs booter -v rp=${root-path}
boot
```
Without this patch applied, the above only worked on Mac OS X 10.10.5 and lower.

*Acknowledgements*:
The details about the Apple NetBoot EFI protocol came from https://github.com/Piker-Alpha/macosxbootloader/blob/master/sdk/include/Protocol/AppleNetBoot/AppleNetBoot.h as Apple has not, to my knowledge, published a specification about their EFI protocols.


ipxe-devel:
On 28/05/16 15:49, Michael Kuron wrote:
> When booting iPXE over the network on a Mac via the method described at
> This pull request adds support for the Apple NetBoot EFI protocol to
> iPXE so that iPXE can pass on its DHCP packet to the operating system.

Thank you.  I've rewritten your patch to make it substantially simpler 
and to add all of the missing error handling code paths.  Could you 
please test:

   http://git.ipxe.org/people/mcb30/ipxe.git/shortlog/refs/heads/apple

and let me know if it works?

Thanks,

Michael

mkuron:
Thanks for the code cleanup. This was my first attempt at UEFI programming and the iPXE code, so I was expecting it would need a rewrite to meet the quality standards.

I just compiled your branch and did a NetBoot in VMware, which worked fine. I'll do another test on real hardware tomorrow, but I expect the code can be merged as-is.

ipxe-devel:
On 29/05/16 13:43, Michael Kuron wrote:
> Thanks for the code cleanup. This was my first attempt at UEFI
> programming and the iPXE code, so I was expecting it would need a
> rewrite to meet the quality standards.
>
> I just compiled your branch and did a NetBoot in VMware, which worked
> fine. I'll do another test on real hardware tomorrow, but I expect the
> code can be merged as-is.

Thank you for testing (in the VM); have pushed to the master branch. 
Let me know if anything doesn't work on the real heardware.

Thanks,

Michael

dreamcat4:
This is awesome. Great news!

FYI:

BSDPy - for mac netbooting - https://github.com/bruienne/bsdpy

Although would it not be even better, if we could perhaps integrate missing functionality directly into dnsmasq?

I mean, for the Apple NetBoot EFI protocol to serve the .nbi file over tftp or whichever.

PR #56, opened 2016-09-07
Support for 1G autoneg on ixgbe

baloo:
This is actually an implementation of the "4.6.4.4 KX / KX4 / KR Link Setup Flow With AutoNegotiation" of  http://www.intel.com/content/dam/www/public/us/en/documents/datasheets/82599-10-gbe-controller-datasheet.pdf which is needed to get 1G links on 10G interfaces

It has been tested on 82599EN with SFP 1G.


h0tw1r3:
Doesn't work for me.

PR #57, opened 2016-10-10
[digest] add sha1verify command

bmwiedemann:
This allows to get a trusted image with known digest value
that is embedded in the ipxe script

example usage:
imgtrust
kernel http://boot.ipxe.org/memtest.0
sha1verify memtest.0 5d78d4c7e97c99bca3d3dff602b84dde5b70c5bf
boot

an md5verify cmd could easily be added,
but would not be recommended to use
because of known weaknesses with MD5 digests


bmwiedemann:
ping

pandrew:
There is weaknesses in sha1 to; https://shattered.io Go for sha256 at least.

mcb30:
It needs a substantial cleanup before being mergeable.  Also, rather than argue over which digest algorithm commands should exist, we should be able to use whichever digests are compiled in (which is controlled by config/crypto.h); this will require either a command syntax such as "digest sha256 <image>" with digests collected via the linker table mechanism, or per-digest files linked in as necessary (as used by e.g. crypto/mishmash/rsa_sha256.c).

bmwiedemann:
I just noticed that this is similar to #34 but I cannot follow up on either. So, if someone wants to pick this up, I'd welcome it.

PR #59, opened 2016-12-19
Change __divmoddi4 from int64 [unknown] to int64_t

warpcoil:
Also matches the header file [libgcc.h]

and solves __divmoddi4.c:3:56: error: unknown type name ‘int64’


mcb30:
Merged in http://github.com/ipxe/ipxe/commit/cede0c5ba; thank you!

PR #63, opened 2017-05-11
Update undinet.c

scott-xu:
Support Broadcom Corporation NetXtreme BCM5717 Gigabit Ethernet PCIe and Intel Corporation 82574L Gigabit Network Connection


mcb30:
Your commit message gives no information about why this workaround is required.  Which systems are known to include these broken PXE ROMs that require the workaround?

scott-xu:
It was confirmed that the 2 NIC types in question had a broken interrupts interface. So they have to use polling mode.

mcb30:
Not all Intel PXE ROMs are broken.  Which systems with these NICs are known to ship with broken PXE ROMs?  For example, the current undinet_irq_broken_list mentions the HP XX70x and HP 745 G3 laptops.

Thanks,

Michael

scott-xu:
Sorry I do not have that information

PR #65, opened 2017-07-27
[acpi] Compute and check checksum for ACPI tables

gourvy:


ipxe-devel:
On 27/07/17 16:35, gourvy wrote:
>   * [acpi] Compute and check checksum for ACPI tables

Some comments:

1. What is the purpose of this patch?  What problem is it trying to solve?

2. ACPI tables use a byte checksum not a CRC, so the debug message is 
misleading.

3. It would almost certainly be simpler to have a function 
acpi_checksum() that just calculates and returns the byte checksum.  You 
can do this without using malloc() by using copy_from_user() for each 
byte in turn; this looks slightly cumbersome but will be optimised away 
by the compiler to the equivalent of just "data[i]" anyway.  The 
function can then return the checksum value, which the caller can then 
just compare against zero.

3a. For bonus points, acpi_fix_checksum() could then become just
   acpi->checksum -= acpi_checksum ( virt_to_user ( acpi ) );

Thanks,

Michael

gourvy:
HI Michael,

Thanks for reviewing my commit :)

1. It doesn't solve any problem, but I see 2 purposes. First, if a table is corrupted we shouldn't use the table (that may prevent big problems). Second, (and why I implemented the check) when manipulating ACPI tables it's good to know that the tables are 'probably' not corrupted. For example, I'm working with MADT in ipxe, but I had to debug. So first thing I did was checking the checksum.

2. Debug message fixed.

3. Nice, I didn't think the compiler would optimize the copy_from_user() calls. Done

3a. Good idea, so it's done

Thanks,

Laurent

PR #66, opened 2017-10-16
Handle DHCPNAK by going back to discover state.

blakerouse:
Adds support for handling a DHCPNAK when performing DHCP. It is possible that once iPXE has picked the IP address and sends the request that the DHCP server response with a DHCPNAK instead of a DHCPACK. According to RFC the DHCP client should start back at discover.

This was identified here: https://bugs.launchpad.net/bugs/1707999

The patch has been tested in Ubuntu 16.04 and verified to work.

### DIFF
--- a/src/net/udp/dhcp.c
+++ b/src/net/udp/dhcp.c
@@ -554,9 +554,18 @@ static void dhcp_request_rx ( struct dhcp_session *dhcp,
 		DBGC ( dhcp, " for %s", inet_ntoa ( ip ) );
 	DBGC ( dhcp, "\n" );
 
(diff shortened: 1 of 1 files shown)


kikoreis:
Gentle bump on this patch, as it's affecting us with KVM testing.

mcb30:
Updated to use the discovery deferral logic: this avoids hammering a misconfigured DHCP server with a non-stop stream of requests and also allows the DHCP process to eventually time out and fail.

Pushed as commit https://github.com/ipxe/ipxe/commit/1192edf39

Thanks!

PR #67, opened 2018-01-11
[dhcp/ipv4] Support for DHCP option 121: Classless static route option

barakelde:
This PR adds support for classless static route option (https://tools.ietf.org/rfc/rfc3442.txt).

I appreciate your comments and feedback, since it is my first contribution to an open-source project.

### DIFF
--- a/src/net/ipv4.c
+++ b/src/net/ipv4.c
@@ -337,9 +337,12 @@ static int ipv4_tx ( struct io_buffer *iobuf,
 	if ( ( next_hop.s_addr != INADDR_BROADCAST ) &&
 	     ( ( miniroute = ipv4_route ( sin_dest->sin_scope_id,
 					  &next_hop ) ) != NULL ) ) {
-		iphdr->src = miniroute->address;
 		netmask = miniroute->netmask;
 		netdev = miniroute->netdev;
+
+		struct settings *settings = netdev_settings ( netdev );
+		fetch_ipv4_setting ( settings, &ip_setting, &iphdr->src );
+		next_hop = miniroute->gateway;
 	}
 	if ( ! netdev ) {
 		DBGC ( sin_dest->sin_addr, "IPv4 has no route to %s\n",
@@ -448,15 +451,13 @@ int ipv4_has_any_addr ( struct net_device *netdev ) {
  * @ret has_addr	Network device has this IPv4 address
  */
 static int ipv4_has_addr ( struct net_device *netdev, struct in_addr addr ) {
-	struct ipv4_miniroute *miniroute;
+	struct in_addr address;
+	struct settings *settings = netdev_settings ( netdev );
+	fetch_ipv4_setting ( settings, &ip_setting, &address );
 
-	list_for_each_entry ( miniroute, &ipv4_miniroutes, list ) {
-		if ( ( miniroute->netdev == netdev ) &&
-		     ( miniroute->address.s_addr == addr.s_addr ) ) {
-			/* Found matching address */
-			return 1;
-		}
-	}
+	if ( address.s_addr == addr.s_addr )
+		return 1;
+
 	return 0;
 }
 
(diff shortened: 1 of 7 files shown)


src/net/ipv4.c
```diff
@@ -844,6 +845,36 @@ static int ipv4_gratuitous_arp ( struct net_device *netdev,
...
+		sprintf(buf, "255.%d.0.0", n_octet);
+		addr_len = 2;
+	} else if (cidr > 0) {
+		sprintf(buf, "%d.0.0.0", n_octet);
+		addr_len = 1;
+	} else if (cidr == 0) {
+		sprintf(buf, "0.0.0.0");
+		addr_len = 0;
+	} else {
+		DBGC(cidr, "Unknown netmask descriptor /%d.\n", cidr);
```
multun: I don't quite get what's the point of having an individual format string for each full byte. Isn't converting it to a netmask already sufficient ?

```python
def cidr_to_netmask(cidr):
     return tuple(0xFF & ~((1 << (8 - min(8, cidr - i * 8))) - 1) for i in range(4))
```

src/net/ipv4.c
```diff
@@ -844,6 +845,36 @@ static int ipv4_gratuitous_arp ( struct net_device *netdev,
...
+	}
+
+	return addr_len;
```
multun: why don't you return `cidr / 8` ?

src/util/elf2efi.c
```diff
@@ -635,6 +635,7 @@ static void process_reloc ( struct elf_file *elf, const Elf_Shdr *shdr,
 		case ELF_MREL ( EM_ARM, R_ARM_THM_JUMP24 ) :
 		case ELF_MREL ( EM_ARM, R_ARM_V4BX ):
 		case ELF_MREL ( EM_X86_64, R_X86_64_PC32 ) :
+		case ELF_MREL ( EM_X86_64, R_X86_64_PLT32 ) :
```
multun: it's not related is it ? why would PLT calls be emitted anyway ?

src/net/ipv4.c
```diff
@@ -890,9 +921,51 @@ static int ipv4_settings ( int ( * apply ) ( struct net_device *netdev,
...
+				addr_octets_len = cidr_to_mask ( c, buf );
+				inet_aton( buf, &netmask );
+
+				/* Address */
+				memset ( uaddr, 0, 4 * sizeof ( uint8_t ) );
+				for ( j = 0; j < addr_octets_len; j++ )
+					uaddr[j] = raw_data[i + j];
+
+				sprintf( buf, "%d.%d.%d.%d", uaddr[0], uaddr[1], uaddr[2], uaddr[3] );
+				inet_aton( buf, &address );
```
multun: why don't you fill out the in_addr directly, using an union for example ?

src/net/ipv4.c
```diff
@@ -890,9 +921,51 @@ static int ipv4_settings ( int ( * apply ) ( struct net_device *netdev,
...
+		/* Configure route (-s) */
+		uint8_t * raw_data = netdev->static_routes;
+		uint8_t option_len = netdev->static_routes_len;
+
+		/* Check if static routes available
+		 * The minimum length for DHCP option 121 is 5 bytes */
+		if ( option_len >= 5 ) {
+			char buf[16];
+			uint8_t uaddr[4];
+			uint8_t i = 0, j, addr_octets_len;
```
multun: declaring multiple variables on a single lines usually is bad style.
the only places you may find it in ipxe is within code ported from linux.

j also is reused multiple times eventhough it could be a different variable every single time.

hazelesque:
@fnfq Oh sweet, someone else wants classless static routes support! Yay! \o/

I wrote a patch last summer to add RFC3442 support ([diff here](https://github.com/ipxe/ipxe/compare/master...hazelesque:static-routes-support) if you're interested) but then I was taken ill and was off work for months, so - whilst it did in fact _work_ - I never got around to tidying it up and submitting a PR for it.

I wish you luck with your PR, and feel free to nick anything you want from my patch if it helps at all! :)

stappersg:
The patch @hazelesque  talks about became #78

mcb30:

PR #68, opened 2018-02-16
Added support for BCM5762

agentorange96:
There are some other cards that could be added pretty easily I could add later. Although this is the only one I could test, and needed right now. So it's the only one I added for right now and the only one I'd be able to test anyway. (It worked.)


agentorange96:
I'm having issues actually with configuration, so I'm going to further work on this and submit a new pull request once I figure that out.

PR #69, opened 2018-03-14
[intel] Add PCI_ROM entry for Intel i354 NIC

lasalvavida:
Have tested on a [SuperMicro A1Sri](http://www.supermicro.com/products/motherboard/atom/x10/a1sri-2758f.cfm?gclid=EAIaIQobChMI57SXw8_s2QIVllmGCh380gdFEAAYASAAEgIAffD_BwE) with this NIC.

Functions properly provided ASDE is disabled as those bits are reserved for this chipset. My observation was that without ASDE disabled, the adapter appears to function properly, not reporting errors or dropped packets in statistics, but not reporting successful transmission of packets either. Transmitted packets get processed but never actually make it onto the network.


mcb30:
Applied; thanks.

PR #70, opened 2018-03-14
[efi_pci] Elevate TPL around pci_probe

lasalvavida:
This resolves a case where my system would hang on the UEFI call to StartImage.

When a PCI NIC gets probed successfully, it calls register_netdev() which in turn uses the least significant bits of the MAC to seed the random number generator. This calls into random() which uses currticks() for the initial seed. On EFI builds, this restores the TPL level to TPL_APPLICATION to allow tick events to fire, and then raises it back to TPL_CALLBACK. Not matching calls to RaiseTPL with calls to RestoreTPL causes undefined behavior, in this case a system hang. Adding the matching calls around pci_probe resolves this issue.

### DIFF
--- a/src/interface/efi/efi_pci.c
+++ b/src/interface/efi/efi_pci.c
@@ -383,6 +383,7 @@ static int efipci_supported ( EFI_HANDLE device ) {
  * @ret rc		Return status code
  */
 static int efipci_start ( struct efi_device *efidev ) {
+	EFI_BOOT_SERVICES *bs = efi_systab->BootServices;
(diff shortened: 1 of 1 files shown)


mcb30:
Good catch.  Should be fixed by the slightly more general solution at http://git.ipxe.org/ipxe.git/commitdiff/10d083ffa - please let me know if this doesn't fix your problem.  Thanks!

lasalvavida:
Confirming that [10d083ffa](http://git.ipxe.org/ipxe.git/commitdiff/10d083ffa) resolved this - Thanks!

PR #73, opened 2018-04-27
[efi_snp] Subtract link-layer header length from MaxPacketSize

lasalvavida:
It seems that the EDK network stack expects this to be 1500, not 1514.

As-is, in the case of fragmentation, the UEFI UDP layer will compute fragment sizes larger than the MTU which typically get dropped.


mcb30:
Merged (with a minor change to use netdev->mtu directly instead of replicating the calculation); thanks!

PR #74, opened 2018-05-10
Add support for Intel X552 NIC.

sthaber:
Tested on... you guessed it... an Intel X552 NIC. Everything works fine; just needed the PCI ID in the driver.


mcb30:
Merged; thanks.

PR #76, opened 2018-08-19
gensdsk: fix default target

bit:
gensdsk creates a syslinux.cfg file that is invalid if the filename ends in lkrn, to fix this point default target to label($b) not filename($g)

### DIFF
--- a/src/util/gensdsk
+++ b/src/util/gensdsk
@@ -48,7 +48,7 @@ do
 	g=${g:0:8}.krn
 	case "$first" in
 	"")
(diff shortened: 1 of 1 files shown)


stappersg:
In twelve days becomes this pull request six months old.

I do hope that this PR gets a real judgement before that date.  Either by a merge or a reject. Leaving it hanging/dangling/rotting here has no use ...

mcb30:
Merged as http://github.com/ipxe/ipxe/commit/4277942ac; thank you.

PR #78, opened 2018-08-31
Static routes support

hazelesque:


stappersg:
Abstract from https://tools.ietf.org/html/rfc3442
> a new Dynamic Host Configuration Protocol (DHCP) option which is passed from the DHCP Server to the DHCP Client to configure a list of static routes in the client.  The network destinations in these routes are classless - each routing table entr includes a subnet mask.

Further text from RFC3442
> Classful routing is no longer in common use, so the DHCP Static Route option is no longer useful.  Currently, classless routing is the most commonly-deployed form of routing on the Internet.

stappersg:
Please close this issue.

saschaarthur:
was this merged? DHCP Option 121 now supported?

hazelesque:
Sadly not, to the best of my knowledge.

The patch I wrote needed tidying up and then splitting cleanly into three
pieces (the routing table structure and helper method changes, the DHCP
option parsing, and the code to take the parsed option data and update the
routing table with it), and I was taken seriously ill before I could finish
it.

I had hoped to come back to it some time, but I've had cancer twice in the
last 5 years, and whilst I'm basically recovered (again) now, that has
really thrown a spanner in the works and I genuinely have no idea when I
would have time to get back to this, even if the developers were still
interested in the idea of it being merged.

(I should also note that last time I checked, they preferred patches to be
submitted via the mailing list rather than GitHub PRs.)

Should you wish to take this on, you would of course be welcome to do so as
far as I am concerned, but I'd strongly recommend reaching out to the
developers, e.g. via IRC in Libera.Chat channel #ipxe.

I hope this helps clarify things.

Thanks,

Hazel

saschaarthur:
Hey,

Thanks for the detailed Update! 

Will have a look and if i can push the progress a bit forward.

Im sorry for your health history and wish you all the best for your future health!

mcb30:

PR #80, opened 2018-10-11
[build] Do not use WORKAROUND_CFLAGS with HOST_CC when cross-compiling

ignatk:
WORKAROUND_CFLAGS are populated based on the capabilities and properties of the
target compiler. When cross-compiling, the target compiler and the host compiler
are different, so the WORKAROUND_CFLAGS might not be compatible with the host
compiler.

For example, when cross-compiling for arm64, the current build system adds
`-mabi=lp64` to the WORKAROUND_CFLAGS. Later, when the build system tries to
compile `elf2efi64` tool, the compilation fails, because x86 gcc does not
understand the `-mabi=lp64` flag, which is arm64 specific.


habetsm:
Acked-by: Martin Habets <mhabets@solarflare.com>

mcb30:
Applied an alternative fix: http://git.ipxe.org/ipxe.git/commitdiff/a4f8c6e31

PR #81, opened 2018-10-15
[build] Fix asm argument modifiers for aarch64 build

jjolly:
When building bin-arm64-efi/snp.efi using the gcc8 compiler, the following error occurs:

    In file included from core/xferbuf.c:28:
    core/xferbuf.c: In function 'xferbuf_malloc_realloc':
    include/errno.h:261:2: error: invalid 'asm': invalid address mode
      __asm__ ( ".section \".einfo\", \"\", " PROGBITS_OPS "\n\t" \
      ^~~~~~~
    include/errno.h:549:16: note: in expansion of macro '__einfo_error'
     #define ENOSPC __einfo_error ( EINFO_ENOSPC )
                    ^~~~~~~~~~~~~
    core/xferbuf.c:192:11: note: in expansion of macro 'ENOSPC'
       return -ENOSPC;
               ^~~~~~

The 'a' address mode is invalid for aarch64. Simply changing this to
the constant operand type 'c' seems to resolve the issue on aarch64 and
does not appear to affect x86_64 or i586.

Signed-off-by: John L. Jolly <jjolly@suse.com>

### DIFF
--- a/src/include/errno.h
+++ b/src/include/errno.h
@@ -262,10 +262,10 @@ static inline void eplatform_discard ( int dummy __unused, ... ) {}
 		  ".align 8\n\t"					\
 		  "\n1:\n\t"						\
 		  ".long ( 4f - 1b )\n\t"				\
-		  ".long %a0\n\t"					\
+		  ".long %c0\n\t"					\
 		  ".long ( 2f - 1b )\n\t"				\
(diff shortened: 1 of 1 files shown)


mcb30:
Applied with an alternative commit message explaining the history of the use of "%a0".  Thank you, and sorry for the delay in applying.

PR #82, opened 2018-10-24
[build] Do not use random build_ids

bmwiedemann:
Do not use random build_ids
to allow for reproducible builds of .rom files
and xen hvmloader (which embeds those)

See https://reproducible-builds.org/ for why this matters.

Signed-off-by: Bernhard M. Wiedemann <bwiedemann@suse.com>

### DIFF
--- a/src/Makefile.housekeeping
+++ b/src/Makefile.housekeeping
@@ -1162,7 +1162,7 @@ blib : $(BLIB)
 # Command to generate build ID.  Must be unique for each $(BIN)/%.tmp,
 # even within the same build run.
 #
-BUILD_ID_CMD	:= perl -e 'printf "0x%08x", int ( rand ( 0xffffffff ) );'
+BUILD_ID_CMD	:= sh -c 'printf 0x%s $$(cat "$$@" | sha1sum - | cut -c-8)' --
 
 # Build timestamp
 #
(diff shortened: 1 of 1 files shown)


NiKiZe:
Commit that introduced this feature: https://git.ipxe.org/ipxe.git/commitdiff/58f6e553625c90d928ddd54b8f31634a5b26f05e

Each ROM have a need for identifier per build
This is described by @mcb30 at http://lists.ipxe.org/pipermail/ipxe-devel/2015-February/003978.html as well - I'm reading as using checksum here is not acceptable.

Also the comment above the rand line says this:
#... Must be unique for each $(BIN)/%.tmp,
# even within the same build run.
if checksum is used that comment needs update as well?

A patch that makes this stable for non roms would probably be acceptable, but for roms it won't be unless the reason for the buildids existence is solved some other way instead.

bmwiedemann:
The idea of the patch was that hashing all relevant inputs still gives build-ids that are different when they should.
I was trying to get xen to build reproducibly, but just noticed that it does not even run this code-path for its etherboot. So I cannot test this.
Feel free to close.

krystian-hebel:
@bmwiedemann You can also use `make BUILD_ID_CMD="echo 0xfeedface"` or something like that

mcb30:
Closed in favour of #234

PR #83, opened 2018-11-04
set INTEL_NO_PHY_RST on I218-V

bleything:
I'm using Intel D34010WYK NUCs which have the I218-V onboard. Prior to this change they always negotiate at 10mbps. After this change they will negotiate at 1gbps.

I... have no idea what I'm doing here, so please let me know if there are further changes or tests or whatever that I need to address!


bleything:
This appears to only work when booting in EFI mode. I'll try to figure out how to get it to work in BIOS mode.

mcb30:
Merged as http://github.com/ipxe/ipxe/commit/afaf4272b, thank you!

PR #84, opened 2018-11-18
[efi] Enable NET_PROTO_IPV6 by default

toreanderson:
IPv6 PXE was included in the UEFI specification over eight years ago, specifically in [version 2.3 (Errata D)](http://www.uefi.org/sites/default/files/resources/UEFI_Spec_2_3_D.pdf).

When iPXE is being chainloaded from a UEFI firmware performing a PXE boot in an IPv6 network, it is essential that iPXE supports IPv6 as well.

I understand that the reason for `NET_PROTO_IPV6` being disabled by default (in `src/config/general.h`) is that it would cause certain space-constrained build targets to become too large. However, this should not be an issue for EFI builds.

It is also worth noting that [RFC 6540](https://tools.ietf.org/html/rfc6540) makes a clear recommendation that IPv6 support should not be considered optional.

Signed-off-by: Tore Anderson <tore@fud.no>

### DIFF
--- a/src/config/general.h
+++ b/src/config/general.h
@@ -35,7 +35,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
  */
 
 #define	NET_PROTO_IPV4		/* IPv4 protocol */
-#undef	NET_PROTO_IPV6		/* IPv6 protocol */
+//#define	NET_PROTO_IPV6		/* IPv6 protocol */
 #undef	NET_PROTO_FCOE		/* Fibre Channel over Ethernet protocol */
 #define	NET_PROTO_STP		/* Spanning Tree protocol */
 #define	NET_PROTO_LACP		/* Link Aggregation control protocol */
--- a/src/config/defaults/efi.h
+++ b/src/config/defaults/efi.h
@@ -23,6 +23,8 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define REBOOT_EFI
 #define ACPI_EFI
 
+#define	NET_PROTO_IPV6		/* IPv6 protocol */
+
 #define DOWNLOAD_PROTO_FILE	/* Local filesystem access */
 
 #define	IMAGE_EFI		/* EFI image support */


src/config/general.h
```diff
@@ -35,7 +35,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
  */
 
 #define	NET_PROTO_IPV4		/* IPv4 protocol */
-#undef	NET_PROTO_IPV6		/* IPv6 protocol */
+//#define	NET_PROTO_IPV6		/* IPv6 protocol */
```
hjensas: Why is this changed to define and commented? Seems to me this should'nt be changed?

NiKiZe:
Since we got a new comment on this, and I wrote this before reading original date, I will post it anyway.
* First major one: code size, or rather binary size, you don't want it on a pcbios build with limited memory, unless you really need to. And there are still targets (ROMs) that are limited to 64Kb binaries, or in some cases even less. Virtual box for example is one such thing.
* Another reason is compability, unless your network is "properly" configured for IPv6, and you expect IPv4 to work, then you might see issues. feel free to read the mailing list

src/config/general.h
toreanderson: IIRC because this file (`src/config/general.h`) is included ***after*** `src/config/defaults/efi.h`. Thus, if the `#undef NET_PROTO_IPV6` is left unchanged here, it ends up overrides the newly added `#define NET_PROTO_IPV6` from `src/config/defaults/efi.h`.

Changing it to `//#define` follows the `IMAGE_EFI` example:
https://github.com/ipxe/ipxe/blob/7656bf6a333a63e1da3483cf0b2a814252f56161/src/config/general.h#L113
https://github.com/ipxe/ipxe/blob/7656bf6a333a63e1da3483cf0b2a814252f56161/src/config/defaults/efi.h#L30

This logic accomplishes making the define in question default enabled for the EFI platform, while at the same time leaving it default disabled for all other platforms (unless explicitly disabled for some other specific platforms too).

toreanderson:
>     * First major one: code size, or rather binary size, you don't want it on a pcbios build with limited memory, unless you really need to. And there are still targets (ROMs) that are limited to 64Kb binaries, or in some cases even less. Virtual box for example is one such thing.

Which is precisely the reason why the change I am proposing *only* enables IPv6 on EFI builds. PCBIOS builds are *not* changed.

>     * Another reason is compability, unless your network is "properly" configured for IPv6, and you expect IPv4 to work, then you might see issues. feel free to read the mailing list

As mentioned in the initial PR comment, IPv6 PXE support is *mandated* by the UEFI specification. This has been the case for over nine years at the point of writing.

If you have a network with some kind of defective IPv6 (non-)connectivity configured, then sure, you «might see issues» (most likely occurring before iPXE is even loaded). On networks with some kind of defective IPv4 (non-)connectivity, the statement is equally true.

It is however not the bootloader's role to fix broken networks, that's a task better left to the network operator. Especially when the attempt to fix the broken networks inevitably results in iPXE being 100% defective on *correctly* configured IPv6 network.

src/config/general.h
hjensas: ok, Thanks for clarifying. I was worried this would enable it for all binary builds.

toreanderson:
Hello maintainers. I believe enough time has passed to warrant a gentle nudge about this PR.

To reiterate the main reasons why it should be merged:

1. IPv6 PXE has already [part of the UEFI spec](http://www.uefi.org/sites/default/files/resources/UEFI_Spec_2_3_D.pdf) for over 12 years. It ought to work «out of the box», which this PR facilitates.
2. This PR only enables IPv6 by default for the EFI target. Concerns regarding space-constrained targets (e.g., PCBIOS) is therefore not relevant, as IPv6 remains disabled by default for those.
3. The updated logic for the `NET_PROTO_IPV6` define follows convention already established by other defines (such as `IMAGE_EFI`).

joshenders:
Is there a working rfc6555 (Happy Eyeballs) implementation in iPXE?

NiKiZe:
> Is there a working rfc6555 (Happy Eyeballs) implementation in iPXE?

Unfortunately not

PR #87, opened 2019-01-07
[infiniband] fix a memory leak in ib_refill_recv

secumod:
While debugging another issue (my UEFI firmware continuously polling the iPXE
option ROM on a Mellanox ConnectX-4 Lx) I noticed that iPXE quickly runs out of
memory. Tracing the code led me to ib_refill_recv. We can see that
ib_refill_recv allocates an iobuf and hands it over to a driver-specific
post-receive hook. However, ib_refill_recv free's the iobuf only, when the hook
returns an error and none of the implementations free the iobuf on success.

So it seems that ib_refill_recv should unconditionally free the iobuf before
processing the error code from the hook (this patch fixed the leak for me at
least).

### DIFF
--- a/src/net/infiniband.c
+++ b/src/net/infiniband.c
@@ -568,10 +568,11 @@ void ib_refill_recv ( struct ib_device *ibdev, struct ib_queue_pair *qp ) {
 		}
 
 		/* Post I/O buffer */
-		if ( ( rc = ib_post_recv ( ibdev, qp, iobuf ) ) != 0 ) {
+		rc = ib_post_recv ( ibdev, qp, iobuf );
+		free_iob ( iobuf );
+		if ( rc != 0 ) {
 			DBGC ( ibdev, "IBDEV %s could not refill: %s\n",
 			       ibdev->name, strerror ( rc ) );
-			free_iob ( iobuf );
 			/* Give up */
 			return;
 		}


ipxe-devel:
On 07/01/2019 13:19, Ignat Korchagin wrote:
> While debugging another issue (my UEFI firmware continuously polling the 
> iPXE
> option ROM on a Mellanox ConnectX-4 Lx) I noticed that iPXE quickly runs 
> out of
> memory. Tracing the code led me to ib_refill_recv. We can see that
> ib_refill_recv allocated an iobuf and hands it over to a driver-specific
> post-receive hook. However, ib_refill_recv free's the iobuf only, when 
> the hook
> returns an error and none of the implementations free the iobuf on success.
> 
> So it seems that ib_refill_recv should unconditionally free the iobuf before
> processing the error code from the hook (this patch fixed the leak for me at
> least).

ib_refill_recv() is refilling the receive ring with I/O buffers.  These 
buffers must remain allocated until the hardware consumes them, at which 
point the driver will reclaim the buffer from the receive ring and pass 
it to ib_complete_recv().

The buffer will be passed up the appropriate path through network stack 
and will eventually be freed by the final consumer (e.g. 
xferbuf_deliver(), if the data is part of a download to RAM).

Your patch is definitely incorrect since it will free the I/O buffer 
while leaving the memory in use by both the Infiniband hardware (for 
DMA) and the Infiniband driver.

I don't know what is causing your memory leak, but this isn't the right 
fix, sorry.

Michael

secumod:
Thanks for the detailed explanation.

Just more context:
- the memory leak is triggered by my server UEFI firmware: when loading the option ROM it finds that the ROM (iPXE in this case) implements an SNP interface, so it opens the interface and starts "polling" it in a tight loop. I don't know why it does that but suppose for similar reasons as in https://github.com/ipxe/ipxe/blob/e226fecd1b602ece7f4f208042f3e87ca540b297/src/interface/efi/efi_snp.c#L1767-L1776 (I actually patched it the same way by opening the SNP interface in DRIVER EXCLUSIVE mode as well)
- since the interface during this polling is not configured - obviously it gets nothing. However, I'm sure the memory is not released, so maybe the `ib_complete_recv()` never gets called

Closing this for now - I'll investigate further if I have time, but if you have some ideas/more pointers, please let me know.

PR #88, opened 2019-02-07
[com32] Fix iPXE stack pointer retrieval after COM32 binary returns.

dbankov-vmware:
This change fixes the offset used when retrieving the iPXE stack pointer
after a COM32 binary returns. The iPXE stack pointer is saved at the top
of the available memory then the the top of the stack for the COM32
binary is set just below it. However seven more items are pushed on the
COM32 stack before the entry point is invoked so when the COM32 binary
returns the location of the iPXE stack pointer is 28 (and not 24) bytes
above the current stack pointer.


mcb30:
Pushed as http://github.com/ipxe/ipxe/commit/70b1a641c; thank you!

PR #89, opened 2019-02-07
[util] Fix a potential problem in ROM.pm.

dbankov-vmware:
This change replaces "my $file_offset = shift // 0x0;" (which results in a
build failure in some environments) with "my $file_offset = scalar @_ ? shift : 0x0;" where
the intention is 0x0 to be the default value used if the optional
file_offset parameter is not provided.

### DIFF
--- a/src/util/Option/ROM.pm
+++ b/src/util/Option/ROM.pm
@@ -282,7 +282,7 @@ sub set {
   my $hash = shift;
   my $self = tied(%$hash);
   my $data = shift;
-  my $file_offset = shift // 0x0;
+  my $file_offset = scalar @_ ? shift : 0x0;
 
   # Store data
   $self->{data} = \$data;


pe-bo:
The Logical Defined-Or operator ( // ) was introduced to Perl 5.10 released on December 18th, 2007. I think it's a long enough time for it to be routinely used.

ipxe-devel:
The proposed patch mentions a _runtime_ error.


Cheers
Geert Stappers
Through E-mail, not sure if it ends up at github ...

dbankov-vmware:
> The proposed patch mentions a _runtime_ error.

I'm sorry this is not phrased correctly (I'll edit the description).

The problem is observed during the script execution which happens as part of the build process.

The result is a broken (make all) build on a build machine with perl 5.8.8.

I don't insist on merging this PR just thought I'd propose it.

stappersg:
> > The Logical Defined-Or operator ( // ) was introduced to Perl 5.10 released on December 18th, 2007.
> 
> The problem is observed during the script execution which happens as part of the build process.
> The result is a broken (make all) build on a build machine with perl 5.8.8.
> I don't insist on merging this PR just thought I'd propose it.

Consider closing this PR for the benefit of iPXE project.

The benefit for iPXE is no drain on resources spent on review the list of open pull requests.

dbankov-vmware:
Thanks for looking at this.

PR #91, opened 2019-02-23
parse response certID hashAlgorithm

stappersg:
Implements parsing of the hashAlgorithm  certID field of incoming OCSP
responses as well as validation of the field value against the respective
request field hashAlgorithm. The remaining fields of the response certID
are validated against the request via memcpy().

The certID part of the OCSP response is defined as an ASN.1 sequence:

CertID ::= SEQUENCE {
 hashAlgorithm           AlgorithmIdentifier,
 issuerNameHash          OCTET STRING, -- Hash of issuer's DN
 issuerKeyHash           OCTET STRING, -- Hash of issuer's public key
 serialNumber            CertificateSerialNumber }

(see https://tools.ietf.org/html/rfc6960#appendix-B.1 )

Parsing hashAlgorithm improves the previous implementation which
used memcmp() to compare the certID memory (raw bits) of request and
response. As ASN.1 semantics are ignored by memcmp(), bitwise different
but semantically identical certIDs were rejected by the previous
implementation. This caused e.g. boot failures of OS images downloaded
via HTTPS.

### DIFF
--- a/src/crypto/ocsp.c
+++ b/src/crypto/ocsp.c
@@ -474,18 +474,43 @@ static int ocsp_parse_responder_id ( struct ocsp_check *ocsp,
  */
 static int ocsp_parse_cert_id ( struct ocsp_check *ocsp,
 				const struct asn1_cursor *raw ) {
+	struct asn1_cursor req;
 	struct asn1_cursor cursor;
+	int rc;
+	struct asn1_algorithm *algo_req, *algo_resp;
 
-	/* Check certID matches request */
+	memcpy ( &req, &ocsp->request.cert_id, sizeof ( req ) );
 	memcpy ( &cursor, raw, sizeof ( cursor ) );
-	asn1_shrink_any ( &cursor );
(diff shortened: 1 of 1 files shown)


stappersg:
This is the commit I had in mind when I commented on #90 

* No white space changes
* Kept the variable name `cursor`  (needs to be replaced with `resp` in a later commit)
* Commit message is a commit message  ( No "This patch is ..." )

Side effect of this refresh pull request is that has the date of the actual change. Github dates https://github.com/ipxe/ipxe/pull/90 with an offset of 8 days.

This pull request waits for the sign-off by the original author  and possible further feedback.

t-lo:
Merged the changes into #90 , thank you very much for helping!

stappersg:
Reason for this _pull request_ was #90.

That _pr_ has been closed with http://git.ipxe.org/ipxe.git/commitdiff/b6ffe28a2 so now closing this _pr_ also.

PR #93, opened 2019-06-26
Circle CI added

lholota:


eworm-de:
Without looking at the code... You should rework your commit and give descriptive commit message instead of just "_Fix_".

PR #94, opened 2019-08-23
Updated digest commands to include SHA-2 support, hash strings, perfo…

chilinux:
This PR adds the following:

- SHA-2 family of digests (sha224sum, sha256sum, sha384sum and sha512sum)
- Multi-round rehashing support
- Hashing of strings (instead of only images)
- Apply result of the data last hashed to the configuration settings as hash:string
- Apply the length of the data last hashed to the configuration settings as hashlen:int32

It is also backward compatible with hashing one or more images as have been available in md5sum and sha1sum previous to the PR.  The major difference is it will write or overwrite hash and hashlen with the last of the multiple images successfully hashed.

Without this PR, the hash functions seem to only provide a way to manually verify the digest results.  With this the resulting digest string and the length of the image can both automatically be confirmed by the ipxe script.

Also consider the following situation, a company creates an ipxe efi module were they want it to go to an ipxe shell when DHCP fails.  However, the company policy requires a password be verified to provide the shell access and also requires the password not be easy to extract even if the efi module is examined with a debugger.

This PR would allow a script such as this to be used:

#!ipxe
dhcp || goto dhcpfail
# continue normal tasks
:dhcpfail
prompt DHCP has failed, press a key to login for shell access
:loginshell
login
iseq ${username:string} ipxeadmin || goto loginshell
set salt:string 3cbc63d0c64482c666d4dfb0516eed35
md5sum --rounds 1000000 --str ${salt:string}${password:string}
iseq ${hash:string} 1577df1969a3b8f0188496804645c406 || goto loginshell
shell

Please feel free to let me know if there is problem with my code or even if there is problems with the concept I am trying to achieve with the code.

Thanks

### DIFF
--- a/src/hci/commands/digest_cmd.c
+++ b/src/hci/commands/digest_cmd.c
@@ -29,6 +29,9 @@ FILE_LICENCE ( GPL2_OR_LATER );
 #include <ipxe/crypto.h>
 #include <ipxe/md5.h>
 #include <ipxe/sha1.h>
+#include <ipxe/sha256.h>
+#include <ipxe/sha512.h>
+#include <ipxe/settings.h>
 #include <usr/imgmgmt.h>
 
 /** @file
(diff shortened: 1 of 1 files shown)


Scarjit:
Would be incredible usefull for us.
We currently require the user to have two "passwords" for installation (one is send by iPXE to our matchbox server to select the correct config) and later we use the 2nd one to decrypt the config.

With this change we would be able to to use 1 password (hashed for selecting the correct config, and as it for decryption)

PR #97, opened 2019-10-05
[deflate] Add support for GZIP decompression

pe-bo:
GZIP file format (RFC 1952) uses DEFLATE algorithm with its
own header and footer. Implementation extends existing RAW
and ZLIB code. Also, a simple test was added.


oschroeder8:
Should this patch allow me to use/load gzipped kernel files directly?
For example, on XenServer/XCP-ng, the kernel file `/boot/xen.gz` has had to be ungzipped before being able to load it directly.
I merged your pull request into my local repo and rebuilt the undionly.kpxe file, but when attempting to load the gzipped file, iPXE responded with:
`Could not select: Exec format error (http://ipxe.org/2e008001)`
Any suggestions?

mcb30:
Alternative implementation merged in https://github.com/ipxe/ipxe/commit/866fa1ce7

PR #98, opened 2019-12-03
Expose DHCPv6 server-id, client-id and ia_na

hjensas:
These values and the ip6 address can be useful for chainloaded
OS to take over DHCP lease or do a DHCPRELEASE to ensure the
DHCP server does not return "no address available" if/when the
chainloaded OS attempts to configure the inerface using DHCPv6.

Signed-off-by: Harald Jensås <hjensas@redhat.com>


stappersg:
LGTM

PR #99, opened 2019-12-13
[snp] try to enable promiscuous multicast receive filter if the regular one fails

ignatk:
Currently, if the snp driver for whatever reason fails to enable receive filters
for multicast frames, it falls back to enabling just unicast and broadcast
filters. This breaks some IPv6 functionality as the network card does not
respond to neighbour solicitation requests.

Some cards refuse to enable EFI_SIMPLE_NETWORK_RECEIVE_MULTICAST filter, but
support enabling EFI_SIMPLE_NETWORK_RECEIVE_PROMISCUOUS_MULTICAST, so try it
before falling back to just unicast+broadcast.

Additionally, according to UEFI specification 2.8 p 24.1 we must set
EFI_SIMPLE_NETWORK_RECEIVE_MULTICAST bit in the "Disable" mask, when
"ResetMCastFilter" is TRUE.


mcb30:
Good catch; thank you.  I have split this into two separate commits and merged to master.

PR #101, opened 2019-12-27
[efi] Don't use timers to gather entropy, if EFI_RNG_PROTOCOL is available

ignatk:
When iPXE is compiled as an option ROM (I tried the e1000 driver), it causes a
complete boot stall on the latest release of OVMF for EDK2 [1], run on the
latest QEMU 4.2.0. Further debugging traced it down to commit d8c500b ("[efi]
Drop to TPL_APPLICATION when gathering entropy"). Specifically, somehow the
whole boot process stalls, when trying to RaiseTPL() back to TPL_CALLBACK in
efi_entropy_disable().

I'm also not actually sure the code in efi_entropy_enable() /
efi_entropy_disable() actually runs at TPL_CALLBACK (at least for the first
time) as I dumped the current TPL, when the first efi_entropy_enable() call was
made and it showed iPXE was running at TPL_APPLICATION level.

While the whole idea of constantly running at elevated task priority level is
rather questionable for me, proposing to fix at least the entropy gathering by
not messing with timers and priority levels at all, if we have EFI_RNG_PROTOCOL
provided.

[1] https://github.com/tianocore/edk2

### DIFF
--- a/src/interface/efi/efi_entropy.c
+++ b/src/interface/efi/efi_entropy.c
@@ -79,16 +79,18 @@ static int efi_entropy_enable ( void ) {
 	DBGC ( &tick, "ENTROPY %s RNG protocol\n",
 	       ( efirng ? "has" : "has no" ) );
 
-	/* Drop to TPL_APPLICATION to allow timer tick event to take place */
-	bs->RestoreTPL ( TPL_APPLICATION );
+	if ( ! efirng ) {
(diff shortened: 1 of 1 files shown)


mcb30:
Thank you for the detailed report.  Have fixed the underlying root cause in commit 2ae5d4338

PR #102, opened 2020-01-10
QRcode

dgtlrift:
This patch is only a proof of concept but links in QR encode and renders the URLs for stderror.c to the console for a smartphone to scan for additional context.

A few considerations:
  Additional Memory requirements
  License for the library
  Readability from non-console screens
  stability
  Dynamic memory requirements (malloc) for the library
  memory optimizations in the library - bit array instead of bytes?

I don't warrant that this is production-ready, but merely trying to show that this could provide additional value.

I did try to subscribe to ipxe-devel@lists.ipxe.org and submit a patch, but it appears my requrest to join the list got dropped to the floor.
-Jim


dgtlrift:
Not sure what happened - this builds fine locally - I'll try to understand the issue on from the logs.

dch:
Interesting but what’s the use case? Can you clarify how having a qrcode would be used in practice during a server or laptop boot sequence?

ipxe-devel:
On Wed, Jan 15, 2020 at 08:31:24AM -0800, Dave Cottlehuber wrote:
> Interesting but what’s the use case? Can you clarify how having a qrcode would be used in practice during a server or laptop boot sequence?
> 

My understanding is that it'll make it easier for you to go to the error
page in ipxe.org instead of having to manually type the url. I think
it's a cool feature...

Thanks,
-Santiago

dgtlrift:
image
Yes the idea is that the QR Code would reflect the issue URL for quick scanning

src/config.h
```diff
@@ -0,0 +1,148 @@
+/* config.h.  Generated from config.h.in by configure.  */
```
stappersg, line 1: Please,  do not put generated files under source code version control.

src/Makefile
```diff
@@ -128,11 +130,24 @@ INCDIRS		+= include .
 # Default build target: build the most common targets and print out a
 # helpfully suggestive message
 #
-ALL		:= bin/blib.a bin/ipxe.dsk bin/ipxe.lkrn bin/ipxe.iso \
-		   bin/ipxe.usb bin/ipxe.pxe bin/undionly.kpxe bin/rtl8139.rom \
-		   bin/8086100e.mrom bin/80861209.rom bin/10500940.rom \
-		   bin/10222000.rom bin/10ec8139.rom bin/1af41000.rom \
-		   bin/8086100f.mrom bin/808610d3.mrom bin/15ad07b0.rom
+ALL	:=
```
stappersg, line 133: This idea must go into a seperate patch.
(It does NOT belong to this patch)

NiKiZe:
This is a cool feature!
For any chance to merge it needs to only have intended changes and nothing else, and clean commits.
This means that it needs to be rebased and cleaned up to reviewable commits.

Also depending on or including external code will not be an option. License is one concern.
Maybe loading that separately and enabling the options could be possible, but that might not be viable in terms of [UBDL](https://github.com/ipxe/ipxe/blob/master/COPYING.UBDL)

dgtlrift:
> This is a cool feature!
> 
> For any chance to merge it needs to only have intended changes and nothing else, and clean commits.
> 
> This means that it needs to be rebased and cleaned up to reviewable commits.
> 
> 
> 
> Also depending on or including external code will not be an option. License is one concern.
> 
> Maybe loading that separately and enabling the options could be possible, but that might not be viable in terms of [UBDL](https://github.com/ipxe/ipxe/blob/master/COPYING.UBDL)

I would love to merge it in, but my day job has me overcapacity. That said just having the infrastructure of representing a QR code on screen for other applications has caught my eye...

Specifically, for USB boot iPXE and the system has WiFi but no Ethernet, a QR-Code for Wi-Fi EasyConnect (aka Device Provisioning Protocol - https://www.wi-fi.org/download.php?file=/sites/default/files/private/Wi-Fi_Easy_Connect_Specification_v2.0.pdf ) could be used with a Configurator (right now only supported with Google Pixel) to onboard the system to the network. In the case of ones with Ethernet, it can be used with the Supplicant if using EAP to authenticate the device on the port. For WiFi EasyConnect - it includes band and channel information with a public key associated with the device.

steverweber:
neat/ I started to look at generating a QRcode server side and passing it down in a chain loaded ipxe script in a retry loop. The QRcode brings user to a webpage to enrol the system. Looks like i'll just continue with my hack for now.

dgtlrift:
> neat/ I started to look at generating a QRcode server side and passing it down in a chain loaded ipxe script in a retry loop. The QRcode brings user to a webpage to enrol the system. Looks like i'll just continue with my hack for now.

I'm getting off topic, but I would strongly recommend taking a look at DPP (device provisioning protocol) mentioned above - first it's ratified as a standard by the Wi-Fi Alliance (albeit an optional feature) but it preferred over PBP (push button provisioning) and WPS (Wi-Fi protected setup) because of the UX, security, and failure mode handling, support has been added to the w1.fi wpa_supplicant project, it supported by Android (albeit current only with the google pixel.)
It is a standard secure way to onboard devices and it's only a matter of time before all the Wi-Fi transceivers in Androids and iPhones support the configurator mode information elements of 802.11 needed for DPP.

https://youtu.be/sEAlAT3-sBM a demo of onboarding a device with no UI using DPP.

https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-reference/network/esp_dpp.html IoT Wi-Fi device supporting DPP

steverweber:
Thanks for pointing out DPP, I got deeper into this project last night and found that I cant use the nice ASCII chars 219,220,254 when using ipxe under efi mode booting from vmware EXI. Seems to be in a Unicode mode.
```
set d:int8 220
echo ${d:string}
Ü
```

steverweber:
Seems that console in ipxe in EFI mode is using UTF-16 but im not sure how to echo custom UTF16 char.

NiKiZe:
@steverweber ASCII is 7bit, = max char 127, Extended ASCII is not really defined here, BIOS mode is often CP437 but no guarantees, iPXE just outputs puts bytes on the output buffer, then BIOS/EFI is responsible for how that is displayed. (From what I understood EFI is always UCS-2)

steverweber:
@NiKiZe Thanks!
I tried a few things but it seems that im unable to get the UCS-2 or 2byte char working for the ISO10646 charset.

```
set h1:hex 00
set h2:hex dc
set d:int16 220
echo ${h1string}${h2string} ${d:string}
```

/shrug/

mcb30:
> @NiKiZe Thanks! I tried a few things but it seems that im unable to get the UCS-2 or 2byte char working for the ISO10646 charset.

iPXE is currently ASCII-only.  I am currently working on adding Unicode support in stages.  The first part (allowing the use of UTF-8 in iPXE scripts, allowing for statements such as `echo Héllö wõrld`) is working but not yet merged to master.  Unicode characters via keyboard input will be added in a subsequent stage.

steverweber:
I would think that using echo with the raw bytes would work. Hope your work continues to progress. Thanks!

stappersg:
There was [some change](https://github.com/ipxe/ipxe/pull/102/files) which says _make this a separate patch_
but [https://github.com/ipxe/ipxe/pull/102/files](https://github.com/ipxe/ipxe/pull/102/files)
still shows changes **un**related to _QRcode_.

afbeelding

Request to @dgtlrift   please provide clean patches, so only related changes in one merge request.

dgtlrift:
> There was [some change](https://github.com/ipxe/ipxe/pull/102/files) which says _make this a separate patch_
> 
> but [https://github.com/ipxe/ipxe/pull/102/files](https://github.com/ipxe/ipxe/pull/102/files)
> 
> still shows changes **un**related to _QRcode_.
> 
> 
> 
> afbeelding
> 
> 
> 
> Request to @dgtlrift   please provide clean patches, so only related changes in one merge request.

Copy-pasta is killing me with tabs/spaces

NiKiZe:
This is a nice feature, but pulling in external libraries etc is a no-go.
QR support must be built in. if scope is limited this shouldn't be impossible.
If there is work progressing on this in the future we can reopen, but closing for now.

PR #104, opened 2020-01-16
[dhcp/ipv4] Support for DHCP option 121: Classless static route option (follow-up)

RaphaelPour:
This PR adds support for classless static route option (https://tools.ietf.org/rfc/rfc3442.txt) and is a follow-up of PR #67.
I refactored it using @multun s [review](https://github.com/ipxe/ipxe/pull/67#pullrequestreview-116469263).

Signed-off-by: Raphael Pour <info@raphaelpour.de>

### DIFF
--- a/src/include/ipxe/netdevice.h
+++ b/src/include/ipxe/netdevice.h
@@ -426,6 +426,12 @@ struct net_device {
 
 	/** Network device configurations (variable length) */
 	struct net_device_configuration configs[0];
+
+	/** Static routes from DHCP option 121 */
+	uint8_t * static_routes;
+
+	/** DHCP option 121 length */
+	uint8_t static_routes_len;
 };
 
 /** Network device is open */
--- a/src/net/ipv4.c
+++ b/src/net/ipv4.c
@@ -337,9 +337,12 @@ static int ipv4_tx ( struct io_buffer *iobuf,
 	if ( ( next_hop.s_addr != INADDR_BROADCAST ) &&
 	     ( ( miniroute = ipv4_route ( sin_dest->sin_scope_id,
 					  &next_hop ) ) != NULL ) ) {
-		iphdr->src = miniroute->address;
 		netmask = miniroute->netmask;
 		netdev = miniroute->netdev;
+
+        struct settings *settings = netdev_settings ( netdev );
+        fetch_ipv4_setting ( settings, &ip_setting, &iphdr->src );
+        next_hop = miniroute->gateway;
 	}
 	if ( ! netdev ) {
 		DBGC ( sin_dest->sin_addr, "IPv4 has no route to %s\n",
@@ -448,16 +451,15 @@ int ipv4_has_any_addr ( struct net_device *netdev ) {
  * @ret has_addr	Network device has this IPv4 address
  */
 static int ipv4_has_addr ( struct net_device *netdev, struct in_addr addr ) {
-	struct ipv4_miniroute *miniroute;
 
-	list_for_each_entry ( miniroute, &ipv4_miniroutes, list ) {
-		if ( ( miniroute->netdev == netdev ) &&
-		     ( miniroute->address.s_addr == addr.s_addr ) ) {
-			/* Found matching address */
-			return 1;
-		}
-	}
-	return 0;
+    struct in_addr address;
+    struct settings *settings = netdev_settings ( netdev );
+    fetch_ipv4_setting ( settings, &ip_setting, &address );
+
+    if ( address.s_addr == addr.s_addr )
+         return 1;
+
+    return 0;
 }
 
 /**
@@ -845,6 +847,36 @@ static int ipv4_gratuitous_arp ( struct net_device *netdev,
 }
 
 /**
+ * Convert netmask from CIDR notation to the common x.x.x.x notation
+ */
+static inline uint8_t cidr_to_mask ( uint8_t cidr, char * buf ) {
+    uint8_t addr_len = 0;
+    uint8_t n_octet = ( cidr % 8 == 0 ? 255 :
+            ( 255 - ( ( 256 >> ( cidr % 8 ) ) - 1 ) ) );
+
+    if (cidr > 24 && cidr <= 32) {
+        sprintf(buf, "255.255.255.%d", n_octet);
+        addr_len = 4;
+    } else if (cidr > 16) {
+        sprintf(buf, "255.255.%d.0", n_octet);
+        addr_len = 3;
+    } else if (cidr > 8) {
+        sprintf(buf, "255.%d.0.0", n_octet);
+        addr_len = 2;
+    } else if (cidr > 0) {
+        sprintf(buf, "%d.0.0.0", n_octet);
+        addr_len = 1;
+    } else if (cidr == 0) {
+        sprintf(buf, "0.0.0.0");
+        addr_len = 0;
+    } else {
+        DBGC(cidr, "Unknown netmask descriptor /%d.\n", cidr);
+    }
+
+    return addr_len;
+}
+
+/**
  * Process IPv4 network device settings
  *
  * @v apply		Application method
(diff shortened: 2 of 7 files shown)


TomSDEVSN:
Can also confirm the functionality under ESXi and works perfectly in our production environment.

SeanMooney:
i have not tested this but i recently had a need for this fucntionality for ironic and openstack.
speficically i had static route defined in my openstack network ot make the tftp server routable which workded fine to retrive the ipxe image but the ipxe image then could not use the same static routes to retrive the ironic ram disk. having support for static routes would have resovled that.  i am working around this by moveing the api/ipx/tftp server to the same subnet so that the static routes are not needed.

konstantin-kelemen:
I've tested the code and can confirm it working.

Although there was an issue building it with the latest gcc version which was giving me zero-length-bounds errors, but it was pretty easy to fix:

```
--- a/src/include/ipxe/netdevice.h	2021-05-11 13:37:37.000000000 +0300
+++ b/src/include/ipxe/netdevice.h	2021-05-10 12:54:37.000000000 +0300
@@ -430,14 +430,14 @@
 	/** Driver private data */
 	void *priv;

-	/** Network device configurations (variable length) */
-	struct net_device_configuration configs[0];
-
 	/** Static routes from DHCP option 121 */
 	uint8_t * static_routes;

 	/** DHCP option 121 length */
 	uint8_t static_routes_len;
+
+	/** Network device configurations (variable length) */
+	struct net_device_configuration configs[0];
 };

 /** Network device is open */
```

Can @mcb30 and @tonobo please review and merge this PR?

saschaarthur:
so was it merged? is it on production?

saschaarthur:
related to https://github.com/ipxe/ipxe/pull/78

RaphaelPour:
> so was it merged? is it on production?

Nope and nope. AFAIK did this PR never got any attention to the maintainer/s since I created it in early '20.

sedlund:
This works for me on Hetzner cloud.  I had to move

> 	struct net_device_configuration configs[0];

to the end of the struct in `netdevice.h` to avoid the new gcc warnings.  Would be great to get this merged!

spike77453:
Can also confirm that it works. Tested on qemu. Would love to see this get merged.

spike77453:
If I change my isc dhcp configuration to

```
option rfc3442-classless-static-routes 24,172,31,254,192,168,178,243, 0,192,168,178,253;
option ms-classless-static-routes      24,172,31,254,192,168,178,243, 0,192,168,178,253;
```

(i.e. handing out the default route as the last five bytes in the option rather than the first five), this happens:

Screenshot_2022-01-28_16:10:32

Note the messed up routing table. So there's something fishy going on, I just can't quite put my finger to the relevant line of code :(

src/net/ipv4.c
```diff
@@ -890,9 +912,52 @@ static int ipv4_settings ( int ( * apply ) ( struct net_device *netdev,
...
+			while ( i < option_len ) {
+				prefix_len = raw_data[i];
+				i++;
+
+				/* Netmask */
+				addr_octets_len = prefix_len / 8;
+				netmask.s_addr = prefix_len_to_mask ( prefix_len );
+
+				/* Address */
+				for ( j = 0; j < addr_octets_len; j++ )
```
spike77453, line 941: This won't work if a default route is handed out in any position but the first within the option. Allow me to explain:

Let's say the dhcp server hands out a route `172.31.254.0/24 gw 192.168.178.243` followed by a default route (5 bytes, `0` plus the gateway).  
After processing the first 8 bytes of the option (i.e. `18 ac 1f fe c0 a8 b2 f3` which is the route `172.31.254.0/24 gw 192.168.178.243`), `address_conversion.octets` will contain data. But the 9th byte is going to be 0 (the network/mask part of the default route) so `prefix_len` is 0, so `addr_octets_len` is 0 and we skip the for loop, never chaning the data in `address_conversion.octets`.

So instead of `address` being `0.0.0.0`, it's going to be the stale value from the route processed before (`172.31.254.0` in this example)

src/net/ipv4.c
```diff
@@ -337,9 +337,12 @@ static int ipv4_tx ( struct io_buffer *iobuf,
 	if ( ( next_hop.s_addr != INADDR_BROADCAST ) &&
 	     ( ( miniroute = ipv4_route ( sin_dest->sin_scope_id,
 					  &next_hop ) ) != NULL ) ) {
-		iphdr->src = miniroute->address;
 		netmask = miniroute->netmask;
 		netdev = miniroute->netdev;
+
+        struct settings *settings = netdev_settings ( netdev );
+        fetch_ipv4_setting ( settings, &ip_setting, &iphdr->src );
```
mcb30, line 344: This code is on the transmit datapath, which is latency sensitive and cannot therefore perform operations such as fetching settings.

src/net/ipv4.c
```diff
@@ -448,16 +451,15 @@ int ipv4_has_any_addr ( struct net_device *netdev ) {
...
-		if ( ( miniroute->netdev == netdev ) &&
-		     ( miniroute->address.s_addr == addr.s_addr ) ) {
-			/* Found matching address */
-			return 1;
-		}
-	}
-	return 0;
+    struct in_addr address;
+    struct settings *settings = netdev_settings ( netdev );
+    fetch_ipv4_setting ( settings, &ip_setting, &address );
```
mcb30, line 457: This code is on the receive datapath, which is latency sensitive and cannot therefore perform operations such as fetching settings.

src/net/udp/dhcp.c
```diff
@@ -567,6 +568,27 @@ static void dhcp_request_rx ( struct dhcp_session *dhcp,
...
+            /* Neither option 121 found nor 249 */
+            dhcp->netdev->static_routes_len = 0;
+        }
+    }
+    if ( offset_opt121 > 0 ) {
+        struct dhcp_option *d_option =
+                dhcp_option ( &dhcppkt->options, offset_opt121 );
+        DBGC ( dhcp, "DHCP option 121 length: %d\n", d_option->len );
+        dhcp->netdev->static_routes_len = d_option->len;
+        dhcp->netdev->static_routes = ( ( uint8_t * ) d_option->data );
```
mcb30, line 589: You can't record a long-lived pointer to this structure here, because you have nothing that guarantees that the thing pointed to will remain valid.

src/include/ipxe/netdevice.h
```diff
@@ -426,6 +426,12 @@ struct net_device {
 
 	/** Network device configurations (variable length) */
 	struct net_device_configuration configs[0];
+
+	/** Static routes from DHCP option 121 */
+	uint8_t * static_routes;
+
+	/** DHCP option 121 length */
+	uint8_t static_routes_len;
```
mcb30, line 434: You cannot put IPv4-specific fields in the protocol-independent `struct net_device`.

Also, the variable-length `configs[0]` field must be at the end of the structure, as with all variable-length fields in C.

mcb30 (changes requested):
Thank you for the suggestion.  This patch would unfortunately require a complete rewrite.  The most fundamental issue is that fast-path operations (such as network transmit and receive) are not allowed to perform slow operations such as fetching settings.  This is why the settings applicator mechanism exists: when settings are changed the `ipv4_settings()` function parses them to build the routing table, which is then consulted in the transmit (and receive) datapaths.

I have made several other comments inline.

Please be aware that there are also multiple stylistic and correctness issues with the code: there is no point in my reviewing these until the overall architecture is essentially correct, but you should be prepared for at least one more rewrite after this one.

spike77453:
TWIMC: #78 also adds support for static routes and seems to be functional (probably not quite ready to merge yet)

RaphaelPour:
Thanks for the honest review and sorry for late response. 

My understanding doesn't go deep enough in order to rewrite the PR, so I've closed it. If anyone wants to give it a shot, feel free to pick my PR up.

mcb30:

PR #106, opened 2020-02-21
Setup INT13 partition on usb images only ifdef CONSOLE_INT13 

Robien:
In config/console.h, the option CONSOLE_INT13 enables log on the USB key.
To do that a 1Mb partition has been added to USB images.
If the option is disabled then the partition is still there and do nothing.

This PR create the partition only when the option is enabled. Thus, it reduce the size of usb images by 1Mb when the option is not activated (Which is the default)

### DIFF
--- a/src/arch/x86/prefix/usbdisk.S
+++ b/src/arch/x86/prefix/usbdisk.S
@@ -1,5 +1,7 @@
 FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL )
 
+#include <config/console.h>
+
 	.text
 	.arch i386
 	.section ".prefix", "awx", @progbits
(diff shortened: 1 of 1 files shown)


ipxe-devel:
The patch feels wrong,  surely **no** _Looks Good To Me_.

Robien:
> The patch feels wrong, surely **no** _Looks Good To Me_.

Can you elaborate a bit so I can make it better? Do you think I should add an option?

I should have also say that the code I am adding here comes from an older version of the file where CONSOLE_INT13 wasn't implemented yet (so no new code is introduced here)

We are using ipxe inside a very small device and we need that extra 1Mb. I understand that is not very important for the majority of users but this patch doesn't take anything to anyone.

stappersg:
> Can you elaborate a bit so I can make it better?

That became https://lists.ipxe.org/pipermail/ipxe-devel/2020-February/006948.html
You are welcome at  the iPXE mailinglist for further discussion.

mcb30:
Reimplemented as https://github.com/ipxe/ipxe/commit/efc1ae5abac01a77c5816884612c9daf0c344e8a

Thanks,

Michael

PR #107, opened 2020-03-06
[serial] Allow serial port configuration at runtime

paravoid:
Add configuration variables for the serial port and its options,
allowing the configuration of a serial port at runtime via either "set",
or DHCP options.

The new "serial-port" option accepts values of "1" through "4" for COM1
to COM4 respectively. For the serial options, rather than adding a
configuration variable for each of them (baud, parity etc.) and
complicate both the code and the configuration unncessarily, add a
single, generic "serial-options" setting that accepts string values such
as "115200" or "38400n81". It also accepts the special value "preserve"
to preserve the previous configuration of the serial port.

Note that this commit does _not_ enable CONSOLE_SERIAL by default,
although that could potentially follow in a subsequent commit, given
this would be now a no-op unless a port is explicitly configured.

### DIFF
--- a/src/config/serial.h
+++ b/src/config/serial.h
@@ -5,27 +5,21 @@
  *
  * Serial port configuration
  *
- * These options affect the operation of the serial console.  They
- * take effect only if the serial console is included using the
- * CONSOLE_SERIAL option.
+ * These options affect the operation of the serial console. They take effect
+ * only if the serial console is included using the CONSOLE_SERIAL option.
+ * These settings only the default settings, and can all be set (or overriden)
+ * at runtime.
  *
  */
 
 FILE_LICENCE ( GPL2_OR_LATER );
 
-#define	COMCONSOLE	COM1		/* I/O port address */
-
-/* Keep settings from a previous user of the serial port (e.g. lilo or
- * LinuxBIOS), ignoring COMSPEED, COMDATA, COMPARITY and COMSTOP.
- */
-#undef	COMPRESERVE
-
-#ifndef COMPRESERVE
-#define	COMSPEED	115200		/* Baud rate */
-#define	COMDATA		8		/* Data bits */
-#define	COMPARITY	0		/* Parity: 0=None, 1=Odd, 2=Even */
-#define	COMSTOP		1		/* Stop bits */
-#endif
+//#define COMCONSOLE	COM1	/* I/O port address */
+#define COMPRESERVE	0	/* Preserve settings by another bootloader */
+#define COMSPEED	115200	/* Baud rate */
+#define COMDATA		8	/* Data bits */
+#define COMPARITY	0	/* Parity: 0=None, 1=Odd, 2=Even */
+#define COMSTOP		1	/* Stop bits */
 
 #include <config/named.h>
 #include NAMED_CONFIG(serial.h)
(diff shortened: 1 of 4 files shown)


ondrejj:
I like this, tested on my boot.salstar.sk, works well.
@mcb30, can you please pull this upstream? Thanks.

paravoid:
Bump! I first submitted this patch to the mailing list a little bit over a year ago -- it'd be nice to see a review and/or merge before it completely fades out of memory :) Thanks a lot!

bofhbug:
@mcb30, is something missing? If not can you please pull this upstream?

paravoid:
Bump :) This was first posted in ipxe-devel on Jun 15, 2019 - it'd be great to see this mainlined. For what it's worth, I just tried rebasing this to master, and seems to apply cleanly. Thanks!

paravoid:
OK, discussed this on IRC with @mcb30. Changes that need to happen in this one are:
- Revert the change where `COMCONSOLE` was commented-out; default `COMCONSOLE` to `COM1` if `CONSOLE_SERIAL` is enabled, as anything else would break backwards compatibility
- While re-adding the above, add a comment that mentions that `COMCONSOLE` can be undef, if one does not wish to use a default build-time port
- Readd early init (via an `init_fn` function), as that helps with early boot debugging and is a regression compared to the current code. Ideally do this in a way that is not replicating the code and blowing up code size too much (the settings code should just gracefully no-op before any settings are registered, so it may be possible to reuse the settings applicator modulo return type - should verify this).
- Finally, remove options parsing, and replace it by a `serial-baud` setting that justs sets the baud rate. Rationale: non-n81 systems are rare (if existent at all?) in the wild, and options parsing is costly in terms of code size, so it does not make sense from a cost/benefit perspective.

There was another stylistic change discussed, as to whether serial port/baud rate are better configured via a setting, or via the console command, but ended up being OK with the former, with a tiny doc edit on the console cmd to clarify it's only meant for graphical/framebuffer consoles.

Hope I captured everything, and appreciate the feedback a lot!

paravoid:
With 2ce1b185b239b220cc6c1847f7014df83b1a8862, 2ce1b185b239b220cc6c1847f7014df83b1a8862 etc. both the code and the logic itself will certainly require and adjustment as platforms now provide their own default port and baud rate.

I still think the the original goal of this PR -a runtime-configurable port-  was useful to have at the time and probably still is to some folks. However, it's been ~6 years since I first posted this and my interests have shifted elsewhere.

@mcb30 you may want to close this. Or reimplement it since you've been touching the surroundings of thise code relatively recently! In any case, thank you for tirelessly maintaining iPXE all these years, it's great to see the project is still alive and kicking!

mcb30:
Thank you.  It would be interesting to look at this again.  Commit https://github.com/ipxe/ipxe/commit/60e167c00 simplified iPXE's internal model for a serial port so that the only configurable parameter (besides the port address) is the baud rate.  We do have some ability to select the serial port at runtime via ACPI SPCR (in commit https://github.com/ipxe/ipxe/commit/f2ea97102): this should cause iPXE to match whichever serial port the BIOS thinks is supposed to be used for console redirection.

I will try to remember to look at this feature the next time I get back to playing with RISC-V bare metal support, since that platform already ends up instantiating multiple serial ports.

PR #111, opened 2020-04-14
[efi] Add support for ${boot-vlan} setting

dbankov-vmware:
Introduce ${boot-vlan} setting exposing the VLAN from the EFI loaded
image device path for iPXE. This setting enables iPXE-boot on UEFI based
machines using the VLAN configured for the PXE adapter.

An example usage of the new setting is presented in the following iPXE
script:

      #!ipxe
      iseq ${boot-vlan} 0 || vcreate -t ${boot-vlan} net0
      autoboot

### DIFF
--- a/src/include/ipxe/efi/efi_utils.h
+++ b/src/include/ipxe/efi/efi_utils.h
@@ -16,6 +16,9 @@ struct device;
 extern EFI_DEVICE_PATH_PROTOCOL *
 efi_devpath_end ( EFI_DEVICE_PATH_PROTOCOL *path );
 extern size_t efi_devpath_len ( EFI_DEVICE_PATH_PROTOCOL *path );
+extern EFI_STATUS efi_devpath_find_node ( EFI_DEVICE_PATH_PROTOCOL *path,
+					  UINT8 type, UINT8 sub_type,
+					  EFI_DEVICE_PATH_PROTOCOL **node );
 extern int efi_locate_device ( EFI_HANDLE device, EFI_GUID *protocol,
 			       EFI_HANDLE *parent );
 extern int efi_child_add ( EFI_HANDLE parent, EFI_HANDLE child );
--- a/src/interface/efi/efi_init.c
+++ b/src/interface/efi/efi_init.c
@@ -22,7 +22,9 @@ FILE_LICENCE ( GPL2_OR_LATER );
 #include <string.h>
 #include <errno.h>
 #include <ipxe/init.h>
+#include <ipxe/settings.h>
 #include <ipxe/efi/efi.h>
+#include <ipxe/efi/efi_utils.h>
 #include <ipxe/efi/efi_driver.h>
 #include <ipxe/efi/Protocol/LoadedImage.h>
 
@@ -46,6 +48,9 @@ int efi_shutdown_in_progress;
 
 /** Event used to signal shutdown */
 static EFI_EVENT efi_shutdown_event;
+
+/** EFI loaded image VLAN **/
+static UINT16 efi_boot_vlan = 0;
 
 /* Forward declarations */
 static EFI_STATUS EFIAPI efi_unload ( EFI_HANDLE image_handle );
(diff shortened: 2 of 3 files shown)


skoef:
In order to UEFI boot on a specific VLAN, I backported your patch to current master (which has some parts of your patch already more or less implemented) and I do get the correct boot-vlan exposed. However, when I run `vcreate -t 100 net0` (regardless of whether I do this in the patched or upstream version) I get an error. After debugging enabled I see:

```
SNPDEV 0x50186408 could not install protocols: Error 0x711fe094 (http://ipxe.org/err/7f11e094)
VLAN net0-100 could not register: Error 0x711fe094 (http://ipxe.org/err/7f11e094)
Could not create VLAN device: Error 0x711fe094 (http://ipxe.org/err/7f11e094)
```

I'm running this on a PowerEdge R640 with recent BIOS. Any idea why registering fails?

skoef:
For the record, the error is generated at https://github.com/ipxe/ipxe/blob/master/src/interface/efi/efi_snp.c#L1722

NiKiZe:
> In order to UEFI boot on a specific VLAN, I backported your patch to current master (which has some parts of your patch already more or less implemented) and I do get the correct boot-vlan exposed. However, when I run `vcreate -t 100 net0` (regardless of whether I do this in the patched or upstream version) I get an error. After debugging enabled I see:
> 
> ```
> SNPDEV 0x50186408 could not install protocols: Error 0x711fe094 (http://ipxe.org/err/7f11e094)
> VLAN net0-100 could not register: Error 0x711fe094 (http://ipxe.org/err/7f11e094)
> Could not create VLAN device: Error 0x711fe094 (http://ipxe.org/err/7f11e094)
> ```
> 
> I'm running this on a PowerEdge R640 with recent BIOS. Any idea why registering fails?

Output of ifstat (so we see driver)? Might be device specific.
Does it work to create vlan for you with current master as-is?

skoef:
With current master, it doesn't work either, so it's not this PR at all messing things up. Output of ifstat:

```
net0: 0c:42:a1:9e:f3:54 using ConnectX-4Lx on 0000:18:00.0 (Ethernet) [closed]
  [Link:down TX:0 TXE:1 RX:0 RXE:0]
  [Link status: Unknown (http://ipxe.org/1a086194)
  [TXE: 1 x "Network unreachable (http://ipxe.org/28086090)"]
net1: 0c:42:a1:9e:f3:55 using ConnectX-4Lx on 0000:18:00.1 (Ethernet) [closed]
  [Link:down TX:0 TXE:0 RX:0 RXE:0]
  [Link status: Unknown (http://ipxe.org/1a086194)
```

However, I am able to `vcreate -t 100 net1` without any problem. So it doesn't seem driver related. The only difference between net0 and net1 is the VLAN tag in the BIOS I configured for net0. Could it have to do something with the fact that net0 is already implicitly a VLAN device so it can't be tagged *again*?

NiKiZe:
> With current master, it doesn't work either, so it's not this PR at all messing things up. Output of ifstat:
> 
> ```
> net0: 0c:42:a1:9e:f3:54 using ConnectX-4Lx on 0000:18:00.0 (Ethernet) [closed]
>   [Link:down TX:0 TXE:1 RX:0 RXE:0]
>   [Link status: Unknown (http://ipxe.org/1a086194)
>   [TXE: 1 x "Network unreachable (http://ipxe.org/28086090)"]
> net1: 0c:42:a1:9e:f3:55 using ConnectX-4Lx on 0000:18:00.1 (Ethernet) [closed]
>   [Link:down TX:0 TXE:0 RX:0 RXE:0]
>   [Link status: Unknown (http://ipxe.org/1a086194)
> ```
> 
> However, I am able to `vcreate -t 100 net1` without any problem. So it doesn't seem driver related. The only difference between net0 and net1 is the VLAN tag in the BIOS I configured for net0. Could it have to do something with the fact that net0 is already implicitly a VLAN device so it can't be tagged _again_?

You should probably create a new discussion since it's not related to this PR, can you try without vlan at boot? Also check existing issues for vlan (including closed ones) might be that similar reports already exists.

PR #112, opened 2020-05-05
Add an option to disable TLS fragmentation

chschenk:
iPXE currently does not support TLS connections with large certificate chains because it can not handle TLS handshake record fragmentation. So i think there should be an option to disable the request for fragmentation. Hence i would add a `confg/tls.h` to make changes to the behaviour of the tls implementation. By default the request for fragmentation is enabled, so nothing changes here, but if desired it can be turned of by undefining `TLS_FRAGMENTATION_ENABLED`
I also added the option `TLS_REQUESTED_MAX_FRAGMENT_LENGTH` for defining, if fragmentation is enabled, the requested maximum fragment length.

### DIFF
--- a/src/config/tls.h
+++ b/src/config/tls.h
@@ -1,0 +1,26 @@
+#ifndef CONFIG_TLS_H
+#define CONFIG_TLS_H
+
+/** @file
+ *
+ * TLS configuration
+ *
+ * These options affect the operation of the behaviour of the
+ * TLS implementation.
+ *
+ */
+
+FILE_LICENCE ( GPL2_OR_LATER );
+
+#define	TLS_FRAGMENTATION_ENABLED	/* If the TLS implementation should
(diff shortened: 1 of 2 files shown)


NiKiZe:
Hi,
Just a quick heads up, iPXE does almost never accept `#if` usage, (in .c files)
See https://dox.ipxe.org/ifdef_harmful.html

mcb30:
Thanks for the suggestion.  As far as I know, omitting the maximum fragment length option doesn't prevent the server from sending fragmented records; it just leaves the server free to choose larger fragment sizes.  I don't think this patch therefore has the effect that you expect it to have.

I see you have created a newer pull request #116 which instead adds support for handling fragmented handshake records.  That pull request still needs review, but it's a better approach overall so I'm closing this one in favour of #116.

PR #113, opened 2020-05-27
[efi] Restore the TPL to the original one

huanghe4:
iPXE added a workaround to raise TPL to TPL_CALLBACK always, but it
would cause UEFI firmware exit boot service callback with the same TPL
level cannot be executed. Worse, some UEFI modules set a pointer at exit
boot service callback and use later, then will encounter exception /
hang for invalid pointer which is not set to a correct one since the
exit boot service callback is not executed.

The solution is to restore the TPL to the original one but not always
set it to TPL_CALLBACK.

### DIFF
--- a/src/interface/efi/efi_timer.c
+++ b/src/interface/efi/efi_timer.c
@@ -129,13 +129,12 @@ static unsigned long efi_currticks ( void ) {
 	if ( efi_shutdown_in_progress ) {
 		efi_jiffies++;
 	} else {
-			Efi_OldTPL = bs->RaiseTPL( TPL_CALLBACK);
-			bs->RestoreTPL ( TPL_APPLICATION );
-			bs->RaiseTPL ( TPL_CALLBACK );
-			if ( Efi_OldTPL != 0 ) {
-				bs->RestoreTPL(Efi_OldTPL);
-			}
-
+		Efi_OldTPL = bs->RaiseTPL( TPL_CALLBACK);
+		bs->RestoreTPL ( TPL_APPLICATION );
+		bs->RaiseTPL ( TPL_CALLBACK );
+		if ( Efi_OldTPL != 0 ) {
+			bs->RestoreTPL(Efi_OldTPL);
+		}
 	}
 
 	return ( efi_jiffies * ( TICKS_PER_SEC / EFI_JIFFIES_PER_SEC ) );
--- a/src/interface/efi/efi_timer.c
+++ b/src/interface/efi/efi_timer.c
@@ -131,8 +131,8 @@ static unsigned long efi_currticks ( void ) {
 	} else {
 			Efi_OldTPL = bs->RaiseTPL( TPL_CALLBACK);
 			bs->RestoreTPL ( TPL_APPLICATION );
-            bs->RaiseTPL ( TPL_CALLBACK );
-            if ( Efi_OldTPL != 0 ) {
+			bs->RaiseTPL ( TPL_CALLBACK );
+			if ( Efi_OldTPL != 0 ) {
 				bs->RestoreTPL(Efi_OldTPL);
 			}
 
--- a/src/interface/efi/efi_timer.c
+++ b/src/interface/efi/efi_timer.c
@@ -124,11 +124,18 @@ static unsigned long efi_currticks ( void ) {
 	 * EFI's violation of this assumption by falling back to a
 	 * simple free-running monotonic counter during shutdown.
 	 */
+	 EFI_TPL Efi_OldTPL;
+	 Efi_OldTPL=0;
 	if ( efi_shutdown_in_progress ) {
 		efi_jiffies++;
 	} else {
-		bs->RestoreTPL ( TPL_APPLICATION );
-		bs->RaiseTPL ( TPL_CALLBACK );
+			Efi_OldTPL = bs->RaiseTPL( TPL_CALLBACK);
+			bs->RestoreTPL ( TPL_APPLICATION );
+            bs->RaiseTPL ( TPL_CALLBACK );
+            if ( Efi_OldTPL != 0 ) {
+				bs->RestoreTPL(Efi_OldTPL);
+			}
+
 	}
 
 	return ( efi_jiffies * ( TICKS_PER_SEC / EFI_JIFFIES_PER_SEC ) );


src/interface/efi/efi_timer.c
```diff
@@ -124,11 +124,18 @@ static unsigned long efi_currticks ( void ) {
...
 	 * simple free-running monotonic counter during shutdown.
 	 */
+	 EFI_TPL Efi_OldTPL;
+	 Efi_OldTPL=0;
 	if ( efi_shutdown_in_progress ) {
 		efi_jiffies++;
 	} else {
-		bs->RestoreTPL ( TPL_APPLICATION );
-		bs->RaiseTPL ( TPL_CALLBACK );
+			Efi_OldTPL = bs->RaiseTPL( TPL_CALLBACK);
```
NiKiZe, line 132: Indentation seems to be of

huanghe4: Updated, please pull again.

huanghe4: Now indentation is correct now.

huanghe4:
This is related to a urgent issue. System would hang.

ipxe-devel:
On Thu, May 28, 2020 at 02:55:02AM -0700, He Huang wrote:
> This is related to a urgent issue. System would hang. 

:-)

huanghe4:
What's the process to get this fix in?

jjohnson42:
I'll second this as a problem. Have been testing 'snponly.efi' and I'm not sure the circumstances, but occasionally linux will error about trying to exit boot services, and esxi pretty much always hangs trying to exit boot services (also some 'keyboard errors' are printed early on) when booted from snponly.efi.

This change does seem to sidestep it. I don't know enough to gauge the cleaner approach (to my inexperienced eyes, it would be equivalent to just deleting the TPL lines from the timer function) nor do I have familiarity with the problem that required those lines in the first place.

mcb30:
Neither the code change nor the commit message make any sense to me, sorry.

Instead of the commit, could you please provide a way to reproduce your observed problem?

Thanks.

huanghe4:
You need an AMI based UEFI BIOS, and then power on and boot ESXi 6.5U3 through UEFI iPXE and without norts =1 as the parameter of boot loader.

huanghe4:
And it also does not make sense to me that iPXE always restore to TPL_CALLBACK, even the original TPL is TPL_APPLICATION.
If you do not have env to reproduct this issue, would you print log to monitor the original/old TPL? Then you will see, it is not always the TPL_CALLBACK.
Please read UEFI spec about the "TPL Restriction".

JamesWangAmiTw:
@mcb30,

It raised TPL to TPL_CALLBACK and never restored it.  NotifyFunction of EVT_SIGNAL_EXIT_BOOT_SERVICES will not be called because of that while booting.

The change of efi_currticks() causing this problem was from [[efi] Run at TPL_CALLBACK to protect against UEFI timers](https://github.com/ipxe/ipxe/commit/c89a446cf09f30a121ae21d91f4a1aa071044084#diff-8abaf825db412e28b0c57b0f8947ed23).

I am wondering if you can find another way to work around the [problem](https://github.com/ipxe/ipxe/commit/c89a446cf09f30a121ae21d91f4a1aa071044084#diff-8abaf825db412e28b0c57b0f8947ed23) or restore TPL before ExitBootService?

mcb30:
@huanghe4 The expected mechanism (as documented in [[efi] Run at TPL_CALLBACK to protect against UEFI timers](https://github.com/ipxe/ipxe/commit/c89a446cf09f30a121ae21d91f4a1aa071044084#diff-8abaf825db412e28b0c57b0f8947ed23) is that iPXE code runs at TPL_CALLBACK almost all of the time, in order to overcome fundamental design flaws in the UEFI specification.

iPXE will record the TPL at which it was entered and will restore this TPL before exiting.  This applies to all externally callable entry points, such as the SNP API (see e.g. the matched RaiseTPL/RestoreTPL calls in `efi_snp_get_status()`).

This also applies to the overall code path from initial invocation through to ExitBootServices, although this is less immediately obvious due to the exchange of responsibilities between different UEFI components.  The expected sequence of events is:
* An application-level entry point such as `_efi_start()` or `efi_snp_load_file()` (as opposed to an API-level entry point such as `efi_snp_get_status()`) will call `efi_snp_claim()` to claim exclusive ownership of the network devices for iPXE (and so to prevent packet loss due to other UEFI code attempting to poll the network).
* This call to `efi_snp_claim()` will record the original TPL in `efi_snp_old_tpl` and will raise to TPL_CALLBACK.
* Before executing an image (such as an OS kernel), iPXE will call `efi_snp_release()`
* This call to `efi_snp_release()` will restore the original TPL as recorded in `efi_snp_old_tpl`
* iPXE then invokes the OS kernel via `StartImage()`.  At this point, the TPL is whatever TPL was recorded when iPXE was started, which should be TPL_APPLICATION.
* The OS kernel is responsible for calling `ExitBootServices()`

I have added debug code to iPXE to dump the TPL prior to calling `StartImage()` and (via the `efi_wrap` mechanism) to dump the TPL at the point that the OS kernel invokes `ExitBootServices()`.  I have tested with both an iPXE EFI ROM and with a chainloaded `snponly.efi`.  In both cases and at both debug points the TPL was TPL_APPLICATION, as expected.

Is it possible that the AMI BIOS is erroneously starting iPXE at TPL_CALLBACK, thereby causing `efi_snp_old_tpl` to end up containing TPL_CALLBACK rather than TPL_APPLICATION?

Michael

jjohnson42:
I can confirm same behavior, using bootx64.efi from esxi 7, the following produces the same behavior he reported:
#!ipxe
chain boot/efi/boot/bootx64.efi -c /path/to/boot.cfg also call efi_snp_release() of course.

> iPXE then invokes the OS kernel via StartImage(). At this point, the TPL is whatever TPL was recorded when iPXE was started, which should be TPL_APPLICATION.

Well, iPXE code may still run at this point. In fact it explicitly does for iPXE booted ESXi boot loader. efi_download.c has three functions that may be called by the loaded efi executable. At the very least they call efi_download_start. I don't know the full flow of the code, but I am wondering if efi_download_start use of intf_init or xfer_open before calling efi_snp_claim causes a RaiseTPL before efi_snp_claim stores the old TPL causing it to erroneously 'save' iPXE induced TPL_CALLBACK as the 'old' TPL? I am a bit preoccupied today, but I'd be tempted to modify efi_download_start() to call claim earlier and call efi_snp_release() on error. I'm a bit rusty and I botched even a simple change so I'm not sure.

If my naive guess happens to be in the ballpark, it doesn't explain a *seeming* improvement in our Linux boot testing, where iPXE DL protocol is not used and we also were seeing errors on kernel trying to exit boot servicing that happened *sometimes*. However this was a little fuzzier so it may not be completely accurate.

jjohnson42:
Alternatively, if efi_snp_old_tpl seems to have been previously assigned, perhaps don't save a new 'old' value?

huanghe4:
@mcb30 ,

The root cause is in static unsigned long efi_currticks ( void );
Then please print the original TPL in this function.
Because this TPL is always raised to TPL_CALLBACK, then when gEfiEventExitBootServicesGuid is triggerred, some notify function which are registerred at TPL_CALLBACK cannot be executed. 
You can register such kind of callback at TPL_CALLBACK, and check whether it can be executed when gEfiEventExitBootServicesGuid is signnalled.

@huanghe4

mcb30:
@jjohnson42 Thanks for the hints about reproduction via ESXi 7's bootx64.efi.  I will try to assemble a test case in qemu that demonstrates the problem.  If you have a minimal test case ready to use, that would be really helpful.

@huanghe4 What you are describing is not a **root** cause.  If efi_currticks() is being called at a level other than TPL_CALLBACK then that indicates a missing RaiseTPL somewhere else in the code.  Masking this bug by modifying efi_currticks() to restore the "original" TPL would leave us with a latent race condition, which will be substantially harder to debug in future.  We need to find the actual root cause, which is the missing RaiseTPL that is causing efi_currticks() to be invoked at the wrong TPL.

JamesWangAmiTw:
@mcb30 

Sorry for not being specific in my [previous comment](https://github.com/ipxe/ipxe/pull/113#issuecomment-651511355).  I didn't mean TPL is never restored anywhere but it was changed in timer event.

When OS image starts, is the `efi_currticks()` still active?  
I assume it only stops when `efi_tick_shutdown()` is called.

So when `efi_snp_release()` the TPL is restored but what happen when `efi_currticks()` is triggered after `efi_snp_release()`?
If so the TPL can still be TPL_CALLBACK before `ExitBootService()`

And `efi_tick_shutdown()` itself is a NotifyFunction of `EVT_SIGNAL_EXIT_BOOT_SERVICES`.
If `EVT_SIGNAL_EXIT_BOOT_SERVICES` is signaled and TPL = `TPL_CALLBACK`, `efi_tick_shutdown()` will not be called because of the same TPL.

James

mcb30:
@JamesWangAmiTw The timer callback function is `efi_tick()`, not `efi_currticks()`.  The timer event controlled by `efi_tick_startup()`/`efi_tick_shutdown()` does not execute the code in `efi_currticks()`.

There should not be any code path that leads to `efi_currticks()` while at TPL_APPLICATION.  If such a code path exists, then that code path is missing a call to RaiseTPL.  This is the bug that we need to find and fix.

The most promising approach seems to be for me to reproduce the issue locally.  The tests that I have done so far do not reproduce the issue: I am observing only the expected TPL at all times.  I am working on attempting to reproduce the problem as per @jjohnson42's suggestion.  If you have an alternative test case that allows me to reproduce the issue locally then I am happy to use that instead.

Michael

JamesWangAmiTw:
@mcb30 
Yes you are right, I have the wrong impression without double checking the code, the timer calls   'efi_tick()' indeed.

JamesWangAmiTw:
@mcb30 

We followed [PXE Boot the ESXi Installer Using a Web Server](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.esxi.upgrade.doc/GUID-147D7509-EFB1-4391-973F-48B015B85C83.html)

In AMI BIOS Setup: Advanced - Network Stack Configuration: Enable all "Network Stack"and "Ipv4 PXE Support".
And choose to boot from UEFI PXE boot option to launch ESXi Installer.

I wonder if you have a system running with AMI BIOS to reproduce?

mcb30:
@JamesWangAmiTw Thanks for the link.  I have reproduced some kind of lockup using qemu with an iPXE ROM (8086100e.efirom).  The last messages seen on screen are:
```
Shutting down firmware services...
Using 'simple offset' UEFI RTS mapping policy
```
Booting qemu from the ISO does not hang at the same place, so this looks promising as a reproduction.

mcb30:
@jjohnson42 Your suspicion seems to be correct: the TPL is being raised by the call to `xfer_open()` and so `efi_snp_claim()` is saving the wrong TPL.

mcb30:
I have created #120 which seems to fix the root cause.  With this fix, I am able to boot ESXi as expected.

@JamesWangAmiTw I don't have a name and email address for you: could you please let me know which name and e-mail to use for the commit log message?

JamesWangAmiTw:
Hi Michael,

Thanks for the fix, we will try that.  I have updated my profile with the public name and email address.

Name: James Wang
Email: jameswang@ami.com.tw

JamesWangAmiTw:
@mcb30 
We have tested [this](https://github.com/ipxe/ipxe/pull/120) on 2 systems and it fixes the problem.  Thanks!

mcb30:
Merged in #120 

Thank you for all the debugging!

stappersg:
afbeelding

Partial text from [that #349 comment](https://github.com/ipxe/ipxe/issues/349#issuecomment-1126793404)
> The problem we are seeing however is still exactly the one that was discussed in https://github.com/ipxe/ipxe/pull/113 and https://github.com/ipxe/ipxe/pull/120 did not fix the problem entirely.

PR #114, opened 2020-05-28
Add uribase setting for getting the host:port portion of a URI.

planier-g:
This is useful for fetching relative URLs based on the DHCP server
provided file URI (as in "${filename|uribase}/some/other/path").

DHCPv6 does not provide a next-server equivalent so using uribase provides something like that for http URIs provided as the filename.

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -697,6 +697,10 @@ int fetch_setting ( struct settings *settings, const struct setting *setting,
 		if ( ( ret = settings->op->fetch ( settings, &tmp,
 						   data, len ) ) >= 0 ) {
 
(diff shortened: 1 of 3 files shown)


ipxe-devel:
On Thu, May 28, 2020 at 08:56:38AM -0700, planier-g wrote:
> This is useful for fetching relative URLs based on the DHCP server
> provided file URI (as in "${filename|uribase}/some/other/path";).
> 
> DHCPv6 does not provide a next-server equivalent so using uribase
> provides something like that for http URIs provided as the filename.
> 

Some how I do read that as "Reminder on documenting scriptlet better"

 
Regards
Geert Stappers
-- 
Silence is hard to parse

PR #115, opened 2020-06-17
fix EMBED: don't eagerly assign over non-empty value

realtime-neil:


mcb30:
What problem is this actually solving?  Any variables specified in the `make` command line will override assignments within the Makefile.  Building with `make EMBED=...` works with no problems already.

realtime-neil:
> What problem is this actually solving?

Allow me to demonstrate:

```
$ echo && cat Makefile.bad && echo && EMBED=foobar make -f Makefile.bad && echo && cat Makefile.good && EMBED=foobar make -f Makefile.good 

EMBED := $(EMBEDDED_IMAGE)

.PHONY: all
all:
	echo "EMBEDDED_IMAGE: $(EMBEDDED_IMAGE)"
	echo "EMBED: $(EMBED)"


echo "EMBEDDED_IMAGE: "
EMBEDDED_IMAGE: 
echo "EMBED: "
EMBED: 

EMBED ?= $(EMBEDDED_IMAGE)

.PHONY: all
all:
	echo "EMBEDDED_IMAGE: $(EMBEDDED_IMAGE)"
	echo "EMBED: $(EMBED)"

echo "EMBEDDED_IMAGE: "
EMBEDDED_IMAGE: 
echo "EMBED: foobar"
EMBED: foobar
```

When the definition of the shell variable `EMBED` precedes `make`, `Makefile.good` shows the effect and `Makefile.bad` does not.

> Any variables specified in the `make` command line will override assignments within the Makefile. Building with `make EMBED=...` works with no problems already.

You're not wrong: `make EMBED=...` works. What doesn't work is `EMBED=... make` or any combination of exporting a shell variable for later consumption by a `make` invocation. For reasons that I have yet to dig out of the GNU Make manual, consuming that variable definition only works when it appears as a argument to the `make` command.

Strictly speaking, it's not incorrect to accept only variable assignments that happen to follow the command, but it will confuse others (like me) that assume the `EMBED` should behave like a shell variable when the immediate assignment prevents exactly that.

realtime-neil:
Found it:

> Any target in the makefile may be specified as a goal (unless it starts with ‘-’ or contains an ‘=’, in which case it will be parsed as a switch or variable definition, respectively).
-- https://www.gnu.org/software/make/manual/make.html#Goals

I make my living on exactly this kind of minutia, but IMHO, one parenthetical clause in the `make` manual is too little reason to flout the principle of least surprise.

realtime-neil:
@mcb30 I can't tell if you're biding your time in thoughtful contemplation or giving me the "Slow No". I want to believe it's the former, but --- since this patch is a literal single character --- I doubt that optimism is justified. If you're _not_ going to merge it, then could you please close this PR, possibly with a comment explaining why?

mcb30:
@realtime-neil I'm planning to reject this change on the basis that all of the iPXE documentation suggests using the form `make EMBED=...` and that there are large numbers of other iPXE Makefile variables that function in the same way (ignoring environment variables).  The GNU make documentation explicitly recommends against relying upon environment variables to provide values.  The fact that it happens to work for `EMBEDDED_IMAGE` is an accident of history.

I have created https://github.com/danderson/netboot/pull/118 to fix `netboot` to use the documented `make EMBED=...` instead.

Hope that works for you!

realtime-neil:
@mcb30 thank you for your explanation. I found the sentiment you referenced:

> When make runs a recipe, variables defined in the makefile are placed into the environment of each shell.
> ...
> Other use of variables from the environment is not recommended. It is not wise for makefiles to depend for their functioning on environment variables set up outside their control, since this would cause different users to get different results from the same makefile. This is against the whole purpose of most makefiles.

-- https://www.gnu.org/software/make/manual/make.html#Environment

PR #116, opened 2020-06-23
[tls] Add support for fragmented tls handshake packets

chschenk:
iPXE currently does not support TLS connections with large certificate chains because it can not handle TLS handshake record fragmentation. So i implemented the detection and reassembling of fragmented packets. Please let me know if there is any problem with the code or the way i handled the reassembling.

### DIFF
--- a/src/include/ipxe/tls.h
+++ b/src/include/ipxe/tls.h
@@ -355,6 +355,11 @@ struct tls_connection {
 	struct io_buffer rx_header_iobuf;
 	/** List of received data buffers */
 	struct list_head rx_data;
+
+	/** Bytes remaining for current handshake record */
+	size_t handshake_fragment_remaining;
+	/** List of fragments of current handshake record */
+	struct list_head handshake_fragments;
 };
 
 /** RX I/O buffer size
--- a/src/net/tls.c
+++ b/src/net/tls.c
@@ -1989,18 +1989,51 @@ static int tls_new_finished ( struct tls_connection *tls,
  */
 static int tls_new_handshake ( struct tls_connection *tls,
 			       const void *data, size_t len ) {
-	size_t remaining = len;
+	size_t remaining;
 	int rc;
+	struct io_buffer *iobuf;
+	uint8_t fragment_finished = 0;
+
+	iobuf = alloc_iob_raw ( len, TLS_RX_ALIGN, 0 );
+	memcpy ( iob_put ( iobuf, len ), data, len );
+
+	if ( tls->handshake_fragment_remaining ) {
+		list_add_tail ( &iobuf->list, &tls->handshake_fragments );
+		if ( tls->handshake_fragment_remaining > len)
+		{
+			tls->handshake_fragment_remaining -= len;
+			return 0;
+		}
+		else {
+			tls->handshake_fragment_remaining = 0;
+			fragment_finished = 1;
+			iobuf = iob_concatenate ( &tls->handshake_fragments );
+			len = iob_len ( iobuf );
+		}
+	}
+
+	remaining = len;
 
 	while ( remaining ) {
 		const struct {
 			uint8_t type;
 			tls24_t length;
 			uint8_t payload[0];
-		} __attribute__ (( packed )) *handshake = data;
+		} __attribute__ (( packed )) *handshake = iobuf->data;
 		const void *payload;
 		size_t payload_len;
 		size_t record_len;
+		size_t frag_len;
+
+		/* Check for fragmentation */
+		payload_len = tls_uint24 ( &handshake->length );
+		frag_len = ( len - sizeof ( *handshake ) );
+
+		if ( payload_len > len && fragment_finished == 0 ) {
+				list_add_tail ( &iobuf->list, &tls->handshake_fragments );
+				tls->handshake_fragment_remaining = payload_len - frag_len;
+				return 0;
+		}
 
 		/* Parse header */
 		if ( sizeof ( *handshake ) > remaining ) {
@@ -2009,13 +2042,7 @@ static int tls_new_handshake ( struct tls_connection *tls,
 			DBGC_HD ( tls, data, remaining );
 			return -EINVAL_HANDSHAKE;
 		}
-		payload_len = tls_uint24 ( &handshake->length );
-		if ( payload_len > ( remaining - sizeof ( *handshake ) ) ) {
-			DBGC ( tls, "TLS %p received overlength Handshake\n",
-			       tls );
-			DBGC_HD ( tls, data, len );
-			return -EINVAL_HANDSHAKE;
-		}
+
 		payload = &handshake->payload;
 		record_len = ( sizeof ( *handshake ) + payload_len );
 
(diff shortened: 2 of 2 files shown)


chschenk:
Hi,
what do you think of this patch? I really would love to see this upstream.

Best regards
Christopher

ipxe-devel:
On Tue, Jul 14, 2020 at 04:02:33AM -0700, Christopher wrote:
> 
> I really would love to see this upstream.
> 

So go furher,  make it happen

odenbach:
We are running this patch in production now, works like a charm!

ctheune:
I've tried this in production and it helped me getting booting from TLS working again. I did not intentionally have a large certificate chain, but I'm using LE and apart from allowing non EC-ciphers this has been a blocker for us to properly boot from TLS.

Any chance to get this merged in the foreseeable future? Any help needed?

grahamc:
Just in case you're coming across this with the recent LetsnEncrypt chain update, you can work around this by preferring the shorter chain in https://community.letsencrypt.org/t/production-chain-changes/150739/1 via the lego flags: `--preferred-chain "ISRG Root X1"`. On NixOS this is provided via:

```nix
   security.acme.certs."${domain}".extraLegoRunFlags = [
        "--preferred-chain" "ISRG Root X1"
   ];
```

rodehoed:
We also have problems with the chain from Lets Encrypt because of it's size. So a fix upstrea would be awesome. 

For now i'm not using the fullchain.pem but cert.pem certificate in nginx as a work-around.

dvzrv:
On Arch Linux we have been struggling with the issue as well (at the time of writing netboot for users is broken) and applying this patch fixes the problem.

It would be really great if this was reviewed and merged! :)

martonivan:
I've also applied this patch in our staging environment and it worked flawlessly. We are also eagerly waiting for having it in master.

eb3095:
No good for me. This fixed the error, however now VM's I'm spinning up are hanging whenever I attempt to chain an https address. No output from debug, seems to just hang after "new session ticket" and it says "ok".

waldner:
What's the status of this? Is a merge expected soon(ish)? After almost two years the problem has only got worse.

bantu:
We applied this patch to our code base and it seemed to work fine. However, in the meantime I arrived at the same conclusion as described in https://github.com/ipxe/ipxe/pull/116#issuecomment-862709507, reverted the patch and instead switched to the shorter Let's Encrypt certification chain: `R3 -> ISRG Root X1` instead of `R3 -> ISRG Root X1 -> DST Root CA X3`. Our certificate chains is only 2887 bytes long although we ise 4096-bit RSA keys. Maybe it is worth pointing out that linking to `DST Root CA X3` seems kind of pointless as this root certificate expired `Thu, 30 Sep 2021 14:01:15 UTC`.

odenbach:
> We applied this patch to our code base and it seemed to work fine. However, in the meantime I arrived at the same conclusion as described in [#116 (comment)](https://github.com/ipxe/ipxe/pull/116#issuecomment-862709507), reverted the patch and instead switched to the shorter Let's Encrypt certification chain: `R3 -> ISRG Root X1` instead of `R3 -> ISRG Root X1 -> DST Root CA X3`. Our certificate chain is now only 2887 bytes long although we use 4096-bit RSA keys. It is probably worth pointing out that linking to `DST Root CA X3` is kind of pointless as this root certificate expired `Thu, 30 Sep 2021 14:01:15 UTC`.

This may be a solution for you or other users of Let's Encrypt, but many other people have other long certificate chains, so we need a general solution. I cannot understand why this patch does not get merged.

omniproc:
Can confirm this does fix it for us as well. Would be great if this could be merged in the next release.

stappersg:
On Fri, Apr 08, 2022 at 12:33:44AM -0700, Omniproc wrote:
> Can confirm this does fix it for us as well. Would be great if this could be merged in the next release.

So go furher,  make it happen

stappersg:
Today, timezone is CEST, at the iPXE  irc channel:
```text
20:23 -!- dvzrv [~dvzrv@archlinux/developer/dvzrv] has joined #ipxe
20:24 < dvzrv> hi! any possibility of this being merged? https://github.com/ipxe/ipxe/pull/116 :)
```
now also at the iPXE mailinglist   :-)

omniproc:
> Time needed to verify correctness of this patch is one possible issue, maybe this could be lessened with proper tests, or at least something that can be used to verify the issue. But this is critical code that and we must be absolute certain does not introduce any new security issues.

Note that there's also https://github.com/chschenk/ipxe/tree/feature/tls_fragmentation_buildoption which add's a build option. By making this code optional and disable that build option by default it at least wouldn't break anything. Regarding security concerns: granted. But we're talking about ~50 lines of code here.

omniproc:
> > Note that there's also https://github.com/chschenk/ipxe/tree/feature/tls_fragmentation_buildoption which add's a build option. By making this code optional and disable that build option by default it at least wouldn't break anything. Regarding security concerns: granted. But we're talking about ~50 lines of code here.
> 
> To that, please see https://dox.ipxe.org/ifdef_harmful.html You can introduce huge security issues with only one line of code.

I wasn't suggesting that you can't. I was trying to say that it's "only" 50 lines of code to review.

dvzrv:
This PR has been open for nearly two years, while fixing a real world problem, which prevents ipxe to function at all when using TLS certificates provided e.g. through letsencrypt.

It is opaque to outsiders why there is no review/ merge, as even after trying to reach out via mail or IRC there has been no response besides speculation about what might be blocking the review of this.

@mcb30 do you have some time to review and merge this? It would be much appreciated! :heart:  It does not seem (from my viewpoint) anyone else is able to do this.

NiKiZe:
@dvzrv which I have responded to.
At least add a test case showing the issue.

I don't think anyone doubts that this PR resolves a real issue. But does it do so without introducing other issues?

stappersg:
Quoting point _six_ of https://datatracker.ietf.org/doc/html/rfc1925
> It is easier to move a problem around (for example, by moving the problem to a different part of the overall network architecture) than it is to solve it.

and _six a_:
> It is always possible to add another level of indirection.


That is to encourage somebody to write
> At URL is documented how to create a `initrd` with a _http to https web proxy_.

From that point on you can _iPXE boot_ one server with that `initrd`.  Boot / install further servers through that _http to https web proxy_. This approach keeps _iPXE_ a _netbootloader_, it prevents _iPXE_ getting fat with _encryption bloat_.

It is **trust** what we need, not _encryption_.

ggiesen:
Seeing this issue with Sectigo certs as well. Would love to see it merged.

nullfact0r:
September 29 and seeing this issue with a recently issued LetsEncrypt cert as well

odenbach:
>  At least add a test case showing the issue.

Here you go:

1. Start iPXE with https support
2. Press CTRL-B to enter a shell
3. enter "kernel https://test-v4-extern.uni-paderborn.de/test.ipxe"
4. enter "boot"

Expected result: get a (test) boot menu
Result without patch: error
Result with patch: as expected

Does that help in any way?

Thanks,

Christopher

mcb30:
Closing in favour of merged PR #930 (with credit to @chschenk for the original implementation).

PR #121, opened 2020-07-21
Fix typo in comment describing length codes

ComputerDruid:
It now matches the value actually used in code at https://github.com/ipxe/ipxe/blob/366206517e9974dea62d0df313bdd59595c89f52/src/crypto/deflate.c#L868


mcb30:
Merged as http://github.com/ipxe/ipxe/commit/8bc85ec6f; thanks

PR #124, opened 2020-07-21
[efi] Check the status code from AllocatePool()

ignatk:
According to the latest UEFI specification (Version 2.8 Errata B) p. 7.2:

"Buffer: A pointer to a pointer to the allocated buffer if the call succeeds;
undefined otherwise."

So implementations are obliged neither to return NULL, if the allocation fails,
nor to preserve the contents of the pointer.

Make the logic more reliable by checking the status code from AllocatePool()
instead of checking the returned pointer for NULL


mcb30:
Merged as http://github.com/ipxe/ipxe/commit/b76052335; thank you.

PR #136, opened 2020-08-21
[efi] Add support for TPM measurement of ipxe scripts on EFI systems

ajyoung-oracle:
This PR adds a new --measure (-m) option to the chain/imgexec/boot commands that will perform a TPM measurement of an ipxe script prior to loading of the script on EFI systems. The code makes use of the EFI_TCG2_PROTOCOL->HashLogExtendEvent() protocol function provided by the UEFI firmware which will measure the script and create a TPM log entry for the script. NOTE: Only TPM2.0 is supported currently. TPM1.2 support is TBD.

See the TCG PC Client Platform Firmware Profile Specification and TCG EFI Protocol Specification for details.

iPXE uses UEFI LoadImage() service function to load other image types which will automatically measure the image as part of the image load. Script image types are not loaded via LoadImage() and thus this new command option is needed to measure the scripts directly prior to load. Scripts need to be measured to comply with TCG specification which dictate that all binaries and configuration data must be measured for a complete measured boot.

Example usage:

chain -m http://1.2.3.4/script.ipxe

Testing: I tested this code on a x86_64 server using qemu with a swtpm emulating TPM2.0. I verified that t
he proper TPM PCR (5) was updated and an TPM Event Log entry was created for the script.

Thanks,

-Aaron Young
aaron.young@oracle.com

### DIFF
--- a/src/core/acpimac.c
+++ b/src/core/acpimac.c
@@ -29,6 +29,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #include <ipxe/base16.h>
 #include <ipxe/ethernet.h>
 #include <ipxe/if_ether.h>
+#include <ipxe/settings.h>
 #include <ipxe/acpimac.h>
 
 /** @file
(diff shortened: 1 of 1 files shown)


mcb30:
Thanks.  I haven't looked at the code yet, but one initial question: if the TCG specification requires that all binaries and configuration data is measured, then does it make sense for this to be controlled via a command option?  Should it instead happen automatically for all executed scripts on any system that supports EFI_TCG2_PROTOCOL?

ajyoung-oracle:
Re: does it make sense for this to be controlled via a command option? Should it instead happen automatically...?

Maybe. Sure... can do that if you think that is best approach. I'm always apprehensive about changing behavior (i.e. TPM measurements in this case) without making it configurable and thus opted for a command option. But, making it automatic probably does make sense.

mcb30:
Thanks.  Follow-up question: what does "all binaries and configuration data" actually cover?  Would it be sufficient to measure only scripts that we start executing, or would it be necessary to measure every image (even non-executable images such as PNG files or initrds) at the point of downloading?

ajyoung-oracle:
Re:  Would it be sufficient to measure only scripts that we start executing, or would it be necessary to measure every image (even non-executable images such as PNG files or initrds) at the point of downloading?

From the TCG specs about SRTM, it states "The goal of the SRTM is to provide an unbroken chain of measurements for components that executed on the platform  or  security relevant  configuration  data  that  may  have  influenced  platform state."

So, this is a bit subjective. But, to me, the answer is anything that could potentially affect platform state should be measured. Scripts are clearly covered. Anything that is part of boot is covered.

Note - as I stated in the initial description - any image that is loaded via LoadImage() BS, is automatically measured by the UEFI firmware (assuming TPM/TCG code is enabled in the UEFI firmware).

Does that help? ;-)

ajyoung-oracle:
We (Oracle) just needed support for measuring scripts - so that was my focus. Other image types could easily be added to the code.

ajyoung-oracle:
Re: Would it be sufficient to measure only scripts that we start executing, or would it be necessary to measure every image (even non-executable images such as PNG files or initrds) at the point of downloading?

 I've further discussed this with colleagues (who are much more knowledgeable than I) and they say it depends on what these other images are used for. There is precedent, for example, for using custom splash screen images to root legacy BIOS. It is best to error on the side of caution and measure anything that could possibly affect system state. Doesn't hurt to measure extra things as long as it is done consistently on every boot.

Does that help?

What are next steps here? Should I resubmit a new PR without command option (so that measurement is done automatically) for starters?

konradwilk:
ping? @mcb30 - should the measurements be done automatically?

PR #137, opened 2020-08-23
[build] Fix building on older versions of gcc

mcb30:
Older versions of gcc (observed with gcc 4.5.3) require attributes to
be specified on the first declaration of a symbol, and will silently
ignore attributes specified after the initial declaration.  This
causes the ASN.1 OID-identified algorithms to end up misaligned.

Fix by adding __asn1_algorithm to the initial declarations in asn1.h.

Debugged-by: Dentcho Bankov <dbankov@vmware.com>


dbankov-vmware:
Just to confirm this change fixes the problem. Thank you very much!

PR #139, opened 2020-09-13
Fixed hang on rtl8139 initialization

UCIS:
On my old Pentium 4 PC the initialization of the rtl8139 network interface (either via ifopen or dhcp) would hang in most situations, requiring a hard reset. The added delay appears to fix this reliably.

I am not entirely sure whether a delay is required here, or whether it's a side effect of the delay call that actually eliminates the problem. The problem also went away when I enabled debugging for the realtek module, likely because of the delay introduced by the subsequent debug log statement. The problem did not occur with the default configuration when using the dhcp command in the beginning of the script with error suppression, but did occur each time without error suppression, when used later in the script, or when both HTTPS and image verification were enabled.

### DIFF
--- a/src/core/image.c
+++ b/src/core/image.c
@@ -481,3 +481,13 @@ int image_set_trust ( int require_trusted, int permanent ) {
 
 	return 0;
 }
+
+/**
+ * Get image trust requirement
+ *
+ * @ret require_trusted		Whether images are required to be trusted
+ */
+int image_get_require_trust ( ) {
+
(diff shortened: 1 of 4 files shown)


MaxPeal:
Shud the 
"Allow to add trusted certificates at run time"
Not in a separate PR?

UCIS:
@MaxPeal Yes, it should. I did not realize GitHub would automatically include subsequent commits (I did make that commit after I made the merge commit).

PR #140, opened 2020-09-13
Fixed hang on rtl8139 initialization

UCIS:
On my old Pentium 4 PC the initialization of the rtl8139 network interface (either via ifopen or dhcp) would hang in most situations, requiring a hard reset. The added delay appears to fix this reliably.

I am not entirely sure whether a delay is required here, or whether it's a side effect of the delay call that actually eliminates the problem. The problem also went away when I enabled debugging for the realtek module, likely because of the delay introduced by the subsequent debug log statement. The problem did not occur with the default configuration when using the dhcp command in the beginning of the script with error suppression, but did occur each time without error suppression, when used later in the script, or when both HTTPS and image verification were enabled.

### DIFF
--- a/src/drivers/net/realtek.c
+++ b/src/drivers/net/realtek.c
@@ -531,6 +531,7 @@ static int realtek_create_buffer ( struct realtek_nic *rtl ) {
 
 	/* Program buffer address */
 	writel ( address, rtl->regs + RTL_RBSTART );
+	udelay ( 1 );
 	DBGC ( rtl, "REALTEK %p receive buffer is at [%08llx,%08llx,%08llx)\n",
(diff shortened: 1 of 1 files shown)


mcb30:
As per my comment on IRC: please run some diagnostics such as memtest86+ first to ensure that the system itself is not faulty.  From the description so far, it sounds as though the fix may be working by pure coincidence, which could be related to e.g. changing the memory layout of the binary in a way that happens to avoid a faulty bit in the DIMM.

mcb30:
From conversations in IRC, it appears that the hardware in question is now physically broken (including a broken DIMM slot and stray wires in PCI slots).  Closing this as unresolvable.

PR #141, opened 2020-09-13
Allow to add trusted certificates at run time

UCIS:
This patch allows to mark certificates imported by the certstore command as trusted. This allows the user to load trusted certificates  at runtime in an existing build using the `certstore --trusted ...` command, rather than having to compile certificates into a new build or managing cross-signed certificates.

I don't think this (technically) reduces security. If an attacker is in a position where iPXE commands can be entered, the attacker can also chainload an insecure binary, unless the imgtrust option is enabled, in which case loaded certificates can no longer be marked as trusted via this patch.

### DIFF
--- a/src/core/image.c
+++ b/src/core/image.c
@@ -481,3 +481,13 @@ int image_set_trust ( int require_trusted, int permanent ) {
 
 	return 0;
 }
+
+/**
+ * Get image trust requirement
+ *
+ * @ret require_trusted		Whether images are required to be trusted
+ */
+int image_get_require_trust ( ) {
+
+	return require_trusted_images;
+}
--- a/src/hci/commands/cert_cmd.c
+++ b/src/hci/commands/cert_cmd.c
@@ -46,12 +46,14 @@ struct cert_options {
 	char *name;
 	/** Keep certificate file after parsing */
 	int keep;
+	/** Trust this certificate (sets the VALIDATED flag) */
+	int trust;
 };
 
 /** "cert<xxx>" option list */
 static union {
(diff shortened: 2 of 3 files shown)


mcb30:
I am wary of this, since it (by intention) opens up the attack surface.  At a minimum, I would expect a feature such as this to have to respect existing restrictions such as those implemented in http://github.com/ipxe/ipxe/commit/bd16deaa8 - i.e. if a non-default root of trust is explicitly specified at build time then absolutely nothing should be able to change that root of trust at runtime.

I'd need to be convinced of a compelling use case that can't be solved using cross-signed certificates before accepting any variant of this patch.  What use case do you have in mind?

UCIS:
I was not aware of that restriction (or the possibility to inject variables like that). Would you prefer the `--trust` option to be excluded entirely if ALLOW_TRUST_OVERRIDE=0, or rather return a permission denied error at run time?

I think you can solve nearly any use case with cross-signed certificates, it just requires significantly more work which may lead to someone dropping a project entirely or going with a less secure approach. I don't think the rom-o-matic.dev site can currently embed custom certificates.

Personally I use this function to load the privately signed certificate for boot.salstar.sk (from the local tftp server) and then load additional certificates and the remote boot script via https. Besides convenience another possible advantage is that the salstar.sk certificates will only be trusted if I choose a specific option from the boot menu and won't be valid otherwise.

Normally the boot.salstar.sk script would just load its own iPXE binary, effectively also discarding existing restrictions. This did not work for me as the NIC initialization in that build failed (PR #140), so I made it run the external script in my own iPXE build.

Issue #147, opened 2020-09-28
Can't assemble TLS cert chain where server includes root


mcsaucy:
Howdy,

I've got an iPXE script that attempts to hit a Heroku webapp, which redirects me to a URL that has the actual content I need. Unfortunately, this it seems like we're unable to trust Heroku's star cert. https://ipxe.org/crypto indicates that anything that "In the default configuration, iPXE will therefore automatically trust the same set of certificates as the Firefox web browser." Firefox likes these, but iPXE doesn't.

Any idea what could be at play or where I could start digging?

My build is pretty simple: `#define DOWNLOAD_PROTO_HTTPS` in `src/config/general.h`, `make bin/ipxe.usb`.

### Attempting to hit anything.herokuapp.com:
```
iPXE> kernel https://foo.herokuapp.com
https://foo.herokuapp.com...
X509 chain 0xeec04 added X509 0xf4234 "*.herokuapp.com"
X509 chain 0xeec04 added X509 0xf4ed4 "DigiCert SHA2 High Assurance Server CA"
X509 chain 0xeec04 added X509 0xf5444 "DigiCert High Assurance EV Root CA"
X509 chain 0xeec04 found no usable certificates
Permission denied (http://ipxe.org/0216eb3c)
```

### Issue seems to be trusting DigiCert:
```
iPXE> kernel https://digicert.com
https://digicert.com...
X509 chain 0xeec04 added X509 0xfdaf4 "digicert.com"
X509 chain 0xeec04 added X509 0xf7614 "DigiCert SHA2 Extended Validation Server CA"
X509 chain 0xeec04 added X509 0xf5444 "DigiCert High Assurance EV Root CA"
X509 chain 0xeec04 found no usable certificates
Permission denied (http://ipxe.org/0216eb3c)
```
### At least one other CA works:
```
iPXE> kernel https://ipxe.org
https://ipxe.org...X509 chain 0xeeab4 added X509 0xf7ff4 "ipxe.org"
X509 chain 0xeeab4 added X509 0xfd024 "Let's Encrypt Authority X3"
<...>
X509 0xf7ff4 "ipxe.org" requires an OCSP check
X509 0xf7ff4 "ipxe.org" successfully validated using issuer 0xfd024 "Let's Encrypt Authority X3"
ok
```
### Firefox seems to like Heroku
So either my Firefox setup is wonky or ipxe.org/crypto is incorrect.
[...]

PR #148, opened 2020-09-29
Clarify debug msg for an untrusted x509 issuer

mcsaucy:
We surface this debugging information in cases where a cert actually lacks an issuer, but also in cases where it *has* an issuer, but we cannot trust it (e.g., due to issues in establishing a trust chain).

I stumbled upon this while debugging #147 and the existing messaging threw me off for a bit. A casual glance at the surrounding code doesn't help either, as `issuer` is derived from the chain (rather than the cert itself) by `x509_validate_chain`. Ultimately, this seems like a reasonable approach (if we can't trust an issuer, we should ignore it), but this PR seeks to surface that a bit more clearly.

### DIFF
--- a/src/crypto/x509.c
+++ b/src/crypto/x509.c
@@ -1337,7 +1337,7 @@ int x509_validate ( struct x509_certificate *cert,
 
 	/* Fail unless we have an issuer */
 	if ( ! issuer ) {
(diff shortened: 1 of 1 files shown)


mcb30:
Shortened the debug message slightly and pushed as http://github.com/ipxe/ipxe/commit/68f1914aa

Thanks!

PR #182, opened 2020-12-16
[efi] Use EFI_PCI_IO_PROTOCOL instance for PCI configuration space access

### LINKED ISSUE #173: EFI PCI configuration access not working on system with more PCI segments
I tried iPXE for my own custom PCI Express card on a Dell PowerEdge R7515 server.
I found an interesting problem - the iPXE driver will only find the card in the first PCIe slot.
I examined this issue in more detail and discovered the following:

1) The server uses 4 PCI Roots and these roots share same PCI segment (0)
 
2) First slot belongs PCI Root with index 0

3) Other slots had a different PCI Root

4) The device binding fails in the `efipci_read` function (see the efi_pci.c file) because` root-> Pci.Read` returns an INVALID PARAMETER error.

5) The `efipci_root_open` function pairs the PCI Root with the PCI segment number and thus always finds the first PCI Root

I think this pairing method is bad. The UEFI specification in chapter **PCI Root Bridge I/O Overview** says:

> A PCI Host Bridge and PCI Root Bridge are different than a PCI Segment. A PCI Segment is a
collection of up to 256 PCI busses that share the same PCI Configuration Space. Depending on
the chipset, a single EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL may abstract a portion of a PCI
Segment, or an entire PCI Segment. A PCI Host Bridge may produce one or more PCI Root
Bridges. When a PCI Host Bridge produces multiple PCI Root Bridges, it is possible to have
more than one PCI Segment.

I have no idea how reliably fix this issue.

pe-bo:
This patch is based on an opened EFI PCI device and a EFI_PCI_IO_PROTOCOL instance stored behalf struct pci_device, introduced in the 36dde9b commit.

Fix #173

### DIFF
--- a/src/interface/efi/efi_pci.c
+++ b/src/interface/efi/efi_pci.c
@@ -138,22 +138,6 @@ static void efipci_root_close ( EFI_HANDLE handle ) {
 }
 
 /**
- * Calculate EFI PCI configuration space address
- *
- * @v pci		PCI device
- * @v location		Encoded offset and width
- * @ret address		EFI PCI address
- */
-static unsigned long efipci_address ( struct pci_device *pci,
-				      unsigned long location ) {
-
-	return EFI_PCI_ADDRESS ( PCI_BUS ( pci->busdevfn ),
-				 PCI_SLOT ( pci->busdevfn ),
-				 PCI_FUNC ( pci->busdevfn ),
-				 EFIPCI_OFFSET ( location ) );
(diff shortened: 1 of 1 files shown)


mcb30:
Unfortunately `struct pci_device` is not always embedded within a `struct efi_pci_device`, and so this approach cannot work.

PR #186, opened 2020-12-27
Increase buffer length in efi_devpath_text

b1f6c1c4:
When using iSCSI, the devpath may become extremely long.


NiKiZe:
Any chance you can also include a test that proves and validates this, so that it remains known in the future?

b1f6c1c4:
I think there's no exiting test that tests any EFI component. This PR does not intent to create test for EFI, just making the log from `efi_devpath_text` more complete, rather than truncated.

mcb30:
@b1f6c1c4 The current length (256 characters) is already approximately three lines of text, which is a very long debug message.  Do you have a concrete example of a situation in which it's actually useful to have more than this displayed?

Thanks.

b1f6c1c4:
@mcb30 Please take a look at this error message. I wouldn't be able to know the exact file path until I increase the debug message length.
image

mcb30:
Pushed with a buffer size of 512 as https://github.com/ipxe/ipxe/commit/485f8ce55

Thanks

PR #205, opened 2021-01-13
[isa] include config/isa.h

mmlb:
drivers/bus/isa.c wasn't including any files that would allow
configuration so would default to only being compiled with the
equivalent of:

By including config/isa.h which also includes config/local/isa.h where
build config can be setup.

Signed-off-by: Manuel Mendez <mmendez534@gmail.com>

### DIFF
--- a/src/drivers/bus/isa.c
+++ b/src/drivers/bus/isa.c
@@ -5,6 +5,7 @@
 #include <errno.h>
 #include <ipxe/io.h>
 #include <ipxe/isa.h>
+#include <config/isa.h>
 
 FILE_LICENCE ( GPL2_OR_LATER );
 


mmlb:
This along with `#define ISA_PROBE_ONLY` fixes https://github.com/tinkerbell/boots/pull/117 for me.

mcb30:
The commit message seems to be missing a line in the middle.  I'm guessing an editor stripped out a line starting with a `#` character.

Pushed with an alternative commit message as https://github.com/ipxe/ipxe/commit/a5fb41873 - thank you!

mmlb:
@mcb30 you're welcome! Thanks for providing iPXE. I think your commit stripped my content but meh no big deal.

BTW I came across this when looking into build failure on NixOS's GCC config. I'd like to try and track down the proper fix but C isn't a strength anymore. I'll at least start with an issue though.

mcb30:
> BTW I came across this when looking into build failure on NixOS's GCC config. I'd like to try and track down the proper fix but C isn't a strength anymore. I'll at least start with an issue though.

@mmlb If you can open an issue with clear and easy instructions to reproduce the build failure, that would be perfect.

PR #211, opened 2021-01-16
Add xorrisofs to the mkisofs equivalent list

Omgalof:
Add xorrisofs support, a Gnu mkisofs equivalent that is available in most Distros Repositories.


mcb30:
Fixed to work with EFI images (since `xorrisofs` doesn't support `-efi-boot` as a synonym for `-e`) and pushed as commit https://github.com/ipxe/ipxe/commit/0fb37a48a

Thanks!

PR #218, opened 2021-01-22
WIP: ipxe ConnectX-3 driver feature backport from Mellanox Upstream soup [do not merge as-is]

chrivers:
Hello good ipxe folks

I've backported the ConnectX-3 support from Flexboot sources, and managed to make it work with upstream ipxe.

It has taken a *lot* of tedious cleanup, refactoring, and organizing to find and extract the relevant changes from flexboot.

However, the current state is still far too invasive, but it *does* work, so consider it a PoC.

Eventually, I would very much like to get this merged into mainline, but I fully realize there is a way to go yet.

After talking to mcb30 on irc, it was suggested to open this for a WIP pre-review.

### DIFF
--- a/src/drivers/infiniband/MT25218_PRM.h
+++ b/src/drivers/infiniband/MT25218_PRM.h
@@ -73,7 +73,7 @@ struct arbelprm_ud_address_vector_st {	/* Little Endian */
 /* -------------- */
     pseudo_bit_t	rgid_31_0[0x00020];    /* Remote GID[31:0] if G bit is set. Must be set to 0x2 if G bit is cleared. */
 /* -------------- */
-}; 
+};
 
 /* Send doorbell */
 
@@ -88,7 +88,7 @@ struct arbelprm_send_doorbell_st {	/* Little Endian */
     pseudo_bit_t	reserved1[0x00002];
     pseudo_bit_t	qpn[0x00018];          /* QP number this doorbell is rung on */
 /* -------------- */
-}; 
+};
 
 /* ACCESS_LAM_inject_errors_input_modifier */
 
@@ -102,7 +102,7 @@ struct arbelprm_access_lam_inject_errors_input_modifier_st {	/* Little Endian */
     pseudo_bit_t	index0[0x00007];
     pseudo_bit_t	q0[0x00001];
 /* -------------- */
-}; 
+};
 
 /* ACCESS_LAM_inject_errors_input_parameter */
 
@@ -114,7 +114,7 @@ struct arbelprm_access_lam_inject_errors_input_parameter_st {	/* Little Endian *
     pseudo_bit_t	ra[0x00010];           /* Row Address */
     pseudo_bit_t	ca[0x00010];           /* Column Address */
 /* -------------- */
-}; 
+};
 
 /*  */
 
@@ -127,7 +127,7 @@ struct arbelprm_recv_wqe_segment_next_st {	/* Little Endian */
                                                   */
     pseudo_bit_t	reserved1[0x0001a];
 /* -------------- */
-}; 
+};
 
 /* Send wqe segment data inline */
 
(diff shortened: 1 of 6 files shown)


src/drivers/infiniband/MT25218_PRM.h
```diff
@@ -73,7 +73,7 @@ struct arbelprm_ud_address_vector_st {	/* Little Endian */
 /* -------------- */
     pseudo_bit_t	rgid_31_0[0x00020];    /* Remote GID[31:0] if G bit is set. Must be set to 0x2 if G bit is cleared. */
 /* -------------- */
-}; 
+};
```
mcb30, line 76: All the whitespace-only changes need to be removed

src/drivers/infiniband/MT25408_PRM.h
```diff
@@ -47,7 +47,7 @@ struct hermonprm_ud_address_vector_st {	/* Little Endian */
     pseudo_bit_t	reserved1[0x00008];
 /* -------------- */
     pseudo_bit_t	hop_limit[0x00008];    /* IPv6 hop limit */
-    pseudo_bit_t	max_stat_rate[0x00004];/* Maximum static rate control. 
+    pseudo_bit_t	max_stat_rate[0x00004];/* Maximum static rate control.
```
mcb30, line 50: Again, all the whitespace-only changes need to be removed

mcb30 (changes requested):
OK, so from a quick initial pass through:

- Most of the code relates to UI or behavioural changes between FlexBoot (which is derived from iPXE) and real iPXE, and is therefore unnecessary (and cannot be upstreamed)
- There should be no changes to files outside of `drivers/infiniband`, with the possible exception of `drivers/net/ipoib.c`
- Whitespace-only changes must be removed, and will substantially reduce the diff for some remaining files.
- Even within those validly changed files, the Mellanox way of doing things is often unnecessarily complicated.  For example:
  - I can't (immediately) see where the `event` parameter to `hermon_cmd` ever takes a non-zero value, in which case all of those changes can be removed
  - The `hermon_cmd_post_*` functions seem to be entirely detrimental since ringing the doorbell via MMIO is an order of magnitude faster than ringing the doorbell via the command queue, so all of that logic can be removed
  - Anything in this code relating to the use of settings should be removed: iPXE has more generic mechanisms for configuring things like the initiator IQN.

I'd suggest taking a fairly brutal hacksaw to this code.  I might even consider starting from the existing upstream hermon.c and trying to pull in the minimal bits required to make it work with CX3, rather than trying to remove all of the cruft from this code.  But it's a tough call, since the "remove the cruft" approach has the benefit of being testable as you go.

src/drivers/infiniband/MT25408_PRM.h
mcb30: My mistake, sorry.  I hadn't realised these were nicely separated out into standalone commits

chrivers:
As we discussed on IRC, I think we agree on the way forward, and I'm working to make a new WIP version.

chrivers:
Here's WIP2 - now less than 10% of the size of the first version.

You were quite right - a "brutal hacksaw" was just what the doctor ordered. I managed to clean up a lot of cruft, and keep (mostly) just the required changes.

It's not quite there yet, but this should be much closer to the end result.

Oh, and it boots UEFI both over Ethernet and Infiband.

chrivers:
Just to be clear about the current state:
 - **NOT YET** thoroughly tested (it can UEFI-boot over ethernet and infiniband though)
 - **NOT YET** tested on connectx-2. It MAY have broken all connectx-2 functionality!
 - I intend to clean up the history once I'm sure of the final form. Some of the existing commits add features to `hermon.h` that have since been removed, for example
 - Other misc cleanups pending, also.

chrivers:
I derped while rebasing, so this one is b0rked - wip4 coming up shortly!

chrivers:
Eh, the build fails, but only because of a warning I don't get locally. I'll look into it.

These are the only missing pieces now:
 - Minor cleanups
 - A few architectural questions to resolve with mcb30
 - ConnectX-2 testing

chrivers:
Accidentally closed instead of updated - please see https://github.com/ipxe/ipxe/pull/228

PR #225, opened 2021-01-26
[shell] Make "ifstat" show the link-layer protocol for each netdev.

chrivers:
This is useful on devices that perform auto-detection for ports. Example output:

    iPXE> ifstat
    net0: 00:11:22:33:44:55 using mt4099 on 0000:00:03.0 (Ethernet) [open]
      [Link:down, TX:0 TXE:0 RX:0 RXE:0]
      [Link status: Unknown (http://ipxe.org/1a086101)]
    net1: 00:11:22:33:44:56 using mt4099 on 0000:00:03.0 (IPoIB) [open]
      [Link:down, TX:0 TXE:0 RX:0 RXE:0]
      [Link status: Initialising (http://ipxe.org/1a136101)]

### DIFF
--- a/src/usr/ifmgmt.c
+++ b/src/usr/ifmgmt.c
@@ -108,10 +108,11 @@ static void ifstat_errors ( struct net_device_stats *stats,
  * @v netdev		Network device
  */
 void ifstat ( struct net_device *netdev ) {
-	printf ( "%s: %s using %s on %s (%s)\n"
+	printf ( "%s: %s using %s on %s (%s) [%s]\n"
 		 "  [Link:%s%s, TX:%d TXE:%d RX:%d RXE:%d]\n",
 		 netdev->name, netdev_addr ( netdev ),
 		 netdev->dev->driver_name, netdev->dev->name,
+		 netdev->ll_protocol->name,
 		 ( netdev_is_open ( netdev ) ? "open" : "closed" ),
 		 ( netdev_link_ok ( netdev ) ? "up" : "down" ),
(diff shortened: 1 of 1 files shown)


src/usr/ifmgmt.c
```diff
@@ -108,10 +108,11 @@ static void ifstat_errors ( struct net_device_stats *stats,
  * @v netdev		Network device
  */
 void ifstat ( struct net_device *netdev ) {
-	printf ( "%s: %s using %s on %s (%s)\n"
+	printf ( "%s: %s using %s on %s (%s) [%s]\n"
 		 "  [Link:%s%s, TX:%d TXE:%d RX:%d RXE:%d]\n",
 		 netdev->name, netdev_addr ( netdev ),
 		 netdev->dev->driver_name, netdev->dev->name,
+		 netdev->ll_protocol ? netdev->ll_protocol->name : "no link-layer",
```
mcb30, line 115: `netdev->ll_protocol` can never be NULL, so this check can be removed

mcb30 (changes requested):
I'm very mildly concerned about rendering large amounts of existing documentation slightly obsolete by changing the format, but I can't find a neater way to encapsulate both link-layer type and open/closed state, and it **does** make some sense to place the open/closed state within square brackets along with the other runtime state information.

src/usr/ifmgmt.c
chrivers: Done

mcb30:
Merged (with trailing full stop stripped from subject line) as https://github.com/ipxe/ipxe/commit/f4c3a0147

Thanks!

PR #226, opened 2021-01-26
[infiniband] Require drivers to report the number of device ports

chrivers:
Allow drivers to override total number of reported infiband ports. This is
required to report the correct number of total ports, on devices with dynamic
port types.

For example, dual-port Mellanox cards configured for (eth, ib) would be rejected
by the subnet manager, because they report using "port 2, out of 1".

### DIFF
--- a/src/drivers/infiniband/arbel.c
+++ b/src/drivers/infiniband/arbel.c
@@ -3061,6 +3061,7 @@ static int arbel_probe ( struct pci_device *pci ) {
 		ibdev->op = &arbel_ib_operations;
 		ibdev->dev = &pci->dev;
 		ibdev->port = ( ARBEL_PORT_BASE + i );
+		ibdev->ports = ARBEL_NUM_PORTS;
 		ib_set_drvdata ( ibdev, arbel );
 	}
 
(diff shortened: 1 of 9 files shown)


mcb30:
Merged with minor wording fixes as https://github.com/ipxe/ipxe/commit/b9de7e6ed - thank you!

Incidentally, is there a clean way to force GitHub to realise that this class of PR (i.e. one that would disappear completely on performing a `git rebase`) has been merged?

NiKiZe:
> Merged with minor wording fixes as [b9de7e6](https://github.com/ipxe/ipxe/commit/b9de7e6ed) - thank you!
> 
> Incidentally, is there a clean way to force GitHub to realise that this class of PR (i.e. one that would disappear completely on performing a `git rebase`) has been merged?

I guess you don't want to use "closes #226" in the commit message?
You might be able to force push the actual commit to the branch that the PR was created from (depending on settings you should have write access to that branch)

mcb30:
With the current branch protection settings (only rebase merging allowed, require linear history), it seems as though the GitHub "rebase and merge" button will mark the PR as merged but won't create a pointless merge commit (which contradicts the documented 'equivalent' command-line instructions).

So I think I can safely use the "Rebase and merge" button as long as I am confident that a `git rebase` on the PR would result in an empty set of commits.

PR #228, opened 2021-01-26
iPXE ConnectX-3 driver support

chrivers:
Whoops - due to a typo I accidentally deleted the branch for the previous in-progress PR, instead of pushing an update.

This is the cleaned-up version of the pull request, which should be good to merge, unless there are any requests for changes.

I cleaned up the commit history, so this should be fairly clean, IMHO.

### DIFF
--- a/src/drivers/infiniband/hermon.c
+++ b/src/drivers/infiniband/hermon.c
@@ -4232,6 +4232,10 @@ static struct pci_device_id hermon_nics[] = {
 	PCI_ROM ( 0x15b3, 0x675a, "mt26458", "MT26458 HCA driver", 0 ),
 	PCI_ROM ( 0x15b3, 0x6764, "mt26468", "MT26468 HCA driver", 0 ),
 	PCI_ROM ( 0x15b3, 0x676e, "mt26478", "MT26478 HCA driver", 0 ),
+
+	/* Mellanox ConnectX-3 VPI (ethernet + infiniband) */
+	PCI_ROM ( 0x15b3, 0x1003, "mt4099", "ConnectX-3 HCA driver", 0 ),
+	PCI_ROM ( 0x15b3, 0x1007, "mt4103", "ConnectX-3 Pro HCA driver", 0 ),
 };
 
 struct pci_driver hermon_driver __pci_driver = {


src/drivers/infiniband/hermon.h
```diff
@@ -926,8 +926,9 @@ struct hermon {
 #define HERMON_HCR_BASE			0x80680
 #define HERMON_HCR_REG(x)		( HERMON_HCR_BASE + 4 * (x) )
 #define HERMON_HCR_MAX_WAIT_MS		2000
-#define HERMON_MBOX_ALIGN		4096
-#define HERMON_MBOX_SIZE		1024
+#define HERMON_MBOX_ALIGN		(1 << 12)
+/* MELLANOX-FIXME: HERMON_MBOX_SIZE is 1<<12 in flexboot upstream */
+#define HERMON_MBOX_SIZE		(1 << 10)
```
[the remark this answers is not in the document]
chrivers: No, you're quite right. It was a bit of leftover cruft.

In the first iteration, the values were not the same, but it turned out that *that* served no purpose. Feel free to drop this change.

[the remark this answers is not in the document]
chrivers: I could also roll a new patchset, when you've had a chance to look through the rest.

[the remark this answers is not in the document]
chrivers: Nevermind, I just looked at the patchset. Commit a253425049c81730d75d0015e0f83ce0662c5a4 should indeed simply be dropped.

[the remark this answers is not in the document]
chrivers: The new version of the patchset has dropped these.

src/drivers/infiniband/hermon.c
```diff
@@ -137,13 +137,13 @@ static int hermon_cmd_wait ( struct hermon *hermon,
...
 
-	for ( wait = HERMON_HCR_MAX_WAIT_MS ; wait ; wait-- ) {
+	for ( wait = 100 * HERMON_HCR_MAX_WAIT_MS ; wait ; wait-- ) {
 		hcr->u.dwords[6] =
 			readl ( hermon->config + HERMON_HCR_REG ( 6 ) );
 		if ( ( MLX_GET ( hcr, go ) == 0 ) &&
 		     ( MLX_GET ( hcr, t ) == hermon->toggle ) )
 			return 0;
-		mdelay ( 1 );
+		udelay ( 10 );
```
mcb30, line 146: Does this make a noticeable difference to anything?  On my test system (CX2 EN) I see a total of 44 commands issued to get to the point that the card is sending and receiving packets, which would give you a maximum saving of 44ms.

chrivers: Do you use a dual-port card? I see a lot more activity. I'll try to add some debug stmts to figure out if there's something to it

chrivers: Yeah, just checked. I very quickly got up to tens of thousands of commands.

Whatever logic is polling the link status is... determined.

chrivers: I added this code to `hermon_cmd()`:

```c
	static int num = 0;
	num++;
	TR("Command[%5d]: %08x", num++, opcode);
```

This is the example output after a couple of seconds:

```
[drivers/infiniband/hermon.c/hermon_cmd:211] Command[11755]: 00000024
```

The explosive amount of commands all seem to be opcode 0x24 - `HERMON_HCR_MAD_IFC` ?

This might indicate a deeper problem..

chrivers: `HERMON_HCR_MAD_IFC` is used in `hermon_mad()`

This might trigger it?

```
	/* No event is generated upon reaching INIT, so we must poll
	 * separately for link state changes while we remain DOWN.
	 */
	if ( ib_is_open ( ibdev ) &&
	     ( ibdev->port_state == IB_PORT_STATE_DOWN ) ) {
		ib_smc_update ( ibdev, hermon_mad );
	}
```

That's from `hermon_poll_eq()`.. so on every event queue poll, we ask the firmware to send a mgmt datagram. That doesn't really seem right, does it?

chrivers: This thing about no event generated for reaching "INIT", does that refer to full online infiniband, or physical link-up?

I have verified that (at least CX3) happily generates link state change events when physical link goes up.

I haven't touched this part of the code, btw, so it seems like a "pre-existing condition", as it were. I will concede that it's looking like the command wait logic does not need to be more agressive, if we would stop posting so many commands.

src/drivers/infiniband/hermon.c
```diff
@@ -857,6 +857,47 @@ hermon_dump_cqctx ( struct hermon *hermon, struct ib_completion_queue *cq ) {
...
+			       MLX_GET ( &eqe->generic, owner ),
+			       MLX_GET ( &eqe->generic, event_type ));
+			DBGC_HDA ( hermon, 0, &eqe->generic, 32 );
+		}
+	}
+	/* DBGC_HDA ( hermon, 0, &eqctx, sizeof ( eqctx ) ); */
+
+	return 0;
+}
+
```
mcb30, line 907: Replaced with https://github.com/ipxe/ipxe/commit/7d32225b5 which creates `hermon_dump_eqctx()` as a direct analogue of `hermon_dump_cqctx()` and `hermon_dump_qpctx()` and splits out the EQE dumping code to a separate `hermon_dump_eqes()` function.

src/drivers/infiniband/hermon.c
chrivers: I'm doing some more testing to figure out what's up with this

src/drivers/infiniband/hermon.c
chrivers: Looks good to me - can we keep the "Hermon init done" debug line below?

src/drivers/infiniband/hermon.c
mcb30: I'm pretty sure this is going to be Infiniband-only, which explains why I'm not seeing it when testing on a CX2 EN.

The "no event generated" was definitely a real problem on at least some hardware supported by `hermon.c`: commit https://github.com/ipxe/ipxe/commit/14a76b592 and my handwritten notes from 2010-09-14 confirm the existence of the problem.

So, we **do** need to perform this `ib_smc_update()` polling while DOWN since at least some versions of the supported hardware will required it.  It's not actually sending anything over the wire: it's just sending something that looks like a management datagram to the firmware since that's the method by which software and firmware communicate.

I don't really want to start guessing as to which versions of the hardware (and very possibly firmware) require this workaround, so we're going to have to keep it for all supported cards.

I'll merge this commit with a message updated to explain the situation in which it's needed.

chrivers: Agreed - guessing is never fun.

I'm going to test if CX3 requires this workaround. If not, we could disable it for CX3 cards? I'll do that after the merge then.

chrivers: We could really consider throttling the number of smc updates, in any case. There's probably zero advantage of hammering the firmware with several thousand requests per second. If we did, say, 5-10 checks per second, we wouldn't reasonably need the more agressive cmd wait logic.

mcb30: Reworded as https://github.com/ipxe/ipxe/commit/6cb12ee2b

src/drivers/infiniband/hermon.c
```diff
@@ -3936,6 +3927,23 @@ static int hermon_probe ( struct pci_device *pci ) {
...
+			       hermon, i, strerror ( rc ) );
+			goto err_query_port;
+		}
+
+		/* The MAC addresses are stored as consecutive bytes, so we have
+		 * to undo the implicit byte-swap in MLX_GET */
+		port->eth_mac.part.h = cpu_to_be16 ( MLX_GET ( &query_port, mac_47_32 ) );
+		port->eth_mac.part.l = cpu_to_be32 ( MLX_GET ( &query_port, mac_31_0 ) );
+	}
+
```
mcb30, line 4012: Modified and merged as https://github.com/ipxe/ipxe/commit/699b9f1d1.  I realised that we could capture the MAC during `hermon_set_port_type()` and save a duplicate `hermon_cmd_query_port()`.  Also reworded the commit message to mention the potential breaking change for anyone who had DHCP reservations based on the old LEMAC.

src/drivers/infiniband/hermon.c
```diff
@@ -2097,7 +2097,7 @@ static void hermon_destroy_eq ( struct hermon *hermon ) {
  * @v eqe		Port state change event queue entry
  */
 static void hermon_event_port_state_change ( struct hermon *hermon,
-					     union hermonprm_event_entry *eqe){
+					     union hermonprm_event_entry *eqe ) {
```
mcb30, line 2118: This pushes the line length over 80 characters, sorry.  There are several other places where I use this to avoid overlength lines, e.g. in `hermon_cmd_sw2hw_cq()`

src/drivers/infiniband/hermon.c
chrivers: That commit link doesn't work for me, currently

chrivers: It does now - that looks like a good solution!

src/drivers/infiniband/hermon.c
mcb30: I'm happy to leave it as-is (now that https://github.com/ipxe/ipxe/commit/6cb12ee2b is merged) on the basis of keeping it simple.  In the common case, the link will stay in DOWN for only a very short length of time, unless the user is trying to use a port for which the cable is physically unplugged.

src/drivers/infiniband/hermon.c
```diff
@@ -2097,6 +2219,7 @@ static void hermon_poll_eq ( struct ib_device *ibdev ) {
 		/* Ring doorbell */
 		MLX_FILL_1 ( &db_reg.event, 0,
 			     ci, ( hermon_eq->next_idx & 0x00ffffffUL ) );
+		barrier();
```
mcb30, line 2222: I don't see any memory barrier in `hermon_poll_cq()`.  Did you mean `hermon_post_send()`?  If so then:

- the `barrier()` before `writel()` in `hermon_post_send()` is to ensure that the WQE is flushed before ringing the doorbell
- no similar requirement exists in `hermon_poll_eq()` since it does not modify DMA-accessible host memory: the EQEs are read-only as far as software is concerned.

chrivers: I misread the code - patch withdrawn.

src/drivers/infiniband/hermon.c
chrivers: That's.. one way to do it ;-)

Patch reworked.

src/drivers/infiniband/hermon.c
chrivers: I reworked the patch with just the "init done" message.

src/drivers/infiniband/hermon.c
```diff
@@ -2162,6 +2225,17 @@ static void hermon_poll_eq ( struct ib_device *ibdev ) {
...
+			hermon_event_port_mgmt_change ( hermon, eqe );
+			break;
+		case HERMON_EV_CQ_ERROR:
+			DBGC ( hermon, "Hermon %p EQN %#lx CQ ERROR\n",
+			       hermon, hermon_eq->eqn );
+			break;
+		case HERMON_EV_WQ_CATAS_ERROR:
+			DBGC ( hermon, "Hermon %p EQN %#lx WQ CATAS ERROR\n",
+			       hermon, hermon_eq->eqn );
+			break;
```
mcb30, line 2238: I've merged the port management bits as https://github.com/ipxe/ipxe/commit/8b07c88df (after making the naming consistent between "mgmt" and "mgmnt", and matching the code style of `hermon_event_port_state_change()`).

I'm not convinced by the value of the `HERMON_EV_CQ_ERROR` and `HERMON_EV_WQ_CATAS_ERROR` handlers.  The information that they would provide is already visible in the `default` case, which also provides a hex dump to allow for more detailed debugging.

src/drivers/infiniband/hermon.c
```diff
@@ -3926,6 +4021,8 @@ static int hermon_probe ( struct pci_device *pci ) {
 	hermon->uar = pci_ioremap ( pci, uar,
 				    HERMON_UAR_NON_EQ_PAGE * HERMON_PAGE_SIZE );
 
+	hermon->physical_function = pci->busdevfn & 0x7;
```
mcb30, line 4024: This is not a valid way to obtain the physical function number: see e.g. commit https://github.com/ipxe/ipxe/commit/76a7bfe93

mcb30 (commented):
@chrivers I think I've cherry-picked all of the uncontroversial parts (some with edits).  Could you rebase so that we can see what remains?

Thanks!

src/drivers/infiniband/hermon.c
chrivers: Okay, I can see that. I'm sure we can just drop this patch. It was included because I felt confident in lifting the change from Flexboot, and it seemed like someone might get some benefit from it. But it's not actually clear to me what cards this target.

Do you know if ConnectX-4 cards have multiple physfuncs?

chrivers:
Yeah, I'll take a look at rebasing

src/drivers/infiniband/hermon.c
chrivers: 8b07c88 lgtm!

I don't have strong feelings about HERMON_EV_CQ_ERROR / HERMON_EV_WQ_CATAS_ERROR. They were useful for debugging, just because the raw numbers are not as easy to recognize the severity of.

I'm fine with dropping these debugs.

src/drivers/infiniband/hermon.c
chrivers: I dropped the `physical_function` patch in the newest re-roll.

chrivers:
I've rebased the patchset, and we are down to 5 patches. Here are my comments on them individually:

683e6d3 generation tagging: needed for next patch. 
e639ce8 generation-specific reset sleep. important to gain back a little speed for cx3 cards (and now simpler, because of the reset merge)
5582dc7 Improved recv error handling.. At least I *think* this is an improvement. @mcb30, what's your take? 
7e3adf0 Reworking the reset handling. I'll argue for this in the next comment.
f9bdd42 The cherry on top :stuck_out_tongue:

src/drivers/infiniband/hermon.c
mcb30: > I don't have strong feelings about HERMON_EV_CQ_ERROR / HERMON_EV_WQ_CATAS_ERROR. They were useful for debugging, just because the raw numbers are not as easy to recognize the severity of.

Any unrecognised event is cause for concern :)

> I'm fine with dropping these debugs.

Ack.

chrivers:
This should be the final roll, of the final patch :)

The other 4 patches in the previous set have all been either handled in master, or dropped.

PR #232, opened 2021-01-27
[cmdline] Allow "show" to display multiple settings at once

chrivers:
When testing or debugging iPXE, it's very common to want to inspect more than one setting at a time. For convenience, it is now possible to display a whole set of options at once, by specifying the scope name, by including a trailing slash.

Before:

    iPXE> show net0/ip
    net0/ip:ipv4 = 1.2.3.4
    iPXE> show net0/netmask
    net0/netmask:ipv4 = 255.255.255.0

After:

    iPXE> show net0/
    net0/busid = 01:80:86:10:0e
    net0/busloc = 0x10
    net0/bustype = PCI
    net0/chip = 82540em
    net0/hwaddr = 52:54:00:12:34:56
    net0/ifname = net0
    net0/mac = 52:54:00:12:34:56
    net0/ip = 1.2.3.4
    net0/netmask = 255.255.255.0
    net0/ip6 = fe80::5054:ff:fe12:3456
    net0/len6 = 64

### DIFF
--- a/src/hci/commands/nvo_cmd.c
+++ b/src/hci/commands/nvo_cmd.c
@@ -63,6 +63,7 @@ static int show_exec ( int argc, char **argv ) {
 	struct named_setting setting;
 	struct settings *origin;
 	struct setting fetched;
+	struct setting *iter;
 	char name_buf[32];
 	char *value;
 	int len;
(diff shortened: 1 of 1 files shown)


berlin4apk:
@chrivers can you extend this to 
Allow "show" to display all settings at once?
(Like set or env on posix)

stappersg:
Why is this merge request still open?


I'm asking because dangling PRs do drain human energy.
We, the iPXE project, pay a price each time when we see open merge requests like this one.

It would be much better if
 * the proposed patch gets merged
 * rejected with a "why"
 * bluntly rejected
 * is explained why the MR is still open


And congratulations for an year being tagged with 'enhancement'

afbeelding

PR #234, opened 2021-01-28
Make iPXE builds reproducible

mmlb:
These 2 commits seem to be all that is needed to make iPXE builds reproducible. I'd love to have reproducible iPXE builds so that I could verify CI builds against local ones for https://github.com/tinkerbell/boots.

### DIFF
--- a/src/Makefile.housekeeping
+++ b/src/Makefile.housekeeping
@@ -1202,7 +1202,7 @@ $(BIN)/version.%.o : core/version.c $(MAKEDEPS) $(GIT_INDEX)
 $(BIN)/%.tmp : $(BIN)/version.%.o $(BLIB) $(MAKEDEPS) $(LDSCRIPT)
 	$(QM)$(ECHO) "  [LD] $@"
 	$(Q)$(LD) $(LDFLAGS) -T $(LDSCRIPT) $(TGT_LD_FLAGS) $< $(BLIB) -o $@ \
-		--defsym _build_id=`$(BUILD_ID_CMD)` \
+		--defsym _build_id=$(shell $(BUILD_ID_CMD)) \
 		--defsym _build_timestamp=$(BUILD_TIMESTAMP) \
 		-Map $(BIN)/$*.tmp.map
 	$(Q)$(OBJDUMP) -ht $@ | $(PERL) $(SORTOBJDUMP) >> $(BIN)/$*.tmp.map
--- a/src/Makefile.housekeeping
+++ b/src/Makefile.housekeeping
@@ -1167,7 +1167,17 @@ BUILD_ID_CMD	:= perl -e 'printf "0x%08x", int ( rand ( 0xffffffff ) );'
 
 # Build timestamp
 #
+# Used only as a means to automatically select the newest version of iPXE if
+# multiple iPXE drivers are loaded concurrently in a UEFI system.
+#
+# It gets rounded down to the nearest minute when used for this purpose.
+ifdef SOURCE_DATE_EPOCH
+BUILD_TIMESTAMP := $(SOURCE_DATE_EPOCH)
+else ifdef GITVERSION
+BUILD_TIMESTAMP := $(git log -1 --pretty=%ct)
+else
 BUILD_TIMESTAMP := $(shell date +%s)
+endif
 
 # Build version
 #
--- a/src/Makefile.housekeeping
+++ b/src/Makefile.housekeeping
@@ -1160,10 +1160,15 @@ $(BLIB) : $(BLIB_OBJS) $(BLIB_LIST) $(MAKEDEPS)
 	$(Q)$(RANLIB) -D $@
 blib : $(BLIB)
 
-# Command to generate build ID.  Must be unique for each $(BIN)/%.tmp,
-# even within the same build run.
+# Build ID
 #
-BUILD_ID_CMD	:= perl -e 'printf "0x%08x", int ( rand ( 0xffffffff ) );'
+# Must be unique for each $(BIN)/%.tmp, even within the same build run.
+#
+# The build ID is supposed to be collision-free across all ROMs that might ever
+# end up installed in the same system. It doesn't just disambiguate targets
+# within a single build; it also disambiguates different builds (such as
+# builds for multiple roms, which are all built from the same blib.a).
+BUILD_ID_CMD = sha1sum $^ | sort | cksum | awk '{printf "0x%08x", $$1}'
 
 # Build timestamp
 #
@@ -1173,8 +1178,8 @@ BUILD_ID_CMD	:= perl -e 'printf "0x%08x", int ( rand ( 0xffffffff ) );'
 # It gets rounded down to the nearest minute when used for this purpose.
 ifdef SOURCE_DATE_EPOCH
 BUILD_TIMESTAMP := $(SOURCE_DATE_EPOCH)
-else ifdef GITVERSION
-BUILD_TIMESTAMP := $(git log -1 --pretty=%ct)
+else ifneq ($(GITVERSION),)
+BUILD_TIMESTAMP := $(shell git log -1 --pretty=%ct)
 else
 BUILD_TIMESTAMP := $(shell date +%s)
 endif


mmlb:
I just had the thought of maybe making it so that I could specify BUILD_ID_CMD externally instead of wholesale replacing it like I do here. Let me know if you'd like me to do that instead and I can change the PR.

NiKiZe:
There has been several that has proposed similar things, all ignoring the issue of that Id should be unique.

I think there is issues/prs on GitHub, but also on the mailing list that you might want to read.

Also read the comment about the id that you removed in this pr.

mmlb:
@NiKiZe I admit to not checking issues or mailing list and thats definitely something I'll do. I did read the comment and pondered it. I laid out my thoughts in the commit message where I think this should be fine.

That being said I'll check out the other attempts at this.

mmlb:
I just looked at https://github.com/ipxe/ipxe/pull/82, that implementation hashes the target (which doesn't exist yet?), mine hashes all the input files. This should be good enough imo. If all the inputs are the same and the build is reproducible then you'd get a bit for bit copy of the binary and that would be no different than what build_id is trying to protect against I think.

mmlb:
I guess if the requirement truly is that iPXE should not be reproducible I'd like to know a bit more of the why, which I'd use to update the BUILD_ID_CMD comment. My reading of the original commit is that NICs may offer up the same iPXE ROM binary multiple times and _someone_ wants to be able to tell the difference between this offering up of the same binary vs another one?

The Makeifle.housekeeping comment says

> Must be unique for each $(BIN)/%.tmp, even within the same build run

which means what exactly? `make bin-x86_64-efi/ipxe.efi bin-x86_64-efi/ipxe.efi` should build twice and each should have a different _build_id? Why would this matter if one is going to be immediage overwritten with the last built one?

mmlb:
Going through the mailing list I find:

https://lists.ipxe.org/pipermail/ipxe-devel/2020-May/007031.html

>  The build ID is supposed to be collision-free across all ROMs that might 
ever end up installed in the same system.  It doesn't just disambiguate 
targets within a single build; it also disambiguates different builds.

https://lists.ipxe.org/pipermail/ipxe-devel/2020-May/007033.html

> Different build configurations (e.g. config/*.h setting, or even 
DEBUG=... options specified on the make command line) also count as 
needing different build IDs.  Basically: anything that could possibly 
affect the runtime memory layout is relevant.
>
> It should really be a hash of the object that results from the link.

I'll see if my approach ends up taking account of DEBUG= and other make invocation settings.

And https://lists.ipxe.org/pipermail/ipxe-devel/2020-May/007038.html

> BUILD_TIMESTAMP has a weaker requirement than BUILD_ID: it gets used 
only as a means to automatically select the newest version of iPXE if 
multiple iPXE drivers are loaded concurrently in a UEFI system.  It's 
already rounded down to the nearest minute when used for that purpose.
>
> For BUILD_TIMESTAMP (but not BUILD_ID), I am happy with using the commit 
date as extracted from git.

Which I'll happily add to the commit message and likely also Makefile comment.

mmlb:
Ideally I'd just patch the resulting .tmp file, but I only know of nix's patchelf which likely doesn't work on non-elf files. I'll look into binutils.

NiKiZe:
I found this thread https://www.mail-archive.com/ipxe-devel@lists.ipxe.org/msg06874.html

So, embed, debug, config etc all need to be part of the id.
Not saying what you suggest is missing any of that.
But this has come up several times, 4-5 at least.

Could the BUILD_ID_CMD be kept, add additional explanations to the comment (not remove) to make it clear about the requirements, and reduce the risk of it being switched to something that could cause issues

mmlb:
Yep I think keeping BUILD_ID_CMD and expanding the comment is something I'll do.

mcb30:
> The Makeifle.housekeeping comment says
> 
> > Must be unique for each $(BIN)/%.tmp, even within the same build run
> 
> which means what exactly? `make bin-x86_64-efi/ipxe.efi bin-x86_64-efi/ipxe.efi` should build twice and each should have a different _build_id? Why would this matter if one is going to be immediage overwritten with the last built one?

The requirement for uniqueness arises because we may be linking multiple **different** output binaries from the same set of compiled objects within a single build run.  For example: `make bin/8086100e.rom bin/10ec8139.rom`.  These must not have the same build ID, even though they are built from the same `blib.a` archive (containing all possible binary objects), because the set of linked-in objects (determined solely by `TGT_LD_FLAGS`) will differ.

As of commit https://github.com/ipxe/ipxe/commit/8290a9551 I **think** that including `$(BIN)/%.version.o` within the hash input (as you do) is probably sufficient, because `TGT_LD_FLAGS` is essentially constructed from the build name, which is included within `$(BIN)/%.version.o` as the `build_name` string.

mcb30:
`EMBED=` affects the contents of `$(BIN)/embedded.o` and hence `$(BLIB)`

`DEBUG=` affects the set of objects that form `$(BLIB)`

Definitions in `config/*.h` would affect the relevant objects that form `$(BLIB)`

So I **think** this is the first version of this long-running suggestion that would actually work.

src/Makefile.housekeeping
```diff
@@ -1187,7 +1186,7 @@ $(BIN)/version.%.o : core/version.c $(MAKEDEPS) $(GIT_INDEX)
 $(BIN)/%.tmp : $(BIN)/version.%.o $(BLIB) $(MAKEDEPS) $(LDSCRIPT)
 	$(QM)$(ECHO) "  [LD] $@"
 	$(Q)$(LD) $(LDFLAGS) -T $(LDSCRIPT) $(TGT_LD_FLAGS) $< $(BLIB) -o $@ \
-		--defsym _build_id=`$(BUILD_ID_CMD)` \
+		--defsym _build_id=$(shell sha1sum $^|sort|cksum|awk '{printf "0x%08x", $$1}') \
```
mcb30, line 1189: Fix this to avoid exceeding 80 characters, and I'm happy to merge this :)

mcb30 (changes requested):

mmlb:
> `EMBED=` affects the contents of `$(BIN)/embedded.o` and hence `$(BLIB)`
> 
> `DEBUG=` affects the set of objects that form `$(BLIB)`
> 
> Definitions in `config/*.h` would affect the relevant objects that form `$(BLIB)`

Yep from my reading of the Makefile I think these configs all make it into the input hash.

mmlb:
> > The Makeifle.housekeeping comment says
> > > Must be unique for each $(BIN)/%.tmp, even within the same build run
> > 
> > 
> > which means what exactly? `make bin-x86_64-efi/ipxe.efi bin-x86_64-efi/ipxe.efi` should build twice and each should have a different _build_id? Why would this matter if one is going to be immediage overwritten with the last built one?
> 
> The requirement for uniqueness arises because we may be linking multiple **different** output binaries from the same set of compiled objects within a single build run. For example: `make bin/8086100e.rom bin/10ec8139.rom`. These must not have the same build ID, even though they are built from the same `blib.a` archive (containing all possible binary objects), because the set of linked-in objects (determined solely by `TGT_LD_FLAGS`) will differ.
> 
> As of commit [8290a95](https://github.com/ipxe/ipxe/commit/8290a9551) I **think** that including `$(BIN)/%.version.o` within the hash input (as you do) is probably sufficient, because `TGT_LD_FLAGS` is essentially constructed from the build name, which is included within `$(BIN)/%.version.o` as the `build_name` string.

I'll check the _build_id with `make V=1 ...`, which should be different and add confidence to uniqueness.

src/Makefile.housekeeping
mmlb: I'll fix this too before resubmitting updates.

mmlb:
@mcb30 Are you ok with the SOURCE_DATE_EPOCH like this or should I also change the `date` call to first attempting to check with git? Something like

```makefile
ifdef (SOURCE_DATE_EPOCH)
BUILD_TIMESTAMP=$(SOURCE_DATE_EPOC)
else
ifne ($(shell git ...),)
BUILD_TIMESTAMP=$(shell git...)
else
BUILD_TIMESTAMP=$(shell date %s)
endif
endif
```

mcb30:
> @mcb30 Are you ok with the SOURCE_DATE_EPOCH like this or should I also change the `date` call to first attempting to check with git? Something like

I'm fine with the `SOURCE_DATE_EPOCH` patch as-is.

If you want to allow for reproducible builds without having to define `SOURCE_DATE_EPOCH` then feel free to extract something sensible from git, whatever that might be.  (https://reproducible-builds.org/docs/source-date-epoch/ seems to suggest `git log -1 --pretty=%ct`).  The build system already defines `GITVERSION` iff building from a git tree, so you might want to use that as the test.

Thanks.

src/Makefile.housekeeping
bmwiedemann: While this is very similar to my #82, there is a difference: the `sha1sum $^` will include the filenames in output - if those are absolute pathnames(are they?), it would mean that a binary built in `/home/bernhard` would be different from yours built from identical sources.
and for some reason I still get 4 random bytes in the binaries - maybe because openSUSE's `.o` files are from LTO and gcc's intermediate representations have randomness. I'll retest without LTO.

mcb30: > While this is very similar to my #82, there is a difference: the `sha1sum $^` will include the filenames in output - if those are absolute pathnames(are they?), it would mean that a binary built in `/home/bernhard` would be different from yours built from identical sources.

Good point.  @mmlb could you please find a way to omit filenames from the resulting hash value?

bmwiedemann:
I had good results with https://github.com/ipxe/ipxe/compare/master...bmwiedemann:rb - only ipxe.iso left with timestamps

mmlb:
> I had good results with [master...bmwiedemann:rb](https://github.com/ipxe/ipxe/compare/master...bmwiedemann:rb) - only ipxe.iso left with timestamps

Which iso tool do you end up using? I ran xorriso and get:
```
GNU xorriso 1.5.2 : RockRidge filesystem manipulator, libburnia project.

xorriso : NOTE : Environment variable SOURCE_DATE_EPOCH encountered with value 315532800
```

mcb30:
> I had good results with [master...bmwiedemann:rb](https://github.com/ipxe/ipxe/compare/master...bmwiedemann:rb) - only ipxe.iso left with timestamps

@bmwiedemann Thanks.  I've merged https://github.com/ipxe/ipxe/commit/f4fcd000a from that series, since that commit at least seems uncontroversial and finalised.

> Which iso tool do you end up using? I ran xorriso and get:

@mmlb FYI: `xorriso` will be used only if nothing quieter can be found.  (I couldn't find a way to get `xorriso` to not spit out a startup banner, even when `-quiet` is used.)

src/Makefile.housekeeping
mmlb: Ooops didn't see this until now. Let me check on this. I don't think they're absolute filenames though.

mmlb: Nope these are all relative paths. I checked by running trying out:

```diff
diff --git a/src/Makefile.housekeeping b/src/Makefile.housekeeping
index 83665f83..2414fbcf 100644
--- a/src/Makefile.housekeeping
+++ b/src/Makefile.housekeeping
@@ -1201,6 +1201,7 @@ $(BIN)/version.%.o : core/version.c $(MAKEDEPS) $(GIT_INDEX)
 #
 $(BIN)/%.tmp : $(BIN)/version.%.o $(BLIB) $(MAKEDEPS) $(LDSCRIPT)
 	$(QM)$(ECHO) "  [LD] $@"
+	sha1sum $^| sort | cksum | awk '{printf "0x%08x", $$1}'
 	$(Q)$(LD) $(LDFLAGS) -T $(LDSCRIPT) $(TGT_LD_FLAGS) $< $(BLIB) -o $@ \
 		--defsym _build_id=$(shell $(BUILD_ID_CMD)) \
 		--defsym _build_timestamp=$(BUILD_TIMESTAMP) \
 
```
and running `make bin-x86_64-efi/ipxe.efi` (in src/) and from my home doing `make -C /home/manny/cloned/ipxe/ipxe/src/ bin-x86_64-efi/ipxe.efi` which gave me:

```
make: Entering directory '/home/manny/cloned/ipxe/ipxe/src'
make: /bin/echo: No such file or directory
make: /bin/echo: No such file or directory
  [VERSION] bin-x86_64-efi/version.ipxe.efi.o
  [LD] bin-x86_64-efi/ipxe.efi.tmp
sha1sum bin-x86_64-efi/version.ipxe.efi.o bin-x86_64-efi/blib.a Makefile Makefile.housekeeping .echocheck arch/x86_64/Makefile arch/x86/Makefile arch/x86_64/Makefile.efi arch/x86/Makefile.efi Makefile.efi bin-x86_64-efi/.fnrec.state scripts/efi.lds| sort | cksum | awk '{printf "0x%08x", $1}'
0x54378e64  [FINISH] bin-x86_64-efi/ipxe.efi
rm bin-x86_64-efi/version.ipxe.efi.o
make: Leaving directory '/home/manny/cloned/ipxe/ipxe/src'
```

src/Makefile.housekeeping
```diff
@@ -1167,7 +1167,17 @@ BUILD_ID_CMD	:= perl -e 'printf "0x%08x", int ( rand ( 0xffffffff ) );'
...
 # Build timestamp
 #
+# Used only as a means to automatically select the newest version of iPXE if
+# multiple iPXE drivers are loaded concurrently in a UEFI system.
+#
+# It gets rounded down to the nearest minute when used for this purpose.
+ifdef SOURCE_DATE_EPOCH
+BUILD_TIMESTAMP := $(SOURCE_DATE_EPOCH)
+else ifdef GITVERSION
+BUILD_TIMESTAMP := $(git log -1 --pretty=%ct)
```
mcb30: Fixed the broken use of `$(git ...)` instead of `$(shell git ...)` and pushed as https://github.com/ipxe/ipxe/commit/9d000c9fd

mcb30:
Merged (with assorted fixes to avoid build breakage) as https://github.com/ipxe/ipxe/commit/9d000c9fd, https://github.com/ipxe/ipxe/commit/bc4979e2c, and https://github.com/ipxe/ipxe/commit/d8dc06fbf

Thanks!

mmlb:
Hmm I caught the missing shell invocation too maybe I forgot to push though. 👍

mmlb:
Hey @mcb30, I noticed some follow up commits dealing with FreeBSD differences, glad those were caught quickly!

bmwiedemann:
To get an reproducible `ipxe.iso` with xorriso I had to patch syslinux isohybrid MBR ID generation https://github.com/distropatches/syslinux/tree/isohybrid and 
```diff
--- ipxe-1.20.1+git20210129.f4fcd000.orig/src/util/genfsimg
+++ ipxe-1.20.1+git20210129.f4fcd000/src/util/genfsimg
@@ -276,7 +276,7 @@ if [ -n "${ISOIMG}" ] ; then
        echo "${0}: cannot find a suitable mkisofs or equivalent" >&2
        exit 1
     fi
-    "${MKISOFS}" -quiet -volid "iPXE" -preparer "iPXE build system" \
+    "${MKISOFS}" --set_all_file_dates "2007110814511300" --modification-date=2007110814511300 -quiet -volid "iPXE" -preparer "iPXE build system" \
            -appid "iPXE - Open Source Network Boot Firmware" \
```

mcb30:
@bmwiedemann The `--set_all_file_dates` and `--modification-date` options seem not to be supported by `mkisofs` or `genisoimage`, so you'd have to test for support before using them.

Also, the patch should probably use `SOURCE_DATE_EPOCH` to construct the date, and override the default behaviour only if `SOURCE_DATE_EPOCH` is specified.

Would you be able to submit a (separate) PR that addresses those issues?

Thanks.

bmwiedemann:
There you are: #252

PR #252, opened 2021-02-10
Allow to override iso build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override iso build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

This `date` call works with GNU date and FreeBSD/MacOSX date.

### DIFF
--- a/src/util/genfsimg
+++ b/src/util/genfsimg
@@ -278,6 +278,11 @@ if [ -n "${ISOIMG}" ] ; then
 	echo "${0}: cannot find a suitable mkisofs or equivalent" >&2
 	exit 1
     fi
+    if [ "${MKISOFS}" = "xorrisofs" -a -n "${SOURCE_DATE_EPOCH:-}" ] ; then
+        DATE_FMT="+%Y%m%d%H%M%S00"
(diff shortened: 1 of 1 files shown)


mcb30:
Thanks.  I think it probably makes sense to check for `SOURCE_DATE_EPOCH` and, if set, append the arguments to `ISOARGS`.  The existing logic will then take care of choosing a mkisofs equivalent that is capable of support those arguments.  In particular, this would mean that a system with both genisoimage and xorrisofs installed would choose xorrisofs when `SOURCE_DATE_EPOCH` was specified.

Please do include the ability to use BSD date as per https://reproducible-builds.org/docs/source-date-epoch/

Also: a UEFI .iso image will include an embedded FAT filesystem: does `genfsimg` need to also do something to set the relevant timestamps within this?

bmwiedemann:
I updated the PR to incorporate the improvements. It makes `ipxe.iso` reproducible for me, so I would consider UEFI out-of-scope for now.

mcb30:
Merged with modifications as https://github.com/ipxe/ipxe/commit/0d68d7151.  For me, this still left the isohybrid MBR ID as non-deterministic: this is addressed in https://github.com/ipxe/ipxe/commit/74c54461c

PR #310, opened 2021-04-02
efi: only the cmdline is expected in the efi LoadOptions

baloo:
The image name is not expected to be placed on the LoadOptions field.
At least in the linuxx64.efi.stub, this is not expected to have any
value as the data there overrides the kernel command line set in the efi
stub.

### DIFF
--- a/src/image/efi_image.c
+++ b/src/image/efi_image.c
@@ -113,15 +113,12 @@ static wchar_t * efi_image_cmdline ( struct image *image ) {
 	wchar_t *cmdline;
 	size_t len;
 
-	len = ( strlen ( image->name ) +
-		( image->cmdline ?
-		  ( 1 /* " " */ + strlen ( image->cmdline ) ) : 0 ) );
+	len = ( image->cmdline ?
+		  ( strlen ( image->cmdline ) ) : 0 );
 	cmdline = zalloc ( ( len + 1 /* NUL */ ) * sizeof ( wchar_t ) );
 	if ( ! cmdline )
 		return NULL;
-	efi_snprintf ( cmdline, ( len + 1 /* NUL */ ), "%s%s%s",
-		       image->name,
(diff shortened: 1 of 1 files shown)


baloo:
https://github.com/systemd/systemd/blob/8ab34a49dbf75fd731973359a6f24c212682f479/src/boot/efi/stub.c#L58 The relevant piece on the other side.

I have a hard time finding a normative description of LoadOptions to choose to go either way, but:
https://uefi.org/sites/default/files/resources/UEFI_Shell_2_2.pdf Section 3.2 Invocation:
> The command-line options are separated by the space or tab character. The options are processed left-to-right retrieved from the LoadOptions field of the EFI_LOADED_IMAGE_PROTOCOL. 

I dont believe that should include the image name.

mcb30:
> > The command-line options are separated by the space or tab character. The options are processed left-to-right retrieved from the LoadOptions field of the EFI_LOADED_IMAGE_PROTOCOL.
> 
> I dont believe that should include the image name.

The UEFI spec leaves this basically undefined.  My view is that we should (and, as far as I remember, do) mimic the behaviour of the UEFI shell.

What do you see in LoadOptions when invoking a binary directly from the UEFI shell?

baloo:
The builtin EFI bootloader when booting over http does not feed the uri as LoadOptions as far as I can tell.

Reading tianocore code it seems it's also the case there:
https://github.com/tianocore/edk2/blob/0ecdcb6142037dd1cdd08660a2349960bcf0270a/OvmfPkg/Library/GenericQemuLoadImageLib/GenericQemuLoadImageLib.c#L131

mcb30:
I have (re)tested the behaviour of the UEFI shell.  When executing a binary from the UEFI shell, I do see the program name at the start of `LoadOptions`.  iPXE's behaviour is therefore consistent with the behaviour of the UEFI shell.  Please submit this instead as a bug in `shim`.

PR #312, opened 2021-04-04
intelxvf: Add additional PCI device ID to table

tj90241:
Adding this missing identifier allows the X557-AT2 chipset
seen on (at least) Super Micro A2SDI-H-TF motherboards to
function with iPXE.

Signed-off-by: Tyler J. Stachecki <stachecki.tyler@gmail.com>


mcb30:
Thanks!

PR #313, opened 2021-04-05
[efi] Improve ELF to PE conversion

mhaeuser:
As currently ELF sections rather than segments are converted to PE sections, PE section alignment is hard-coded, and some fields describing virtual properties are assigned values based on file alignment, EFI PE drivers generated by elf2efi are generally not well-aligned. This can cause issues when enforcing strict PE parsing rules (as I am currently proposing for EDK II / OVMF), or when enforcing section permissions.

This PR fixes all discovered issues regarding PE image alignment. Furthermore, the constructed debug and relocation sections are marked as discardable for security reasons. All changes made are in conformance with the PE specification.

The changes have been validated using a stricter EDK II image loader, with OVMF booting and successfully loading the appropriate EFI ROM. Images created by the old utility are rejected as expected. Please note that I have not been able to test the full functionality of the ROM image, only that it loads successfully.

### DIFF
--- a/src/scripts/efi.lds
+++ b/src/scripts/efi.lds
@@ -106,5 +106,6 @@ SECTIONS {
 	*(.einfo.*)
 	*(.discard)
 	*(.discard.*)
+	*(.pci_devlist.*)
     }
 }
--- a/src/scripts/efi.lds
+++ b/src/scripts/efi.lds
@@ -8,22 +8,22 @@
 SECTIONS {
 
     /* The file starts at a virtual address of zero, and sections are
-     * contiguous.  Each section is aligned to at least _max_align,
-     * which defaults to 32.  Load addresses are equal to virtual
+     * contiguous.  Each section is aligned to at least _page_align,
+     * which defaults to 4096.  Load addresses are equal to virtual
      * addresses.
      */
 
-    _max_align = 32;
+    _page_align = 4096;
 
-    /* Allow plenty of space for file headers */
-    . = 0x1000;
+    /* Allow one page of space for file headers, common PE/COFF layout */
+    . = _page_align;
 
     /*
      * The text section
      *
      */
 
-    . = ALIGN ( _max_align );
+    . = ALIGN ( _page_align );
     .text : {
 	_text = .;
 	*(.text)
@@ -36,7 +36,7 @@ SECTIONS {
      *
      */
 
-    . = ALIGN ( _max_align );
+    . = ALIGN ( _page_align );
     .rodata : {
 	_rodata = .;
 	*(.rodata)
@@ -49,7 +49,7 @@ SECTIONS {
      *
      */
 
-    . = ALIGN ( _max_align );
+    . = ALIGN ( _page_align );
     .data : {
 	_data = .;
 	*(.data)
@@ -65,7 +65,7 @@ SECTIONS {
      *
      */
 
-    . = ALIGN ( _max_align );
+    . = ALIGN ( _page_align );
     .bss : {
 	_bss = .;
 	*(.bss)
--- a/src/util/elf2efi.c
+++ b/src/util/elf2efi.c
@@ -758,6 +758,7 @@ create_reloc_section ( struct pe_header *pe_header,
 	reloc->hdr.VirtualAddress = pe_header->nt.OptionalHeader.SizeOfImage;
 	reloc->hdr.SizeOfRawData = section_filesz;
 	reloc->hdr.Characteristics = ( EFI_IMAGE_SCN_CNT_INITIALIZED_DATA |
+				       EFI_IMAGE_SCN_MEM_DISCARDABLE |
 				       EFI_IMAGE_SCN_MEM_NOT_PAGED |
 				       EFI_IMAGE_SCN_MEM_READ );
 
@@ -822,6 +823,7 @@ create_debug_section ( struct pe_header *pe_header, const char *filename ) {
 	debug->hdr.VirtualAddress = pe_header->nt.OptionalHeader.SizeOfImage;
 	debug->hdr.SizeOfRawData = section_filesz;
 	debug->hdr.Characteristics = ( EFI_IMAGE_SCN_CNT_INITIALIZED_DATA |
+				       EFI_IMAGE_SCN_MEM_DISCARDABLE |
 				       EFI_IMAGE_SCN_MEM_NOT_PAGED |
 				       EFI_IMAGE_SCN_MEM_READ );
 	debug->fixup = fixup_debug_section;
--- a/src/util/elf2efi.c
+++ b/src/util/elf2efi.c
@@ -125,7 +125,8 @@
 #define R_ARM_V4BX 40
 #endif
 
-#define EFI_FILE_ALIGN 0x20
+#define EFI_FILE_ALIGN  0x20
+#define EFI_IMAGE_ALIGN 0x1000
 
 struct elf_file {
 	void *data;
@@ -173,9 +174,9 @@ static struct pe_header efi_pe_header = {
 			.Magic = EFI_IMAGE_NT_OPTIONAL_HDR_MAGIC,
 			.MajorLinkerVersion = 42,
 			.MinorLinkerVersion = 42,
-			.SectionAlignment = EFI_FILE_ALIGN,
+			.SectionAlignment = EFI_IMAGE_ALIGN,
 			.FileAlignment = EFI_FILE_ALIGN,
-			.SizeOfImage = sizeof ( efi_pe_header ),
+			.SizeOfImage = EFI_IMAGE_ALIGN,
 			.SizeOfHeaders = sizeof ( efi_pe_header ),
 			.NumberOfRvaAndSizes =
 				EFI_IMAGE_NUMBER_OF_DIRECTORY_ENTRIES,
@@ -214,6 +215,16 @@ static void * xmalloc ( size_t len ) {
  */
 static unsigned long efi_file_align ( unsigned long offset ) {
 	return ( ( offset + EFI_FILE_ALIGN - 1 ) & ~( EFI_FILE_ALIGN - 1 ) );
+}
+
+/**
+ * Align section within PE image
+ *
+ * @v offset		Unaligned offset
+ * @ret aligned_offset	Aligned offset
+ */
+static unsigned long efi_image_align ( unsigned long offset ) {
+	return ( ( offset + EFI_IMAGE_ALIGN - 1 ) & ~( EFI_IMAGE_ALIGN - 1 ) );
 }
 
 /**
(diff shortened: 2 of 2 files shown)


mhaeuser:
Actually this may be incomplete. When trying to boot shimx64.efi, I realised the binary's relocations are malformed, i.e. it is not ensured that relocation blocks have a size aligned by 4 Bytes as per the PE spec. While I did not have any error with my generated PE iPXE binary, I do not quite see (yet?) where in the code this condition is ensured. Should it not be around [here](https://github.com/ipxe/ipxe/tree/1192edf394aaebde666c6aafda1c8b782b08a845/srcutil/elf2efi.c#L300)? If the logic is indeed missing, I can probably help writing it, but testing will be hard with no malformed binary discovered so far.

Thank you.

mcb30:
Thank you.  Some issues observed from a quick initial check of the patch:

* Please reformat code to remain within an 80-character width

* Please reflow commit messages to remain within an 80-character width

* Running `objdump -x bin-x86_64-efi/ipxe.efi` no longer shows meaningful information about the PE sections.

  Before:
  ```
  Sections:
  Idx Name          Size      VMA               LMA               File off  Algn
    0 .text         000a0148  0000000000001000  0000000000001000  000002c0  2**4
                    CONTENTS, ALLOC, LOAD, READONLY, CODE
    1 .rodata       0002c10e  00000000000a1160  00000000000a1160  000a0420  2**2
                    CONTENTS, ALLOC, LOAD, READONLY, DATA
    2 .data         00023f40  00000000000cd280  00000000000cd280  000cc540  2**4
                    CONTENTS, ALLOC, LOAD, DATA
    3 .bss          000a3e0c  00000000000f1200  00000000000f1200  00000000  2**4
                    ALLOC
    4 .reloc        00001a9c  0000000000195020  0000000000195020  000f0480  2**2
                    CONTENTS, ALLOC, LOAD, READONLY, DATA
    5 .debug        00000040  0000000000196ac0  0000000000196ac0  000f1f20  2**0
                    CONTENTS, READONLY, DEBUGGING
  ```
  After:
  ```
  Sections:
  Idx Name          Size      VMA               LMA               File off  Algn
    0               000f01a0  0000000000001000  0000000000001000  00000240  2**2
                    CONTENTS, ALLOC, LOAD, CODE
    1 .reloc        00001ac0  0000000000196000  0000000000196000  000f03e0  2**2
                    CONTENTS, ALLOC, LOAD, READONLY, DATA
    2 .debug        00000040  0000000000198000  0000000000198000  000f1ea0  2**0
                    CONTENTS, READONLY, DEBUGGING
  ```

mcb30:
> Actually this may be incomplete. When trying to boot shimx64.efi, I realised the binary's relocations are malformed, i.e. it is not ensured that relocation blocks have a size aligned by 4 Bytes as per the PE spec. While I did not have any error with my generated PE iPXE binary, I do not quite see (yet?) where in the code this condition is ensured. Should it not be around [here](https://github.com/ipxe/ipxe/tree/1192edf394aaebde666c6aafda1c8b782b08a845/src/util/elf2efi.c#L300)? If the logic is indeed missing, I can probably help writing it, but testing will be hard with no malformed binary discovered so far.

The existing code will pad the section filesize to 32-byte alignment and zero-fill any unused space.

PE relocation records are two bytes each, so if the PE spec requires 4-byte alignment of the size of relocation records then that seems to be a bug in the PE spec.  Do you have a direct link to the relevant requirement?

mcb30:
From a closer look at the code, it looks as though we already ensure that the relocation section size is 4-byte aligned:
```
num_relocs = ( ( pe_rel->used_relocs + 1 ) & ~1 );
```

mhaeuser:
> Thank you. Some issues observed from a quick initial check of the patch:
> 
> * Please reformat code to remain within an 80-character width
> * Please reflow commit messages to remain within an 80-character width

OK, sorry.

> * Running `objdump -x bin-x86_64-efi/ipxe.efi` no longer shows meaningful information about the PE sections.
>   Before:
>   ```
>   Sections:
>   Idx Name          Size      VMA               LMA               File off  Algn
>     0 .text         000a0148  0000000000001000  0000000000001000  000002c0  2**4
>                     CONTENTS, ALLOC, LOAD, READONLY, CODE
>     1 .rodata       0002c10e  00000000000a1160  00000000000a1160  000a0420  2**2
>                     CONTENTS, ALLOC, LOAD, READONLY, DATA
>     2 .data         00023f40  00000000000cd280  00000000000cd280  000cc540  2**4
>                     CONTENTS, ALLOC, LOAD, DATA
>     3 .bss          000a3e0c  00000000000f1200  00000000000f1200  00000000  2**4
>                     ALLOC
>     4 .reloc        00001a9c  0000000000195020  0000000000195020  000f0480  2**2
>                     CONTENTS, ALLOC, LOAD, READONLY, DATA
>     5 .debug        00000040  0000000000196ac0  0000000000196ac0  000f1f20  2**0
>                     CONTENTS, READONLY, DEBUGGING
>   ```
>   
>   
>   After:
>   ```
>   Sections:
>   Idx Name          Size      VMA               LMA               File off  Algn
>     0               000f01a0  0000000000001000  0000000000001000  00000240  2**2
>                     CONTENTS, ALLOC, LOAD, CODE
>     1 .reloc        00001ac0  0000000000196000  0000000000196000  000f03e0  2**2
>                     CONTENTS, ALLOC, LOAD, READONLY, DATA
>     2 .debug        00000040  0000000000198000  0000000000198000  000f1ea0  2**0
>                     CONTENTS, READONLY, DEBUGGING
>   ```

ELF segments do not have names, so it's sort of expected. I suppose we could have heuristics for ".text", ".data", and other common ones? I dislike heuristics and lack of information equally, so I am fine with whatever you prefer.

I also just realised that this change alone will not be acceptable as the source ELF file has one rwx segment and applies more granular permissions per section. For this change to be extensive, this needs to be fixed too somehow. I will change the state to Draft and hope you have a good suggestion at hand, otherwise I will have to go back to the ELF docs. :)

> The existing code will pad the section filesize to 32-byte alignment and zero-fill any unused space.
> 
> PE relocation records are two bytes each, so if the PE spec requires 4-byte alignment of the size of relocation records then that seems to be a bug in the PE spec. Do you have a direct link to the relevant requirement?

This one is funny because it's **heavily** implied and basically the only sensical way I see to implement it, but not strictly spelled out. The two relevant pieces are these:
- "Each block must start on a 32-bit boundary." [1]
- "IMAGE_REL_BASED_ABSOLUTE": "The base relocation is skipped. This type can be used to pad a block." [2]

While the records themselves are 2 bytes each as you say and thus only require a 2 byte alignment, the block headers have 2 records of 4 bytes each and hence require a 4 byte alignment. This can be accomplished by inserting an absolute relocation to reach an even amount of entries in a block.

The spec is very vague on this, but let's be honest, the spec is very vague on everything. I do not think there is an alternative interpretation that makes sense in context of other, similar examples.

[1] PE format, "The .reloc Section (Image Only)" (https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#the-reloc-section-image-only)
[2] PE format, "The .reloc Section (Image Only)", "Base Relocation Types" (https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#base-relocation-block)

mcb30:
> ELF segments do not have names, so it's sort of expected. I suppose we could have heuristics for ".text", ".data", and other common ones? I dislike heuristics and lack of information equally, so I am fine with whatever you prefer.

I would like the existing section names (as defined deliberately in `efi/script.lds`) to be preserved.  If a future patch modifies `efi/script.lds` to add a new section, then that section should appear automatically in the final EFI binary with no modification to `elf2efi.c` required.

This avoids any kind of heuristics, and matches the existing behaviour.

As far as I'm concerned, the linker scripts are **definitive** to describe the intended layouts of the final binaries across all platforms (not just UEFI).  The function of `elf2efi.c` should be to merely provide the mechanism that transcribes the policy defined in `scripts/efi.lds` from ELF format into PE format.

> I also just realised that this change alone will not be acceptable as the source ELF file has one rwx segment and applies more granular permissions per section. For this change to be extensive, this needs to be fixed too somehow. I will change the state to Draft and hope you have a good suggestion at hand, otherwise I will have to go back to the ELF docs. :)

I would probably suggest retaining the existing behaviour of copying sections rather than segments, since that seems to already more closely match what you are aiming to achieve.  Would it suffice to simply change the section alignment (`_max_align`) to 4kB rather than 32 bytes?

mhaeuser:
> I would probably suggest retaining the existing behaviour of copying sections rather than segments, since that seems to already more closely match what you are aiming to achieve. Would it suffice to simply change the section alignment (`_max_align`) to 4kB rather than 32 bytes?

That would be fine by me, just I researched the ELF format a little and like the Mach-O format, a loader should not be concerned with sections (linker info) but segments. Another concern is that the PE sections must be adjacent. If both conditions are fulfilled however, I see no reason to not take your route.

I will do the edits you suggested locally and verify with fully constrained loading. If it works, I will submit a new PR. Sorry, I'm just not familiar with ELF or GNU utilities. :)

mhaeuser:
@mcb30 Sorry to bother again, but I faced an unaligned ".pci_dev" section in the resulting PE. It is neither declared to discarded in the LDS. Is it needed?

mcb30:
> @mcb30 Sorry to bother again, but I faced an unaligned ".pci_dev" section in the resulting PE. It is neither declared nor
> discarded in the LDS. Is it needed?

Good catch!  That must be `.pci_devlist` truncated to the 8 characters available for PE section names.  It's part of the mechanism used for BIOS PCI ROMs to build up a list of other PCI IDs supported by the same ROM image (and linked to as part of the PCI ROM header).

It's irrelevant for UEFI, where we don't have any such capability to mark a ROM as supporting multple PCI IDs.  Doesn't show up in `bin-x86_64-efi/ipxe.efi` since "ipxe" is not a PCI vendor:device ID, but it does currently (and erroneously) show up as a section in e.g. `bin-x86_64-efi/8086100e.efi`.

It should be safe to add it to the discard list.  If that doesn't work (e.g. because the link-time check complains about an explicitly requested symbol being discarded), then just add it in to `.rodata`.

mhaeuser:
@mcb30 Sorry, misclicked. Confirmed working, will submit a V2. Tyvm!

mcb30 (commented):
Looks much better: thank you for updating it!  One minor whitespace fix required (see inline), plus some tweaks to the commit messages:

* "Discard .pci_devlist_* sections" should read "Discard .pci_devlist.* sections"
* The second commit has a stray "af843d" after your Signed-off-by line
* If you have a real email address (rather than `@users.noreply.github.com`) then that would be preferable, but not essential
* I realise that my "80 character width" request was not specific enough, sorry.  The code is fine, but the commit messages need to wrap such that the output of `git log` will look correct in an 80-column terminal.  Since `git log` adds 4 extra spaces, this requires wrapping the original message at fewer than 80 columns.  I'd suggest wrapping at 70 (which is what the One True Editor will do by default :slightly_smiling_face: )

I could fix these up myself if you prefer, but then you wouldn't get the satisfaction of a "Merged" status in the GitHub UI.

mhaeuser:
@mcb30 Sorry for the hassle, I believe I corrected everything. Thanks for the fast review!

mcb30:
@mhaeuser Only the author email still needed fixing, so I just did that one myself.  Thanks for the improvements!

PR #328, opened 2021-04-20
[drivers/net] Add Marvell AQtion network controller driver

pcfist:
Add driver for Marvell AQtion/FastLinQ Edge family multi-gigabit network controllers (https://www.marvell.com/products/ethernet-adapters-and-controllers/fastlinq-edge-ethernet-controllers.html). Currently supports AQC107/AQC108/AQC109 based devices.


src/include/ipxe/errfile.h
```diff
@@ -213,6 +213,8 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define ERRFILE_usbblk		     ( ERRFILE_DRIVER | 0x00ce0000 )
 #define ERRFILE_iphone		     ( ERRFILE_DRIVER | 0x00cf0000 )
 #define ERRFILE_slirp		     ( ERRFILE_DRIVER | 0x00d00000 )
+#define ERRFILE_aqc1xx		     ( ERRFILE_DRIVER | 0x00d00000 )
```
NiKiZe, line 216: Already used by slirp?

pcfist: Oops. Sorry, did not notice this conflict after rebase. Thanks!

pcfist: Fixed 👌

src/drivers/net/marvell/aqc1xx.h
```diff
@@ -0,0 +1,253 @@
...
+			uint32_t dx_type:3;
+			uint32_t rsvd1:1;
+			uint32_t buf_len:16;
+			uint32_t dd:1;
+			uint32_t eop:1;
+			uint32_t cmd:8;
+			uint32_t rsvd2:2;
+			uint32_t rsvd3:14;
+			uint32_t pay_len:18;
+		};
```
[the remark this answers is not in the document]
mcb30: This comment applies to all uses of bitfields within this driver.

src/drivers/net/marvell/atl_hw.c
```diff
@@ -0,0 +1,317 @@
...
+{
+	uint32_t val;
+	int k = 0;
+
+	ATL_WRITE_REG(0x40e1, ATL_GLB_CTRL2);
+	mdelay(50);
+
+	/* Cleanup SPI */
+	val = ATL_READ_REG(ATL_GLB_NVR_PROV4);
+	ATL_WRITE_REG(val | 0x10, ATL_GLB_NVR_PROV4);
```
mcb30, line 54: Please define constants instead of using magic values.

src/drivers/net/marvell/atl_hw.c
```diff
@@ -0,0 +1,317 @@
...
+
+	ATL_WRITE_REG((ATL_READ_REG(ATL_GLB_STD_CTRL) &
+		      ~ATL_GLB_CTRL_RST_DIS) | 0x8000, ATL_GLB_STD_CTRL);
```
mcb30, line 57: Please define constants instead of using magic values.

(Assume this comment applies everywhere else within this driver)

src/drivers/net/marvell/atl_hw.c
```diff
@@ -0,0 +1,317 @@
...
+			goto err;
+
+		buffer[i] = ATL_READ_REG(ATL_MBOX_CTRL5);
+	}
+
+	ATL_WRITE_REG(1, ATL_SEM_RAM);
+
+	return 0;
+err:
+	DBG("AQUANTIA: download_dwords error\n");
```
mcb30, line 287: Please include meaningful error messages.  You have what appears to be two unrelated possible causes for this error: they should not produce the same message.

This comment applies throughout the driver.

src/drivers/net/marvell/atl_hw.c
```diff
@@ -0,0 +1,317 @@
...
+	uint32_t efuse_addr = ATL_READ_REG(ATL_GLB_MCP_SP26);
+
+	if (efuse_addr != 0) {
+		uint32_t mac_efuse_addr = efuse_addr + 40 * sizeof(uint32_t);
+		err = atl_hw_read_mem(nic, mac_efuse_addr, mac_addr, 2);
+		if (err != 0)
+			return err;
+
+		mac_addr[0] = __bswap_32(mac_addr[0]);
+		mac_addr[1] = __bswap_32(mac_addr[1]);
```
mcb30, line 304: Will this always require a swap, or should this be using `cpu_to_be32` instead?

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	memset(ring->ring, 0, ring->length);
+
+	/* Write ring address (hi & low parts).*/
+	phy_addr = virt_to_bus(ring->ring);
+	ATL_WRITE_REG((uint32_t)phy_addr, reg_base + 0);
+	ATL_WRITE_REG((uint32_t)(((uint64_t)phy_addr) >> 32), reg_base + 4);
+	/* Write ring length.*/
+	ATL_WRITE_REG(ATL_RING_SIZE, reg_base + 8);
+
+	/* #todo: reset head and tail pointers */
```
mcb30, line 72: Please address this `todo`

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	DBG("ATLANTIC %p ring is at [%08llx,%08llx), reg base %#x\n",
+	    nic, ((unsigned long long)phy_addr),
+	    ((unsigned long long) phy_addr + ring->length), reg_base);
```
mcb30, line 75: Please use the `DBGC` form of debug macro, which applies automatic colourisation and therefore makes it much easier to follow multiple streams of debug output.  The first (colour) argument may be pretty much any value that identifies a debug stream.  Drivers tend to use the pointer that represents the network device: in your case this would therefore be: `DBGC ( nic, "ATLANTIC %p ..." )`

This comment applies throughout the driver.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+
+
+static int atl_ring_alloc(const struct atl_nic *nic, struct atl_ring *ring,
+			  uint32_t desc_size, uint32_t reg_base)
+{
+	physaddr_t phy_addr;
+
+	/* Allocate ring buffer.*/
+	ring->length = ATL_RING_SIZE * desc_size;
+	ring->ring = malloc_phys(ring->length, ATL_RING_ALIGN);
```
mcb30, line 58: Please use `dma_malloc()` rather than `malloc_phys`, so that your driver will work on platforms that have an IOMMU enabled.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+
+	/* #todo: reset head and tail pointers */
+	DBG("ATLANTIC %p ring is at [%08llx,%08llx), reg base %#x\n",
+	    nic, ((unsigned long long)phy_addr),
+	    ((unsigned long long) phy_addr + ring->length), reg_base);
+
+	return 0;
+
+err_alloc:
+	if (ring->ring)
```
mcb30, line 80: In the event of an allocation error, `ring->ring` is guaranteed to be NULL.  This error handling path should look like either:
```
        int rc;
        .........
        if ( ! ring->ring ) {
                rc = -ENOMEM;
                goto err_alloc;
        }
        .........
        return 0;

        dma_free ( ... );
err_alloc:
        return rc;
```

or, if there is no other possible error and no other cleanup that needs to be done, then just:
```
        if ( ! ring->ring )
                return -ENOMEM;
```

This comment applies throughout the driver.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+
+	ring->ring = NULL;
+	ring->length = 0;
+
+	return -ENOMEM;
+}
+
+static void atl_ring_free(struct atl_ring *ring)
+{
+	if (ring->ring)
```
mcb30, line 91: Are there code paths that may call `atl_ring_free()` when the ring has not, in fact, been allocated?  If not, then this check is entirely redundant (and should be removed).  If so, then there is potentially a bigger problem.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	ring->ring = NULL;
+	ring->length = 0;
+}
+
+static void atl_ring_next_dx(unsigned int *val)
+{
+	++(*val);
+	if (*val == ATL_RING_SIZE)
+		*val = 0;
+}
```
mcb30, line 102: This is unnecessary complexity: please just mask the values against `ATL_RING_SIZE - 1` at the point of use instead, and remove this function.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+			/* Wait for next refill */
+			break;
+		}
+
+		/* Get next receive descriptor */
+		rx = (struct atl_desc_rx *)nic->rx_ring.ring +
+		nic->rx_ring.sw_tail;
+
+		/* Populate receive descriptor */
+		address = virt_to_bus(iobuf->data);
```
mcb30, line 133: Please use `alloc_rx_iob()` instead so that your driver will work on systems with an IOMMU enabled.  You can use `drivers/net/intel.c` as a reference for how to do this.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+static int atl_open(struct net_device *netdev)
+{
+	struct atl_nic *nic = netdev->priv;
+	uint32_t ctrl = 0;
+	DBG("AQUANTIA: atl_open()\n");
+
+	/* Tx ring */
+	if (atl_ring_alloc(nic, &nic->tx_ring, sizeof(struct atl_desc_tx),
+	    ATL_TX_DMA_DESC_ADDR) != 0)
+		goto err_alloc;
```
mcb30, line 170: Please use separate error handling labels for TX and RX allocation failures.  See `intel_open()` for an example.  This will allow you to eliminate all of the checks for `ring->ring == NULL`.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	/*Enable rings*/
+	ATL_WRITE_REG(ATL_READ_REG(ATL_RING_TX_CTRL) | ATL_RING_TX_CTRL_EN,
+		      ATL_RING_TX_CTRL);
+	ATL_WRITE_REG(ATL_READ_REG(ATL_RING_RX_CTRL) | ATL_RING_RX_CTRL_EN,
+		      ATL_RING_RX_CTRL);
+
+	atl_rx_ring_fill(nic);
+
+	nic->hw_ops->start(nic);
+	DBG("AQUANTIA: code 0()\n");
```
mcb30, line 246: This debug message appears to be meaningless since it provides no information.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	struct atl_desc_tx *tx;
+	physaddr_t address;
+	size_t len;
+
+	/* Get next transmit descriptor */
+	if (atl_ring_full(&nic->tx_ring)) {
+		DBG("AQUANTIA: %p out of transmit descriptors\n", nic);
+		return -ENOBUFS;
+	}
+	tx = (struct atl_desc_tx *)nic->tx_ring.ring + nic->tx_ring.sw_tail;
```
mcb30, line 306: This pointer cast should not be necessary: please eliminate it.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+
+	/* Populate transmit descriptor */
+	address = virt_to_bus(iobuf->data);
```
mcb30, line 309: Please use `iob_dma()` so that your driver will work on systems with an IOMMU enabled.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	/* if (icr & ATL_IRQ_TX) */
+	atl_poll_tx(netdev);
+
+	/* Poll for RX completions, if applicable */
+	/* if (icr & ATL_IRQ_RX) */
+	atl_poll_rx(netdev);
+
+	/* Refill RX ring */
+	atl_rx_ring_fill(nic);
+}
```
mcb30, line 441: Does the hardware not provide some kind of interrupt status register that can be used to avoid unnecessarily performing all three actions (check link, poll TX, poll RX) on every poll?

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+			return;
+
+		/* Populate I/O buffer */
+		iobuf = nic->iobufs[nic->rx_ring.sw_head];
+		nic->iobufs[nic->rx_ring.sw_head] = NULL;
+		len = le16_to_cpu(rx->pkt_len);
+		iob_put(iobuf, len);
+
+		/* Hand off to network stack */
+		/*to do: process error*/
```
mcb30, line 409: Please address this todo

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	if (!netdev) {
+		rc = -ENOMEM;
+		goto err_alloc;
+	}
+	netdev_init(netdev, &atl_operations);
+	nic = netdev->priv;
+	pci_set_drvdata(pci, netdev);
+	netdev->dev = &pci->dev;
+	memset(nic, 0, sizeof(*nic));
+	nic->port = PCI_FUNC(pci->busdevfn);
```
mcb30, line 505: How will this behave when using PCI pass-through inside a VM?  For passed-through devices, the virtual PCI function number is likely to be zero regardless of the real function number.  See e.g. commit https://github.com/ipxe/ipxe/commit/76a7bfe93

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	nic->flags = pci->id->driver_data;
+
+	/* Fix up PCI device */
+	adjust_pci_device(pci);
+
+	/* Map registers */
+	nic->regs = ioremap(pci->membase, ATL_BAR_SIZE);
```
mcb30, line 512: Please use `pci_ioremap`.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+	}
+
+	switch (nic->flags) {
+	case ATL_FLAG_A1:
+		nic->hw_ops = &atl_hw;
+		break;
+	default:
+		goto err_unsupported;
+		break;
+	}
```
mcb30, line 525: This appears to be entirely redundant: every supported device provides `ATL_FLAG_A1` and the driver cannot support any other kind of device.  Please remove `nic->flags` and the associated logic.

src/drivers/net/marvell/aqc1xx.c
```diff
@@ -0,0 +1,634 @@
...
+
+	/* Register network device */
+	if (register_netdev(netdev) != 0)
+		goto err_register_netdev;
+
+	/* Set initial link state */
+	netdev_link_down(netdev);
+
+	DBG("AQUANTIA: atl_probe code 0\n");
+	return 0;
```
mcb30, line 543: Again, this debug message seems to be redundant.

mcb30 (changes requested):
Thanks for the submission!  The code generally looks like a good first draft for an iPXE driver.  There are many review comments that need to be addressed, several of which apply throughout the whole driver (e.g. using named constants rather than magic numbers).  I'll review again once these have been addressed.

Also, please rebase and squash into a form where every commit builds and works.  (The easiest way to do this is probably to rebase and squash into a single commit that just adds the finished form of the driver.)

Thank you!

pcfist:
> Thanks for the submission! The code generally looks like a good first draft for an iPXE driver. There are many review comments that need to be addressed, several of which apply throughout the whole driver (e.g. using named constants rather than magic numbers). I'll review again once these have been addressed.
> 
> Also, please rebase and squash into a form where every commit builds and works. (The easiest way to do this is probably to rebase and squash into a single commit that just adds the finished form of the driver.)
> 
> Thank you!

Hello Michael,

Thank you for comprehensive review! I will go over all your commentaries, and then squash the changes into one or multiple commits.

src/drivers/net/marvell/aqc1xx.h
```diff
@@ -0,0 +1,253 @@
...
+
+struct atl_nic;
+
+struct atl_hw_ops {
+	int (*reset) (struct atl_nic *nic);
+	int (*start) (struct atl_nic *nic);
+	int (*stop) (struct atl_nic *nic);
+	int (*get_link) (struct atl_nic *nic);
+	int (*get_mac) (struct atl_nic *, uint8_t *mac);
+};
```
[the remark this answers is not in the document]
pcfist: We are planning to support another device, AQC113, which is a successor to AQC107. This is why this abstraction layer exists. Actually, some work has already been done to implement it, but we decided to go step by step.

Do you want me to remove it completely from the initial submission and then add it back, or is it OK to keep it?

PR #331, opened 2021-04-21
[bnxt] Use hexadecimal values in PCI_ROM entries

jw14812:
Use hexadecimal values instead of macros in PCI_ROM entries so Perl script
can parse them correctly. Move PCI_ROM entries from header file to C file.


NiKiZe:
This introduces a disconnect between some defines and `PCI_ROM` lines (`vf_nics` referring to same device as in `PCI_ROM` line)
While some e.g. `DID_57417_RDMA_MF` becomes unused?
If I'm not mistaken some entries are also added?

Maybe `bnxt_vf_nics` could be removed by using a data flags instead? (fff9281b8454dadf5e31c297af904959fe35800b)
If this is modified maybe the whitespace should be added in the `PCI_ROM ( 0x, ..., 0 ),` to match style used elsewhere?

PR #336, opened 2021-04-23
[bnxt] Use hexadecimal values in PCI_ROM entries

jw14812:
Use hexadecimal values instead of macros in PCI_ROM entries so Perl script
can parse them correctly. Move PCI_ROM entries from header file to C file.
Integrate bnxt_vf_nics array into PCI_ROM entries by introducing
BNXT_FLAG_PCI_VF flag into driver_data field. Add whitespaces in PCI_ROM
entries for style consistency.


src/drivers/net/bnxt/bnxt.h
```diff
@@ -885,15 +886,19 @@ struct bnxt {
...
+#define DID_57454_MHB_RDMA   0x1605
+#define DID_57454_VF_RDMA    0x1606
+#define DID_57454_VF         0x1609
+#define DID_57454            0x1614
+#define DID_58802            0xD802
+#define DID_58804            0xD804
+#define DID_57454_VF_HV      0x1607
+#define DID_57454_VF_HV_RDMA 0x1608
+#define DID_57412_VF_HV      0x16c5
+#define DID_57412_VF_HV_RDMA 0x16bd
```
NiKiZe, line 902: With the new flag, these defines are unused, correct?

NiKiZe (commented):
Continuation from #331, This looks cleaner

Any reason why the `/* Stratus Device IDs */` defines are added/keept?

src/drivers/net/bnxt/bnxt.h
jw14812: Correct observation.  I have removed the unnecessary defines as suggested

NiKiZe (approved):

NiKiZe:
@mcb30 I think that this is a good step in the right direction.

mcb30 (approved):

PR #350, opened 2021-05-01
[multiboot] try other regions for large modules

jclulow:
Multiboot modules are placed immediately after the primary executable
image.  If the memory region where the executable is placed is not very
large, modules may not fit in that region.  Try skipping over holes to
see if the module will fit in a subsequent region.

### DIFF
--- a/src/arch/x86/image/multiboot.c
+++ b/src/arch/x86/image/multiboot.c
@@ -178,6 +178,36 @@ static physaddr_t multiboot_add_cmdline ( struct image *image ) {
 	return virt_to_phys ( mb_cmdline );
 }
 
+static int multiboot_find_new_start ( struct image *image, physaddr_t *start ) {
+	struct memory_map memmap;
+	unsigned int i;
+	unsigned int current;
+
+	get_memmap ( &memmap );
+
+	/* Find the current region: */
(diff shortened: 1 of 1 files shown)


jclulow:
I hit this when trying to load a ~450MB RAM disk into a system with these memory regions:

```
               0 -            96c00 sz            603 K type 1
          100000 -          4000000 sz             63 M type 1
         4048000 -         75db0000 sz              1 G type 1
        76000000 -         a6efc000 sz            782 M type 1
        aa5ff000 -         abe87000 sz             24 M type 1
--- 64 bit boundary ---
       100000000 -       104ef00000 sz             61 G type 1
      1050000000 -       204fc00000 sz             63 G type 1
```

This was using a BIOS boot ISO (`make bin/ipxe.iso`).  From the debug output it seemed that the HTTP download method was getting memory (via `memtop_urealloc()`) from the top of the region starting at `0x4048000`.  Ideally I'd like to be able to skip the malloc region, probably, to avoid conflicts -- but I couldn't immediately see an easy way to tell which region that is.

jclulow:
Hi!  Wondering if you had any thoughts on this -- or is there anything I should be doing to get review or to move this along?  Thanks.

dancrossnyc:
Hey! This would be SUPER useful; would there any chance to get this merged?

PR #393, opened 2021-06-18
[nbd] Add support for the NBD (Network Block Device) protocol

pe-bo:
The NBD is a lightweight block access protocol (see https://nbd.sourceforge.io) that allows access to a block device over TCP and can be exposed as an iPXE SAN device.

NBD uses the URI `format nbd://<host>[:port]/<export>`

The driver was tested using UEFI on a Dell 3050 (NIC Realtek RTL 8111) and a Linux nbd 3.21 server. It is able to run Windows PE iso (~ 300 MB) in about 30 seconds.


NiKiZe:
We have wimboot for Windows, and also iSCSI. What is the benefitt of NBD, and is there any specific use case?

src/config/defaults/pcbios.h
```diff
@@ -44,6 +44,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define	SANBOOT_PROTO_IB_SRP	/* Infiniband SCSI RDMA protocol */
 #define	SANBOOT_PROTO_FCP	/* Fibre Channel protocol */
 #define SANBOOT_PROTO_HTTP	/* HTTP SAN protocol */
+#define SANBOOT_PROTO_NBD	/* Network Block Device protocol */
```
mcb30, line 47: This should definitely not be enabled by default for BIOS builds, since they are size sensitive and NBD is a relatively obscure protocol.  (The other obscure protocol, IB_SRP, depends on the presence of an Infiniband driver and so will not be included in an Ethernet-only build)

src/config/defaults/efi.h
```diff
@@ -37,6 +37,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define	SANBOOT_PROTO_IB_SRP	/* Infiniband SCSI RDMA protocol */
 #define	SANBOOT_PROTO_FCP	/* Fibre Channel protocol */
 #define	SANBOOT_PROTO_HTTP	/* HTTP SAN protocol */
+#define SANBOOT_PROTO_NBD	/* Network Block Device protocol */
```
mcb30, line 40: This should not be enabled by default for UEFI, unless there is external funding available for a Secure Boot review of the code.

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ */
+
+/**
+ * Check NBD flow-control window
+ *
+ * @v nbd		NBD session
+ * @ret len		Length of window
+ */
+static size_t nbd_block_window ( struct nbd_session *nbd ) {
+	DBGC ( nbd, "NBD %p block window\n", nbd );
```
mcb30, line 233: This is likely to be a very high-volume debug message: please reduce to an appropriate level (or remove the message completely)

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+
+	if ( nbd->rx_state >= NBD_RX_TRANS_REP_CMD &&
+	     nbd->command == NULL )
+		/* We cannot handle concurrent commands */
```
mcb30, line 237: Please use braces for a multi-line `if` statement such as this

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ * Issue NBD block read
+ *
+ * @v nbd		NBD session
+ * @v block		Block data interface
+ * @v lba		Starting logical block address
+ * @v count		Number of blocks to transfer
+ * @v buffer		Data buffer
+ * @v len		Length of data buffer
+ * @ret rc		Return status code
+
```
mcb30, line 253: Please remove blank lines in the middle of comment blocks

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+			    userptr_t buffer, size_t len ) {
+	int rc;
+
+	DBGCP ( nbd, "NBD %p block %p read LBA 0x%08llx count 0x%04x\n",
+		nbd, block, lba, count );
+
+	rc = nbd_command ( nbd, block, NBD_BLOCK_READ,
+			   lba, count, buffer, len );
+	if ( rc != 0 )
+		return rc;
```
mcb30, line 267: Please use the more iPXE-idiomatic:
```
if ( ( rc = nbd_command (......) ) != 0 )
    return rc;
```
here and at all other locations within this code

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ * @ret rc		Return status code
+ */
+static int nbd_block_write ( struct nbd_session *nbd,
+			     struct interface *block,
+			     uint64_t lba, unsigned int count,
+			     userptr_t buffer, size_t len ) {
+	int rc;
+
+	DBGCP ( nbd, "NBD %p block %p write LBA 0x%08llx count 0x%04x\n",
+		nbd, block, lba, count );
```
mcb30, line 292: This should probably be `DBGC2()`

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ *
+ * @v nbd		NBD session
+ * @v block		Block data interface
+ * @ret rc		Return status code
+ */
+static int nbd_block_read_capacity ( struct nbd_session *nbd,
+				     struct interface *block ) {
+	int rc;
+
+	DBGCP ( nbd, "NBD %p block read capacity %p\n", nbd, block );
```
mcb30, line 320: As above

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+}
+
+/**
+ * Describe as an EFI device path
+ *
+ * @v nbd		NBD session
+ * @ret path		EFI device path, or NULL on error
+ */
+static EFI_DEVICE_PATH_PROTOCOL * nbd_efi_describe ( struct nbd_session *nbd ) {
+	DBGCP ( nbd, "NBD %p EFI describe\n", nbd );
```
mcb30, line 339: As above

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+	INTF_OP ( intf_close, struct nbd_session *, nbd_close ),
+	EFI_INTF_OP ( efi_describe, struct nbd_session *, nbd_efi_describe ),
+};
+
+/** NBD block interface descriptor */
+static struct interface_descriptor nbd_block_desc =
+	INTF_DESC ( struct nbd_session, block, nbd_block_op );
+
+
+static inline void nbd_neg_done ( struct nbd_session *nbd ) {
```
mcb30, line 360: Please give this function a comment block to explain what it is for

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+		union {
+			struct nbd_proto_opt_export_name opt_export_name;
+			struct nbd_proto_opt_go opt_go;
+		};
+	} __attribute__ (( packed )) *opt;
+	size_t name_len, opt_len;
+	int rc;
+
+	name_len = strlen ( nbd->export_name );
+	if ( nbd->use_opt_go )
```
mcb30, line 432: As above (multi-line `if`)

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+		opt->opt_go.name_length = htonl ( name_len );
+		strcpy ( opt->opt_go.export_name, nbd->export_name );
+	} else {
+		/* Construct NBD_OPT_EXPORT_NAME option */
+		opt->opt_export_name.request.request_magic =
+			htonll ( NBD_OPT_REQ_MAGIC );
+		opt->opt_export_name.request.option =
+			htonl ( NBD_OPT_EXPORT_NAME );
+		opt->opt_export_name.request.length = htonl ( name_len );
+		strcpy ( opt->opt_export_name.export_name, nbd->export_name );
```
mcb30, line 470: Is there a guarantee that this copy cannot overrun?

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+	struct nbd_command *nbdcmd = nbd->command;
+	struct block_device_capacity capacity;
+
+	/* Sanity check */
+	assert ( nbdcmd != NULL && nbdcmd->type == NBD_BLOCK_READ_CAPACITY );
+
+	capacity.blocks = nbd->export_size / NBD_BLOCK_SIZE;
+	capacity.blksize = NBD_BLOCK_SIZE;
+	/* Use reasonable amount of data */
+	capacity.max_count = 16 * 1024 / NBD_BLOCK_SIZE;
```
mcb30, line 505: Set to `-1U` (as done in `scsi.c`) if there is no fixed limit.

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+	struct io_buffer *iobuf;
+	int rc;
+
+	/* Sanity check */
+	assert ( nbdcmd != NULL && nbdcmd->type == NBD_BLOCK_WRITE );
+
+	len = nbdcmd->data_len - nbdcmd->data_offset;
+	/* Always send 512-byte data chunks */
+	if ( len > 512 )
+		len = 512;
```
mcb30, line 573: Use a named constant.

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ * Process initial negotiation
+ *
+ * @v nbd		NBD session
+ * @v data		Received data
+ * @v len		Length of received data
+ * @ret rc		Return status code
+ */
+static int nbd_rx_neg_init ( struct nbd_session *nbd ) {
+	DBGC2 ( nbd, "NBD %p initial handshake:\n", nbd );
+        DBGC2_HD ( nbd, &nbd->rx_neg_init, sizeof ( nbd->rx_neg_init ) );
```
mcb30, line 673: Something odd has happened here with whitespace

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+	DBGC2 ( nbd, "NBD %p export name reply:\n", nbd );
+        DBGC2_HD ( nbd, &nbd->rx_exp_name_reply,
+		   sizeof ( nbd->rx_exp_name_reply ) );
+
+	nbd->export_size = ntohll ( nbd->rx_exp_name_reply.export_size );
+	nbd->trans_flags = ntohs ( nbd->rx_exp_name_reply.trans_flags );
+
+	/* Discard 124 bytes if applicable */
+	if ( ! ( nbd->handshake_flags & NBD_FLAG_NO_ZEROES ) )
+		nbd->discard_len = 124;
```
mcb30, line 710: Where does this magic 124 come from?

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+	size_t len = iob_len ( iobuf );
+	DBGCP ( nbd, "NBD %p cmd %p data len: %zd\n", nbd, nbdcmd, len );
+
+	if ( nbdcmd == NULL ) {
+		DBGC ( nbd, "NBD %p no cmd to process\n", nbd );
+		return -EPROTO;
+	}
+
+	/* Sanity checks */
+	if ( nbdcmd->data_offset + len > nbdcmd->data_len ) {
```
mcb30, line 868: Is this safe against integer overflow?

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ * This function takes ownership of the I/O buffer.
+ */
+static int nbd_socket_deliver ( struct nbd_session *nbd,
+				struct io_buffer *iobuf,
+				struct xfer_metadata *meta __unused ) {
+	int ( * rx ) ( struct nbd_session *nbd );
+	size_t req_len, part_len;
+	int rc;
+
+	DBGCIO ( nbd, "NBD %p deliver %zd bytes\n", nbd, iob_len ( iobuf ) );
```
mcb30, line 905: As above

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+	nbd = zalloc ( sizeof ( *nbd ) );
+	if ( ! nbd )
+		return -ENOMEM;
+
+	ref_init ( &nbd->refcnt, nbd_free );
+	intf_init ( &nbd->block, &nbd_block_desc, &nbd->refcnt );
+	intf_init ( &nbd->socket, &nbd_socket_desc, &nbd->refcnt );
+
+	nbd->uri = uri_get ( uri );
+	name = strchr ( uri->path, '/' );
```
mcb30, line 1053: For a URI with a non-NULL host and path, `uri->path` should always start with a '/'.  What circumstances did you encounter in which this was not the case?

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+	else
+		nbd->export_name = "";
+
+	DBGC ( nbd, "NBD %p open %s (%s)\n", nbd, uri->host, nbd->export_name );
+
+	/* Use NBD_OPT_EXPORT_NAME option by default */
+	nbd->use_opt_go = 0;
+
+	if ( strstr ( uri->query, "use-opt-go" ) )
+		nbd->use_opt_go = 1;
```
mcb30, line 1065: What does `use-opt-go` do, and is it necessary to support this in iPXE?

mcb30 (changes requested):
Thanks.  I've done an initial superficial review only: this is just to address the most obvious errors and to request changes to be made prior to looking at the functionality in detail.

src/config/defaults/pcbios.h
pe-bo: OK, fixed

src/config/defaults/efi.h
pe-bo: OK, fixed

src/net/tcp/nbd.c
pe-bo: OK, debug message removed

src/net/tcp/nbd.c
pe-bo: OK, braces added

src/net/tcp/nbd.c
pe-bo: OK, fixed

src/net/tcp/nbd.c
pe-bo: OK, fixed

src/net/tcp/nbd.c
pe-bo: OK, changed

src/net/tcp/nbd.c
pe-bo: OK, changed

src/net/tcp/nbd.c
pe-bo: OK, changed

src/net/tcp/nbd.c
pe-bo: OK, comment added, function moved near relevant functions

src/net/tcp/nbd.c
pe-bo: OK, braces added

src/net/tcp/nbd.c
pe-bo: Yes, length of nbd->export_name is calculated before allocation of IOB

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+		opt->opt_export_name.request.request_magic =
+			htonll ( NBD_OPT_REQ_MAGIC );
+		opt->opt_export_name.request.option =
+			htonl ( NBD_OPT_EXPORT_NAME );
+		opt->opt_export_name.request.length = htonl ( name_len );
+		strcpy ( opt->opt_export_name.export_name, nbd->export_name );
+	}
+
+	DBGCIO ( nbd, "transmit:\n" );
+	DBGCIO_HD ( nbd, opt, opt_len );
```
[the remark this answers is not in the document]
pe-bo: OK, `DBGCIO` changed to `DBGCP`. I would like to keep the possibility of a deeper inspection of the protocol, at least for now.

src/net/tcp/nbd.c
pe-bo: The SCSI does not enforce a limit, but iSCSI itself divides the PDU into 8192 resp. 512 bytes. NBD does not have an internal division into PDUs. I think it is good to keep the limit on the maximum size of network requests.

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+		 ( nbdcmd->type == NBD_BLOCK_READ ||
+		   nbdcmd->type == NBD_BLOCK_WRITE ) );
+
+	request.request_magic = htonl ( NBD_REQUEST_MAGIC );
+	request.flags = 0;
+	if ( nbdcmd->type == NBD_BLOCK_READ )
+		request.type = htons ( NBD_CMD_READ );
+	else
+		request.type = htons ( NBD_CMD_WRITE );
+	request.handle = (uint64_t) (intptr_t) nbdcmd;
```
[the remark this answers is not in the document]
pe-bo: OK, solved using the nbdcmd_new_handle function

src/net/tcp/nbd.c
pe-bo: iSCSI code uses the same unnamed constant...
OK, fixed

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ *
+ * @v nbd		NBD session
+ */
+static void nbd_tx_step ( struct nbd_session *nbd ) {
+	int ( * tx ) ( struct nbd_session *nbd );
+	int rc;
+
+	while ( 1 ) {
+		DBGCIO ( nbd, "NBD %p try to transmit from state %d\n",
+			 nbd, nbd->tx_state );
```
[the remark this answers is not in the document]
pe-bo: OK, changed

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1095 @@
...
+ * @v nbd		NBD session
+ * @v data		Received data
+ * @v len		Length of received data
+ * @ret rc		Return status code
+ */
+static int nbd_rx_neg_init ( struct nbd_session *nbd ) {
+	DBGC2 ( nbd, "NBD %p initial handshake:\n", nbd );
+        DBGC2_HD ( nbd, &nbd->rx_neg_init, sizeof ( nbd->rx_neg_init ) );
+
+	if ( ntohll ( nbd->rx_neg_init.init_magic ) != NBD_INIT_PASSWD ||
```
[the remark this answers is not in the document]
pe-bo: OK, changed

src/net/tcp/nbd.c
pe-bo: See https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md#newstyle-negotiation
This constant is unnamed in protocol documentation

src/net/tcp/nbd.c
pe-bo: Yes, because nbdcmd->data_len is limited by capacity.max_count

src/net/tcp/nbd.c
pe-bo: OK, changed

src/net/tcp/nbd.c
pe-bo: I modified the code so that if the path is not set, an empty name will be used (this is allowed by the NBD protocol), otherwise the leading slash will be omitted.

src/net/tcp/nbd.c
pe-bo: The NBD protocol has two options for starting the transmission phase: older NBD_OPT_EXPORT_NAME and newer NBD_OPT_GO. If a server stops supporting NBD_OPT_EXPORT_NAME, there is an option to use a newer method.

mcb30: Is there an RFC or equivalent document that requires the server to support the older format?

If not, is it viable for iPXE to try the newer format and fall back to trying the older format, so that the user doesn't have to be aware of this detail?  I try to avoid imposing that sort of burden on the user wherever possible

mcb30 (changes requested):

src/net/tcp/nbd.c
pe-bo: OK, memcpy used

src/net/tcp/nbd.c
pe-bo: The NBD_MAX_BLOCK_COUNT used

src/net/tcp/nbd.c
pe-bo: OK

src/net/tcp/nbd.c
pe-bo: This limit is imposed in sandev_rw function (sanboot.c). Conversely, using -1U as max_count can be dangerous.

src/net/tcp/nbd.c
pe-bo: As far as I know, there is no such document. The nbd server older than 3.15 doesn't  support newer option. Fallbask is possible, but requires close and reopen of connection... I'd rather improve it in the future.

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1112 @@
...
+
+/**
+ * Assign new NBD request handle
+ *
+ * @v nbdcmd		NBD command
+ */
+static void nbdcmd_new_handle ( struct nbd_command *nbdcmd ) {
+	static uint32_t handle_idx;
+
+	nbdcmd->handle = htonl ( handle_idx );
```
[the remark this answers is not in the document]
pe-bo: Yes, my mistake. Fixed.

It is interesting that it does not matter at the moment, as there can be only one outstanding request and handle can be reused.

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1112 @@
...
+
+	/* Sanity check */
+	assert ( nbdcmd != NULL &&
+		 ( nbdcmd->type == NBD_BLOCK_READ ||
+		   nbdcmd->type == NBD_BLOCK_WRITE ) );
+
+	request.request_magic = htonl ( NBD_REQUEST_MAGIC );
+	request.flags = 0;
+	if ( nbdcmd->type == NBD_BLOCK_READ )
+		request.type = htons ( NBD_CMD_READ );
```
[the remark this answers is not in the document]
pe-bo: OK, fixed

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1112 @@
...
+ * Process reply for NBD_OPT_EXPORT_NAME
+ *
+ * @v nbd		NBD session
+ * @v data		Received data
+ * @v len		Length of received data
+ * @ret rc		Return status code
+ */
+static int nbd_rx_neg_exp_name ( struct nbd_session *nbd ) {
+	DBGC2 ( nbd, "NBD %p export name reply:\n", nbd );
+        DBGC2_HD ( nbd, &nbd->rx_exp_name_reply,
```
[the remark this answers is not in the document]
pe-bo: OK, fixed

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1112 @@
...
+ * @v data		Received data
+ * @v len		Length of received data
+ * @ret rc		Return status code
+ */
+static int nbd_rx_neg_opt_info ( struct nbd_session *nbd ) {
+	uint32_t opt_type, opt_len;
+
+	DBGC2 ( nbd, "NBD %p option reply:\n", nbd );
+        DBGC2_HD ( nbd, &nbd->rx_opt_reply, sizeof ( nbd->rx_opt_reply ) );
+
```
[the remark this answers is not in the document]
pe-bo: OK, fixed

src/net/tcp/nbd.c
```diff
@@ -0,0 +1,1112 @@
...
+ * Process NBD_REP_INFO
+ *
+ * @v nbd		NBD session
+ * @v data		Received data
+ * @v len		Length of received data
+ * @ret rc		Return status code
+ */
+static int nbd_rx_neg_rep_info ( struct nbd_session *nbd ) {
+	DBGC2 ( nbd, "NBD %p rep info:\n", nbd );
+        DBGC2_HD ( nbd, &nbd->rx_rep_info_export, nbd->reply_info_length );
```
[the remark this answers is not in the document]
pe-bo: OK, fixed

JuniorJPDJ:
Oh how cool :D
It will make me not write additional tool to convert nbd to iscsi or http!
(I'm trying to dynamically generate fat images from directory to use with DOS, nbdkit is the best software able to do so I found)

NiKiZe:
While scrolling around, since there is no "hand over" between NBD in iPXE to OS this implementation is very similar to http support that does range requests?
(And don't get me wrong I love NBD)
But maybe a http -> NBD proxy could be a useful alternative here?

JuniorJPDJ:
This would be A WAAAAY slower.
For people running DOS/BIOS-based tools (I do sometimes) it would be much faster to be able to use NBD or similar protocol.

593749519:
don't know why removed pcbios and efi support, turn the #define NBD, compiled under pcbios and able to connect to nbd server but failed to boot, why?

vkornilyev1988 (commented):
Add support for the NBD (Network Block Device) protocol - Please add this protocol please!!!

TiannV:
When using an NBD connection for diskless boot, Windows seems unable to load. Are there any additional steps required for the Windows image?

vkornilyev1988:
92 / 5,000
Yes, you need a driver that sets the IP address and the NBD driver itself.
I'll help you later, after work)

ср, 15 окт. 2025 г., 06:28 ttttttw ***@***.***>:

TiannV:
I'm using this project (https://github.com/cloudbase/wnbd) as the Windows
NBD driver and have made some modifications to automatically connect to the
NBD service when the driver loads. However, when using iPXE NBD for
sanboot, Windows still cannot boot. It seems like there's some additional
work needed between iPXE NBD and system boot, which is causing me trouble.

vkornilyev1988 ***@***.***> 于2025年10月15日周三 11:12写道：

Issue #390, opened 2021-06-17
Failure to bring the Realtek RTL-8110SC/8169 interface up


vl-80:
I encountered an issue where iPXE fails to bring the Realtek `RTL-8110SC/8169` NIC controller up.

The motherboard is `ZOTAC IONITX-S-E`.

The problem is that when the OS (Linux, systemd) is gracefully shutting down or rebooting it brings the network interface down. At the very end of the shutdown process the kernel reports:

    r8169: 000:03:0b.0 enp3s11: Link is Down

and the LEDs on the card turn off.

If at this point we boot iPXE (from a USB drive) it will be unable to bring the interface up.

With the debug turned on the output on the screen is:

    REALTEK 0xe04bc appears to be an RTL8169
    MII 0xe54c restarted autonegotiation
    MII 0xe54c reset after 0 ms
    MII 0xe54c restarted autonegotiation
    MII 0xe54c registers 00-07: 1000 7949 001c c912 01e1 0000 0004 2001
    MII 0xe54c registers 08-0f: 0000 0300 0000 0000 1007 f880 0200 3000
    REALTEK 0x04bc PHY status is 04 (GMII, 10Mbps, LinkDown, HalfDuplex)

    (embedded script is calling ifopen)

    REALTEK 0x04bc ring 20 is at [bfdfa800,bfdfa840)
    REALTEK 0x04bc ring e4 is at [bfdfa900,bfdfa940)
    MII 0xe54c registers 00-07: 1000 7949 001c c912 0141 0000 0004 2001
    MII 0xe54c registers 08-0f: 0000 0300 0000 0000 1007 f880 0200 3000
    REALTEK 0x04bc PHY status is 04 (GMII, 10Mbps, LinkDown, HalfDuplex)

The only way to recover from this is to power cycle the system - physically unplug the power cord.

`CTRL + ALT + DELETE`, reset button, power button do not help. Only physically unplugging the power works.

At the same time the built-in realtek PXE boot manager suffers from the same issue. It also fails with "Media test failure". But if Linux OS is booted from a USB drive it brings the interface up.

So, there is something that Linux kernel does that brings the interface back up, but I have failed to pin point it.

What I did discover is that kernel ends up calling `genphy_suspend()` and it sets the interface down. When I commented out this function the interface was remaining working between reboots and iPXE worked just fine.

https://elixir.bootlin.com/linux/v5.4.109/source/drivers/net/phy/phy_device.c#L2000

```c
[...]

PR #394, opened 2021-06-19
Realtek RTL-8110SC/8169 fix

vl-80:
Related to issue #390

Further digging revealed that Linux calls these functions when suspending and resuming the interface:

https://github.com/torvalds/linux/blob/d9403d307dba1a71ee6462b22300c2d3be773b1c/drivers/net/phy/realtek.c#L408-L420

```c
static int rtl8211b_suspend(struct phy_device *phydev)
{
	phy_write(phydev, MII_MMD_DATA, BIT(9));


	return genphy_suspend(phydev);
}


static int rtl8211b_resume(struct phy_device *phydev)
{
	phy_write(phydev, MII_MMD_DATA, 0);


	return genphy_resume(phydev);
}
```

This behavior (modifying `MII_MMD_DATA` register) was adopted in the iPXE's realtek driver and it enabled it to bring the interface up.

### DIFF
--- a/src/drivers/net/realtek.c
+++ b/src/drivers/net/realtek.c
@@ -684,6 +684,12 @@ static int realtek_open ( struct net_device *netdev ) {
 	if ( ( rc = realtek_create_buffer ( rtl ) ) != 0 )
 		goto err_create_buffer;
 
+	/* Wakeup interface */
+	if ( rtl->phy_id == 0x001cc912 )
+	{
+		realtek_mii_write ( &rtl->mdio, 0, MII_MMD_DATA, 0 );
+	}
+
 	/* Accept all packets */
 	writel ( 0xffffffffUL, rtl->regs + RTL_MAR0 );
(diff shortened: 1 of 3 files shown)


NiKiZe:
Constant for the magic phyid would probably be preferable. Maybe dumb question, But what happens if this is done on all phyids?

vl-80:
> Constant for the magic phyid would probably be preferable.

This is fixable. But the rationale behind magic constant was that there is only one phyid for which the Linux kernel calls these functions. See this in the kernel code:

https://github.com/torvalds/linux/blob/d9403d307dba1a71ee6462b22300c2d3be773b1c/drivers/net/phy/realtek.c#L810-L819

> Maybe dumb question, But what happens if this is done on all phyids?

I do not know. The kernel only applies these calls to that particular phyid:

```c
PHY_ID_MATCH_EXACT(0x001cc912),
```

mcb30:
@vl-80 Please try the simplified version of this fix in #397

PR #397, opened 2021-06-22
[realtek] Work around hardware bug on RTL8211B

### LINKED ISSUE #390: Failure to bring the Realtek RTL-8110SC/8169 interface up

mcb30:
The RTL8211B seems to have a bug that prevents the link from coming up
unless the MII_MMD_DATA register is cleared.

The Linux kernel driver applies this workaround (in rtl8211b_resume())
only to the specific RTL8211B PHY model, along with a matching
workaround to set bit 9 of MII_MMD_DATA when suspending the PHY.
Since we have no need to ever suspend the PHY, and since writing a
zero ought to be harmless, we just clear the register unconditionally.

Debugged-by: VL-80 <vl-80@users.noreply.github.com>

Fixes: ipxe/ipxe#390 
Fixes: ipxe/ipxe#394

### DIFF
--- a/src/drivers/net/realtek.c
+++ b/src/drivers/net/realtek.c
@@ -420,6 +420,16 @@ static int realtek_phy_reset ( struct realtek_nic *rtl ) {
 		 */
 	}
 
+	/* Some cards (e.g. RTL8211B) have a hardware errata that
+	 * requires the MII_MMD_DATA register to be cleared before the
+	 * link will come up.
+	 */
+	if ( ( rc = mii_write ( &rtl->mii, MII_MMD_DATA, 0 ) ) != 0 ) {
+		/* Ignore failures, since the register may not be
+		 * present on all PHYs.
+		 */
+	}
+
 	/* Restart autonegotiation */
 	if ( ( rc = mii_restart ( &rtl->mii ) ) != 0 ) {
 		DBGC ( rtl, "REALTEK %p could not restart MII: %s\n",
--- a/src/include/mii.h
+++ b/src/include/mii.h
@@ -23,6 +23,8 @@ FILE_LICENCE ( GPL2_ONLY );
 #define MII_EXPANSION	    0x06	/* Expansion register	       */
 #define MII_CTRL1000	    0x09	/* 1000BASE-T control	       */
 #define MII_STAT1000	    0x0a	/* 1000BASE-T status	       */
+#define MII_MMD_CTRL	    0x0d	/* MMD Access Control Register */
+#define MII_MMD_DATA	    0x0e	/* MMD Access Data Register    */
 #define MII_ESTATUS	    0x0f	/* Extended Status */
 #define MII_DCOUNTER	    0x12	/* Disconnect counter	       */
 #define MII_FCSCOUNTER	    0x13	/* False carrier counter       */


vl-80:
@mcb30 I tested this simplified fix. It works well.

mcb30:
@vl-80 Thank you for testing!  Please let me know if you'd like to use a real name and e-mail address for your credit in the commit log message, rather than `vl-80@users.noreply.github.com`.  I'll merge as soon as you let me know (either way).

vl-80:
@mcb30 You can put my real name

Issue #355, opened 2021-05-10
Bring in LLDP support


Teknoenie:
Hi iPXE devs.  I'm looking to see if support for LLDP is possible or not.  While searching Google I came across this repository[1] which seems to have it, but the code looks quite old.  Would it be possible to get this merged into the primary iPXE code?

This would be incredibly useful for my site where I have ~3k computers that utilize LLDP for identifying their rack positions and naming them based on type, dimensions and the like.

[1] https://github.com/mgenov/ipxe/commit/ea0686cf7cc0dec5e5fde16ee412f7e6392b7a9e

NiKiZe:
Looks like this was sent to the iPXE devel mailing list https://lists.ipxe.org/pipermail/ipxe-devel/2014-June/003589.html

Teknoenie:
Yes, this is what lead me to check to see if it ever did get merged into iPXE, but a quick search of the iPXE code didn't show anything having LLDP support so I thought I'd double check in case I missed something.
\

NiKiZe:
The lldp commit can probably be rebased easily(?) Not sure about the settings commit.

Teknoenie:
Is there an estimation as to when this could be done?

NiKiZe:
> Is there an estimation as to when this could be done?

Maybe there was a misunderstanding.
If you want to use the LLDP commit on top of current master, it has minimal conflicts and you should be able to rebase it (on your own)

If it is something that is urgent, you could [reach out and offer to pay](https://www.fensystems.co.uk/development/ebootfeatures.php) for the time to implement the feature.

Teknoenie:
I think i've been able to rebase the code to master but I'm not really sure what I'm doing as this is my first attempt at submitting a patch to a project like iPXE and I'm not by any stretch a strong git user/coder.


Since I've never done this before I thought it would make more sense to submit a text version of the patch rather than submit any MR/PR.  I tested

```
make
make bin-x86_64-efi/snponly.efi
make bin-x86_64-efi/ipxe.efi
```
all with and without `#define LLDP` in `src/config/general.h` and it seems to not have broken anything.  I have not yet tested that it works and is able to poll LLDP information from a switch yet.  Hopefully I'll get to that later today.

If this is acceptable, how should I go about properly submitting this patch?

[lldp-merge.txt](https://github.com/ipxe/ipxe/files/6724604/lldp-merge.txt)

JuniorJPDJ:
[...]

PR #413, opened 2021-07-14
Bring in LLDP support

Teknoenie:
Bring in LLDP support from https://github.com/mgenov/ipxe/commit/ea0686cf7cc0dec5e5fde16ee412f7e6392b7a9e

I have tested this against the following `make` targets successfully without issue

```
make
make bin-x86_64-efi/snponly.efi
make bin-x86_64-efi/ipxe.efi
```

I've also performed a simple test and an LLDP packet was found in my environment on the interface.

Based on this I am creating this pull request.


src/include/ipxe/errfile.h.orig
```diff
@@ -0,0 +1,400 @@
+#ifndef _IPXE_ERRFILE_H
```
NiKiZe, line 1: This file is probably not supposed to be here.

src/include/ipxe/settings.h
```diff
@@ -78,6 +78,7 @@ struct setting {
 #define SETTING_AUTH_EXTRA	16 /**< Authentication additional settings */
 #define SETTING_CRYPTO		17 /**< Cryptography settings */
 #define SETTING_MISC		18 /**< Miscellaneous settings */
+#define SETTING_LLDP		15 /**< LLDP settings */
```
NiKiZe, line 81: Numbering needs fix

NiKiZe (changes requested):

src/include/ipxe/settings.h
Teknoenie: I have removed the `src/include/ipxe/errfile.h.orig` and corrected the SETTING_LINE (15->19)

antonym:
For what it's worth, we've been using LLDP support in production for years successfully.  It would be great to have this in and have this in and not have to run a fork.  It's a great feature for large scale deployments.

NiKiZe:
This relates to #355 
that mgenov/ipxe@ea0686c can be taken and rebased on current master.
That would conserve original commits.

santosh0705:
Is there anyone tested this feature? I would have but don't have hardware switches to test it.
It would be an awesome feature if this got implemented in ipxe.

santosh0705:
> We used it successfully on a very large production fleet for identification and provisioning of hosts. We had to maintain a fork to get this patch in but it works really well. I am a plus one to get it added.

Thanks @antonym for confirming.

ITJamie:
would love to see this merged in!

rissson:
Pinging @mareo for the initial work on this.

Mareo:
I am glad to learn that LLDP support has been useful to some people.

I'd advise against merging it in its current state since it introduce a crash when a LLDP frame is received while an user is browsing the LLDP setting namespace in the `config` UI.

NiKiZe:
I would like to see the original commits rebased on master rather than recreated to keep original author.

As such I do not see that this PR will be merged in its current state.
@Mareo do you think the settings logic can be reimplemented to remove the crash? Or what is the reason that it is an issue here, but not for some other settings we have that might get updated on the fly? I would also like to see your own PR on this. 😉

Mareo:
@NiKiZe I worked a fix in my old personal repository, before it was moved to GitHub https://github.com/ipxe/people-mareo-ipxe/commit/0d5ceabc3990c0b48aa580359931cc3516ea8c29. @mcb30 had some issues with this commit and it has been stuck ever since. The UI code changed a lot since then and it is likely that this approach is not longer an option.

I would gladly rebase the "LLDP support" commit on current master and open a new PR, could you give me write access on [ipxe/people-mareo-ipxe](https://github.com/ipxe/people-mareo-ipxe)?

mcb30:
> I would gladly rebase the "LLDP support" commit on current master and open a new PR, could you give me write access on [ipxe/people-mareo-ipxe](https://github.com/ipxe/people-mareo-ipxe)?

No need: just create the pull request from your own GitHub fork of iPXE.  The assorted `ipxe/people-*` repos are a legacy from the pre-GitHub days.

In terms of fixing the potential crash: the most robust approach is probably for the code to register a `.lldp` settings block for each network device at the time the network device is created (via a `struct net_driver`), and then to update the per-netdev saved LLDP packet whenever one is received (without unregistering and re-registering the settings block).

PR #448, opened 2021-08-21
[uri] bugs with ##params

NiKiZe:
Trying to fix oddity with params, ##paramsXfoo works just as well as ##params=foo, there is no check for= just one char jump.

And also found one issue with params name using remaining, no check for other termination. - not sure how to fix that tho so only added testcase.


NiKiZe:
Removed 7e9948a since that is a non issue, ff44ab7 should still be a good fix

PR #455, opened 2021-09-03
[ntp] Offset with system date might be greater than a signed long

yhemery:
If system date is completely out of sync with the reality, the resulting
offset might be too much for a signed long. For example, with a BIOS
date configured in 1921:

NTP 0xa068e6a8 sending request
NTP 0xa068e6a8 delta -1139207296 seconds


yhemery:
Pushed --force following a discovery by a colleague: When the ipxe client has a negative offset compared to its NTP server, the resulting value of the "delta" variable was incorrect.

PR #463, opened 2021-09-10
Add Xen console support

A0lson:
Add a Xen console driver so iPXE CLI can be accessed from the Xen PV console (hvc0).

(This applies when iPXE is run as an HVM guest in Xen).


src/interface/xen/xencon.c
```diff
@@ -0,0 +1,157 @@
...
+	xenevent_send(g_xen, &event);
+}
+
+
+/*
+ * Returns true if a console buffer is full:
+ */
+static bool xenconsole_full(XENCONS_RING_IDX prod, XENCONS_RING_IDX cons, unsigned size)
+{
+	return MASK_XENCONS_IDX(prod-cons, size) == (size-1);
```
A0lson, line 62: sorry, I need to change the size argument to be the buffer, not its length

A0lson: Fixed in updated revision.

PR #518, opened 2021-11-17
[intel] Add PCI ID Intel X553 0x15e4

benibr:
In our lab we have a Asrock board with a Intel X553 network card with 4x1GBit/s which uses the intelx driver but does not work with iPXE out of the box.
A colleague figured out that the PCI device ID was missing (thx Stefan).
I wanted to submit this here for other people who may want to use the same board with iPXE.


NiKiZe:
This is great, any chance that there is a fuller name/description so that it becomes unique?

NiKiZe:
This is great, any chance that there is a fuller name/description so that it becomes unique?

mcb30:
I couldn't find anything that disambiguates this from the existing 0x15e5 device, so have arbitrarily renamed this one "x553a" and merged as https://github.com/ipxe/ipxe/commit/3ad27fbe7.

Thank you!

PR #50, opened 2016-03-15
allow to build ISO image with EFI support (ipxe.eiso)

eworm-de:

PR #522, opened 2021-11-22
util/genfsimg: changes for uefi boot

eworm-de:


eworm-de:
For a long time I used my own patch from #50 but decided to go with upstream changes now.
Tried `ipxe-hybrid.iso` on _Thinkpad X250_ and _LG 24CK550N_, both failed to boot.
With the supposed changes both systems do boot successfully, building the image with `xorrisofs` and its option `-isohybrid-gpt-basdat` does the trick.

Handle with care, I did not check or test other combinations.

mcb30:
I've pushed commit https://github.com/ipxe/ipxe/commit/0f4cc4b5a based on the first commit in this PR, to add the `--uefi` option for `isohybrid`.

In my tests, this produces a working bootable hybrid image even when using `xorrisofs`.  Could you check whether or not this change alone (i.e. latest master) works on your systems?

Thanks!

eworm-de:
I did three test:

❌ Master (0f4cc4b5a7f780111c3ef2e1d4bba175896f8537) does not boot
✅ Current pull request #522 (cbaf3c436ae693065a522f0b157cac2ed7f3f199 - with `xorrisofs -isohybrid-gpt-basdat`) boots fine
❌ My change (with `xorrisofs -isohybrid-gpt-basdat`) and unconditional `isohybrid --uefi` on top breaks the boot

So looks like this is still required for me.

PR #530, opened 2021-12-02
Add docker build container

Makeshift:
Adds an option to build using Docker, bypassing requirements to install dependencies and the like on the host.

### DIFF
--- a/README
+++ b/README
@@ -1,8 +1,25 @@
-iPXE README File
-
-Quick start guide:
-
-   cd src
-   make
+# iPXE
 
 For any more detailed instructions, see http://ipxe.org
+
+## Quick start guide
+
+```sh
+   cd src
+   make
+```
+
+## Build using Docker
+
+```
+docker build -t ipxe-build docker
+docker run -it -v $(pwd)/src:/src ipxe-build
+```
+
+You can specify additional arguments to `make`, such as:
+
+```sh
+docker run -it -v $(pwd)/src:/src ipxe-build everything
+```
+
+Outputs will be available in `src/bin*`.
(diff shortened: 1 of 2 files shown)


mcb30:
Thank you for the contribution.  I fundamentally disagree with the idea of projects providing any kind of custom build environment since it encourages the use of non-standard toolchains, and so this will not be merged.

berlin4apk:
Is it possible to add a check for ALL the packages and libs to the makefile?

With a check at the beginning, 
as it is on a new OS installation, time consuming to remember all needed libs and tools. In combination of all the different build targets / settings

berlin4apk:
Or at least keep the list up-to-date in the readme file,
Draft PR https://github.com/ipxe/ipxe/pull/531
Feedback are welcome

Makeshift:
> Thank you for the contribution. I fundamentally disagree with the idea of projects providing any kind of custom build environment since it encourages the use of non-standard toolchains, and so this will not be merged.

Fair enough. My build environment was missing some of the dependencies and I couldn't find a dependency list, so I got stuck in the run-and-see-what-errors loop for a little bit. I just wanted to save others the hassle.

NiKiZe:
> > Thank you for the contribution. I fundamentally disagree with the idea of projects providing any kind of custom build environment since it encourages the use of non-standard toolchains, and so this will not be merged.
> 
> Fair enough. My build environment was missing some of the dependencies and I couldn't find a dependency list, so I got stuck in the run-and-see-what-errors loop for a little bit. I just wanted to save others the hassle.

https://ipxe.org/download#source_code has a list

Makeshift:
> https://ipxe.org/download#source_code has a list

Ah, I didn't see it, thanks!

PR #538, opened 2021-12-19
Added multibootv2 support

Luis-Hebendanz:
Added multibootv2 support by the spec found [here](https://www.gnu.org/software/grub/manual/multiboot2/multiboot.html)

### DIFF
--- a/src/config/defaults/pcbios.h
+++ b/src/config/defaults/pcbios.h
@@ -7,7 +7,7 @@
  *
  */
 
-FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
+FILE_LICENCE( GPL2_OR_LATER_OR_UBDL );
 
 #define UACCESS_LIBRM
 #define IOAPI_X86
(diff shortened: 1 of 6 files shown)


src/config/defaults/pcbios.h
```diff
@@ -30,28 +30,29 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
...
+#define SANBOOT_PROTO_HTTP	 /* HTTP SAN protocol */
+
+#define USB_HCD_XHCI /* xHCI USB host controller */
+#define USB_HCD_EHCI /* EHCI USB host controller */
+#define USB_HCD_UHCI /* UHCI USB host controller */
+#define USB_KEYBOARD /* USB keyboards */
+#define USB_BLOCK	 /* USB block devices */
+
+#define REBOOT_CMD /* Reboot command */
+#define CPUID_CMD  /* x86 CPU feature detection command */
```
NiKiZe, line 56: What is up with the whitespace changes?

stappersg:
Summary:  Please make a new pull request, fee free to  use (and abuse)  #538

Several weeks ago was in the iPXE  IRC said  integration with existing code is very important.
This first merge request, #538, makes is very hard to blend in with what is already there.

@Luis-Hebendanz    my review is written with the purpose of making iPXE better, for getting multiboot v2 support.  you might recieve my review as an attack.  Please accept the energy that the review radiates. (energy can be positive and energy can be negative)

stappersg (commented):
This change should go in a seperate diff  because it is not only for multiboot V2.

Luis-Hebendanz:
@stappersg Thanks for your review & fix!  Is your fixed fork https://github.com/ipxe/ipxe/pull/539 not enough? And I don't quite understand what you mean with code reuse? Like do you have a specific example where I should have reused code?

Luis-Hebendanz:
> This change should go in a seperate diff because it is not only for multiboot V2.

Ah yeah makes sense.

Issue #542, opened 2021-12-30
Wrong UUID again (raw instead of pretty print)


n-ice-ch:
The UUID on Virtualbox get obtained in the wrong format.
And I guess there are other situations where the UUID from iPXE is the wrong formatted.

As nowadays all relevant system do use the pretty format for UUID, we should support that in iPXE too.

Microsoft can explain it here:

- https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc978312(v=technet.10)?redirectedfrom=MSDN

There was already some code commit but unfortunatley this seems to be removed in the latest sources?

- https://github.com/ipxe/ipxe/commit/9e896d0

May can I ask to commit a final solution or at least put additional values into the settings?
${uuid}
${uuid_raw}
${uuid_conv}

NiKiZe:
Do you have a reproducible example that uses the latest version of iPXE?

n-ice-ch:
> Do you have a reproducible example that uses the latest version of iPXE?

With pleasure.

How can I provide you the information?
What you need exactly?
How can I generate DEBUG information?

I can only see which UUID get displayed with command CONFIG or by ECHO ${uuid}
My environment is a virtualbox guest V6.1.30.

NiKiZe:
> > Do you have a reproducible example that uses the latest version of iPXE?
> 
> With pleasure.
> 
> How can I provide you the information? What you need exactly? How can I generate DEBUG information?
> 
> I can only see which UUID get displayed with command CONFIG or by ECHO ${uuid} My environment is a virtualbox guest V6.1.30.

I would expect that you give a command line for vbox with guid, and iPXE showing the uuid output using echo.
Preferably over serial or similar so that no UI is needed.

Just a sample that shows the issue and is easy for anyone to reproduce.

n-ice-ch:
[pxe-client.vdi.zip](https://github.com/ipxe/ipxe/files/7796308/pxe-client.vdi.zip)
[undionly.zip](https://github.com/ipxe/ipxe/files/7796309/undionly.zip)
![wrong_uuid_marked](https://user-images.githubusercontent.com/51759985/147826032-bb18230b-bb5b-41d4-814b-9e19df0885a1.jpg)
![wrong_uuid_pure](https://user-images.githubusercontent.com/51759985/147826034-ca1fdc44-4395-46c7-aa2e-309ef8d9982a.jpg)

mcb30:
[...]

PR #545, opened 2022-01-04
[settings] Support formatting UUIDs as little-endian GUIDs

mcb30:
The RFC4122 specification defines UUIDs as being in network byte
order, but an unfortunately significant amount of (mostly Microsoft)
software treats them as having the first three fields in little-endian
byte order.

In an ideal world, any server-side software that compares UUIDs for
equality would perform an endian-insensitive comparison (analogous to
comparing strings for equality using a case-insensitive comparison),
and would therefore not care about byte order differences.

Define a setting type name ":guid" to allow a UUID setting to be
formatted in little-endian order, to simplify interoperability with
server-side software that expects such a formatting.

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -2199,7 +2199,7 @@ const struct setting_type setting_type_base64 __setting_type = {
 };
 
 /**
- * Format UUID setting value
+ * Format UUID/GUID setting value
  *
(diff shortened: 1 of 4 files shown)


mcb30:
Fixes: #542

PR #546, opened 2022-01-04
[efi] Allow executing EFI images from local filesystems without networking

ignatk:
If we try to execute an EFI image from a local filesystem (through `file:` scheme
) without opening any network interfaces, we would get the following (with
debugging enabled on efi_image.c):

```
iPXE> chain file:vmlinuz console=ttyS0
file:vmlinuz...
 ok
EFIIMAGE 0x7dffc6c8 could not identify SNP device
Could not boot: No such device (http://ipxe.org/2c048087)
```

This is because the code uses last_opened_snpdev() in efi_image_exec().
last_opened_snpdev() will return NULL, if we never opened any network
interfaces, which would be pretty common if we use iPXE as a local "on disk"
bootloader and don't need any networking at all (or our system does not have
any network interfaces).

Fix this by making NULL return from last_opened_snpdev() non-fatal. Seems its
main purpose is to provide a meaningful device path prefix for the LoadImage()
boot service call, but according to UEFI spec [1] it seems to be not important,
if we LoadImage() from memory, which is what iPXE does.

I've tested this in QEMU/EDK2 on Debian Bullseye and with this patch I'm now
able to execute the Linux kernel from disk without the need to ifopen any
network interfaces.

[1]: https://uefi.org/sites/default/files/resources/UEFI_Spec_2_9_2021_03_18.pdf

### DIFF
--- a/src/image/efi_image.c
+++ b/src/image/efi_image.c
@@ -135,7 +135,7 @@ static wchar_t * efi_image_cmdline ( struct image *image ) {
 static int efi_image_exec ( struct image *image ) {
 	EFI_BOOT_SERVICES *bs = efi_systab->BootServices;
 	struct efi_snp_device *snpdev;
-	EFI_DEVICE_PATH_PROTOCOL *path;
+	EFI_DEVICE_PATH_PROTOCOL *path = NULL;
 	union {
 		EFI_LOADED_IMAGE_PROTOCOL *image;
 		void *interface;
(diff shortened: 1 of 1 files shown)


NiKiZe:
If this gets merged, or a variant of it, remember to update the error page and remove the file: mention http://ipxe.org/2c048087

bbwharris:
👍

ignatk:
fixed `NULL` comparison, when checking errors after trying to create a device path with `snpdev` being non null

ignatk:
Going to close this for now as it needs work: while a Linux kernel can be executed with this change, the kernel itself is not able to load an `initrd` specified by an `initrd` cmdline option. I suspect that because we don't set the image path anymore, the Linux EFI stub cannot find the base path for the initrd.

PR #574, opened 2022-01-25
[efi] Enable DOWNLOAD_PROTO_HTTPS by default

NiKiZe:
HTTPS is the current standard and should be used when possible.
Very much in common with 0c25daad38b2b345fbe6c4838ea2d9c82cc08228

Signed-off-by: Christian Nilsson <nikize@gmail.com>

### DIFF
--- a/src/config/defaults/efi.h
+++ b/src/config/defaults/efi.h
@@ -25,7 +25,9 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define ACPI_EFI
 #define FDT_EFI
 
+/* Enable IPv6 and HTTPS */
 #define	NET_PROTO_IPV6		/* IPv6 protocol */
+#define	DOWNLOAD_PROTO_HTTPS	/* Secure Hypertext Transfer Protocol */
 
 #define DOWNLOAD_PROTO_FILE	/* Local filesystem access */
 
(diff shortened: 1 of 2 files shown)


mcb30:
I don't object to this in principle, but there is a cost: seeding the entropy pool will add a noticeable startup delay (sometimes 5-10 seconds) on some platforms.  Is this cost worth paying, given that many users are probably booting from a LAN-only infrastructure that probably doesn't support HTTPS (or, at least, doesn't support HTTPS using the public cross-signed CA certificates)?

NiKiZe:
> I don't object to this in principle, but there is a cost: seeding the entropy pool will add a noticeable startup delay (sometimes 5-10 seconds) on some platforms. Is this cost worth paying, given that many users are probably booting from a LAN-only infrastructure that probably doesn't support HTTPS (or, at least, doesn't support HTTPS using the public cross-signed CA certificates)?

I wonder if this is still an issue with all recent EFI improvements, and also with SB releases in mind.

We also continue to get things like [this](https://github.com/ipxe/ipxe/issues/1570#issuecomment-3626889757) and to some extent I would just like to not get that kind of posts at all.

So with that in mind, I do think that the possible delays in default builds now actually outweigh the downsides of having this enabled. For those that really really want to optimize their LAN boot experience, they can disable this on their own builds instead. Or they can come here and tell us more about their issues with it enabled so it can be improved/fixed instead (?)

mcb30:
With Secure Boot signed versions forthcoming, the plan is to establish a baseline feature list that will primarily be guided by which parts of the codebase can be included in a signed build (e.g. because they have been previously subject to a security review).  HTTPS will be one of those features, since it has previously passed review.

Sau1707:
Note, I was just checking this, it was enabled by default a few days ago in the commit  05cb930

PR #576, opened 2022-02-02
This change respects CC

dyeske:
respect CC and therefore work with other compilers and ccache


stappersg:
Looks Good To Me.

NiKiZe:
PRs with merge commits or any other irrelevant changes will not be considered.
The way to fix is to do proper rebase and then ask for re-open

PR #580, opened 2022-02-02
This is added functionality so that you can add a string or brand at buildtime

dyeske:
This makes it so you can have a different branch for each build

### DIFF
--- a/src/config/branding.h
+++ b/src/config/branding.h
@@ -187,6 +187,9 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
  */
 #define PRODUCT_SBAT_GENERATION 0
 
+#include <config/named.h>
+#include NAMED_CONFIG(branding.h)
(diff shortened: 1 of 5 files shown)


dyeske:
https://user-images.githubusercontent.com/869529/152235644-5ba28105-e5c9-4af3-a821-fe42b8a53347.mp4

This is a video I took to demonstrate the change

mcb30:
I'm happy with the idea of adding `branding.h` to the list of headers that use the named config mechanism.  I don't see any value in the "vbox brand" etc messages replacing the default tag line - let's just have empty files for `config/*/branding.h`.

Also, please ensure that the commit message follows the normal conventions for the iPXE commit log - see the existing log for examples.

Thanks!

dyeske:
I propose to have them, but in a comment.  That's what I wanted to submit in the first place, but I had to compile it and run the code to get the demo.  I will redo it.

PR #583, opened 2022-02-04
This is an updated fix for branding configs that doesn't contain

dyeske:
You may need to #undef PRODUCT_TAG_LINE before defining it  There is a small video of what this does.


dyeske:
https://user-images.githubusercontent.com/869529/152510032-9a15fdfd-cfe4-4362-b0bd-f82be6319e61.mp4

PR #586, opened 2022-02-10
[efi] Enable IMAGE_GZIP by default for AArch64

mcb30:
AArch64 kernels tend to be distributed as gzip compressed images.
Enable IMAGE_GZIP by default for AArch64 to avoid the need for
uncompressed images to be provided.

Fixes: #585 

Originally-implemented-by: Alessandro Di Stefano <aleskandro@redhat.com>

### DIFF
--- a/src/config/defaults/efi.h
+++ b/src/config/defaults/efi.h
@@ -58,4 +58,8 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define NAP_EFIARM
 #endif
 
+#if defined ( __aarch64__ )
+#define	IMAGE_GZIP		/* GZIP image support */
+#endif
+
 #endif /* CONFIG_DEFAULTS_EFI_H */


NiKiZe:
Any reason not to enable this in all (efi) builds?

mcb30:
> Any reason not to enable this in all (efi) builds?

Only AArch64 seems to routinely distribute gzip-compressed images that the bootloader is expected to decompress

aleskandro:
Thanks @mcb30 

Refers fedora coreos/fedora-coreos-tracker#1019

Rtoax:
Hi, I'm wondering, if iPXE not support gzip, does it print `Exec format error`?

aleskandro:
yes, see https://github.com/coreos/fedora-coreos-tracker/issues/1019#issue-1050194334

mcb30:
> Hi, I'm wondering, if iPXE not support gzip, does it print `Exec format error`?

You would get the same error as for any unsupported binary format, which will be https://ipxe.org/err/2e0080

Issue #260, opened 2021-02-15
Remove Multiboot module limits


mcb30:
# Description

As a user, I would like to be able to boot Multiboot images with an arbitrary number of modules, so that I can boot operating systems such as Illumos/SmartOS with an unmodified iPXE.

# Acceptance criteria

- [ ] `MAX_MODULES` removed from `multiboot.c`
- [ ] `MB_MAX_CMDLINE` removed from `multiboot.c`
- [ ] Ability to boot into Xen
- [ ] Ability to boot into SmartOS

# Background

The existing limits were imposed to work around issues with Xen in commit https://github.com/ipxe/ipxe/commit/688bac165.

Restructuring the code to perform a single call to prep_segment() for the combined length of all modules would eliminate all possible subsequent error or return paths, thereby allowing the command line and module array to be constructed at the top of free base memory.

mcb30:
SmartOS instructions at https://wiki.smartos.org/pxe-booting-smartos/

PR #594, opened 2022-02-17
Find a segment where all modules fit

jmpesp:
When performing a multiboot of SmartOS, it would error with
ERANGE_SEGMENT. This is due to the fact that the boot_archive was >300M,
and this wasn't fitting in the same segment the kernel occupied.

With the script:

    #!ipxe
    dhcp
    chain https://netboot.smartos.org/smartos.ipxe

The following output was seen:

    https://netboot.smartos.org/os/20220210T011229Z/platform/i86pc/kernel/amd64/unix... ok
    https://netboot.smartos.org/os/20220210T011229Z/platform/i86pc/amd64/boot_archive... ok
    https://netboot.smartos.org/os/20220210T011229Z/platform/i86pc/amd64/boot_archive.hash... ok
    Could not boot: Requested memory not available (https://ipxe.org/46038101)

This commit changes where modules are placed: after placing the image,
grab the memory map and find a segment that has space for *all* modules.
Set the module starting address to this instead of concatenating all
modules right after the image.

After this, SmartOS iPXE boots ok.

### DIFF
--- a/src/arch/x86/image/multiboot.c
+++ b/src/arch/x86/image/multiboot.c
@@ -179,6 +179,101 @@ static physaddr_t multiboot_add_cmdline ( struct image *image ) {
 }
 
 /**
+ * What is the total size of all modules?
+ *
+ * @v image		Multiboot image
+ * @ret total_len		Return total size in bytes
+ */
+static size_t multiboot_total_modules_size (struct image *image) {
+	struct image *module_image;
+	size_t total_len = 0;
+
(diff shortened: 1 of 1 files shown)


jmpesp:
This addresses the "Ability to boot into SmartOS" criteria of https://github.com/ipxe/ipxe/issues/260 but not the refactoring to remove limits. I also unfortunately don't have a copy of XenServer to test with.

PR #596, opened 2022-02-20
[readline] Add TAB support

AtieP:
This commit adds TAB support, so when you're typing a command and you press TAB, similar commands are shown.

Signed-off-by: AtieP <nosequiensoy8664@gmail.com>


NiKiZe:
The indentation here looks of.
Nice feature for anyone in the shell and for debugging!

In most usecases of iPXE however this is not really useful.
Could this be split out to be used by a handle, and that way be a compile time option?

AtieP:
Thanks!

I apologize for the wrong indentation, I didn't notice. I have my editor and GitHub configured so indents are two spaces. I will proceed to replace my spaces with tabs.

And, how could I make it a compile time option? Is making a macro at config/console.h and then doing an `#ifdef` on the code fine, or is there a better way? I don't really understand how to use `ipxe/tables.h` on this situation, so I'd appreciate a bit of help.

NiKiZe:
`#ifdef`s are not accepted other than in the configs, as such it needs to use tables.
I will be honest and say that I'm not very knowledgeable myself on the topic

AtieP:
Don't worry. Anyways, I think making TAB autocompletion on the shell optional is useless because pretty much any shell allows TAB autocompletion.

AtieP:
Fixed the indentation issues.

PR #598, opened 2022-02-23
Montgomery and faster ariphmetics for HTTPS

asashnov:
Hello to iPXE developers.

It's rather late, but better then never.

This pull request is not ready to merge and requires additional work. But I'm not got familiar with iPXE codebase.

These changes are made by a colleague of mine who unfortunately passed away.

He did that change in order to speed up booting when iPXE boot script downloads something over HTTPS.

On Acer Aspire One (Intel Atom 1 core 1.7 GHz CPU) this change speed up X509 4096 bit operations from 45 seconds to about 4-5, so about 10 times faster! (we used self-generated X509 certificates with 4096 key size).

I wanted to find time to prepare a proper PR against the actual code base, with tests, but couldn't.

Hope here in the community there is someone who can port that change to the actual codebase, test and accept that change. I'm sure that code is useful. Currently, for example, DigiCert uses 2048 bit root CA, but in future it can be 4096.


mcb30:
Montgomery multiplication is used for RSA as of commit https://github.com/ipxe/ipxe/commit/83ac98ce2

Issue #447, opened 2021-08-16
Feature Request: boot from local disk in uefi


edgood1:
I need in the ability to run the equivalent of `sanboot --no-describe --drive 0x80` for uefi enabled hosts. That command only will boot a local disk in legacy bios mode.

The method of allowing ipxe to exit and fall back to disk via the bootlist is not an option in our scenario.

I found this thread https://lists.ipxe.org/pipermail/ipxe-devel/2019-February/006504.html but it doesn't seem that that this feature ever was merged.

NiKiZe:
Related or close to duplicate of
#428 #213

johnnykeats:
Not sure it helps. On BlueBanquise project (https://github.com/bluebanquise/bluebanquise), we used a grub2 image with an embed script that look for OS and boot it.
It however does not work on all scenario. We could use it on almost anything, but @oxedions reported this was not working on Nvidia DGX systems.

To build the grub2 image:
```
        if [ $distribution == "Ubuntu" ]; then
            grub-mkstandalone -O $ipxe_arch-efi -o grub2_efi_autofind.img "boot/grub/grub.cfg=grub2-efi-autofind.cfg"
            grub-mkstandalone -O $ipxe_arch-efi -o grub2_shell.img "boot/grub/grub.cfg=grub2-shell.cfg"
        else
            grub2-mkstandalone -O $ipxe_arch-efi -o grub2_efi_autofind.img "boot/grub/grub.cfg=grub2-efi-autofind.cfg"
            grub2-mkstandalone -O $ipxe_arch-efi -o grub2_shell.img "boot/grub/grub.cfg=grub2-shell.cfg"
        fi
```

(Code taken from https://github.com/bluebanquise/infrastructure/blob/master/auto_builder.sh)

**grub2-efi-autofind.cfg** content is:

```
# EFI disk boot - auto find script
# Benoit Leveugle <benoit.leveugle@sphenisc.com>

echo " _________________"
echo "|# :           : #|"
echo "|  : Blue      :  |"
echo "|  :  Banquise :  |"
echo "|  :EFI - Grub2:  |"
echo "|  :___________:  |"
echo "|     _________   |"
echo "|    | __      |  |"
echo "|    ||  |     |  |"
echo "\____||__|_____|__|lc"
echo
echo " Loading modules..."
insmod part_gpt
insmod fat
insmod chain
insmod part_msdos
insmod ext2
echo
echo "Scanning, first pass..."
[...]

PR #612, opened 2022-03-11
Uefi boot (patch from Aaron Young)

wapolinar:
*Original comment from Aaron Young:*

This patch series adds support to boot a locally attached
UEFI disk. The disk must installed with a EFI System Partition
and filesystem.

This improves upon the current solution to boot a local UEFI
disk from iPXE - which requires one to exit out of iPXE to fallback
into the EFI boot manager. This solution allows iPXE to boot a
UEFI disk directly and stay in control if the boot fails.

Example syntax to boot a local UEFI disk:

> sanboot --no-describe --drive 0 --filename \EFI\redhat\grubx64.efi

This syntax is very similar and consistent with the
syntax currently used to boot a local legacy HDD, i.e.:

> sanboot --no-describe --drive 0x80

The extra --filename argument is required for UEFI as the UEFI
spec allows the bootloader to be located in non-standard
locations on the disk. The --filename is optional and if omitted,
the standard spec defined default filepath will be used, e.g.
\EFI\BOOT\BOOTX64.efi on x86_64.

This patch series also includes a new efimap command which
displays a map of all locally attached UEFI filesystems.
Example output of the efimap command is as follows:

iPXE> efimap
Drive#  [Volume Label] Path
------  -------------------
0       [ORACLE LINUX] PciRoot(0x0)/Pci(0x3,0x0)/HD(2,MBR,0x2E0E0369,0xBBC,0x3708)
1       [DATA DISK3] PciRoot(0x0)/Pci(0x5,0x0)/HD(1,GPT,F82F29A0-...,0x800,0x64000)
2       [NO VOLUME LABEL] PciRoot(0x0)/Pci(0x6,0x0)/HD(1,GPT,F82F29A0-...,0x64000)

Note that in the event that the UEFI EFI_DEVICE_PATH_TO_TEXT protocol is
not implemented on a system, the output of the efimap will contain a hex
string dump of the device path, like so:

iPXE> efimap
Drive#  [Volume Label] Path
------  -------------------
0       [ORACLE LINUX] 02010c00d041030a0000000001010600000304012a0002...
1       [DATA DISK3] 02010c00d041030a0000000001010600000404012a000100...
2       [NO VOLUME LABEL] 02010c00d041030a0000000001010600000504012a0...

The drive# from the efimap command can be passed as an argument to the
sanboot command --drive option to specify which disk/filesystem to boot.

Note that the mappings are sorted by device path which gives a
PCI BDF ordering as the device paths typically begin with
PciRoot()/Pci() nodes. This sorting will help to give a consistent
mapping between boots.

Testing:
This patchset was testing in a virtual/cloud environment with QEMU
attached PV disks and also on a baremetal system.

v2 Changes:
    - Removed 1st (strcmp/strncmp fix) patch from original series since it was pulled.
    - Improved efimap output to include the Volume Label (if present) to
      aid in identifying boot drives/filesystems.

### DIFF
--- a/src/config/general.h
+++ b/src/config/general.h
@@ -127,6 +127,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define	AUTOBOOT_CMD		/* Automatic booting */
 #define	NVO_CMD			/* Non-volatile option storage commands */
 #define	CONFIG_CMD		/* Option configuration console */
+#define EFIMAP_CMD              /* EFI Map command */
 #define	IFMGMT_CMD		/* Interface management commands */
 #define	IWMGMT_CMD		/* Wireless interface management commands */
 #define IBMGMT_CMD		/* Infiniband management commands */
--- a/src/hci/commands/efimap_cmd.c
+++ b/src/hci/commands/efimap_cmd.c
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2019 Oracle. All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ *
+ * You can also choose to distribute this program under the terms of
+ * the Unmodified Binary Distribution Licence (as given in the file
+ * COPYING.UBDL), provided that you have satisfied its requirements.
+ */
+
+
+FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
+
+#include <stdio.h>
+#include <errno.h>
+#include <getopt.h>
+#include <string.h>
+#include <ipxe/command.h>
+#include <ipxe/parseopt.h>
+#include <usr/efiboot.h>
+
+/** @file
+ *
+ * EFIMAP command
+ *
+ */
+
+/** "efimap" options */
+struct efimap_options {};
+
+/** "efimap" option list */
+static struct option_descriptor efimap_opts[] = {};
+
+/** "efimap" command descriptor */
+static struct command_descriptor efimap_cmd =
+	COMMAND_DESC ( struct efimap_options, efimap_opts, 0, 0, NULL);
+
+/**
+ * The "efimap" command
+ *
+ * @v argc		Argument count
+ * @v argv		Argument list
+ * @ret rc		Return status code
+ */
+static int efimap_exec ( int argc, char **argv ) {
+	struct efimap_options opts;
+	int rc;
+
+	/* Parse options */
+	if ( ( rc = parse_options ( argc, argv, &efimap_cmd, &opts ) ) != 0 )
+		return rc;
+
+	efi_boot_display_map ();
+
+	return 0;
+}
+
+/** EFIMAP command */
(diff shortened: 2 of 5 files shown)


wapolinar:
Just added this patch because of #447

marcohald:
Sorry for pinging @mcb30 is there any work to do before this patch can be merged.
It would be great to fall back to local Disks from an UEFI IPXE Menu

oxedions:
We are also very interested by this feature, this would allow smoother UEFI boots on disk 😊

chris-tux:
I would also be very interested. It would help me with 150 clients which I could then also boot with iPXE.

jcpunk:
Also interested

huntersan9:
Also want to add my interest for this feature.

stappersg:
Hi Aaron Young,

I think that you are the @ajyoung-oracle.  I did just see https://github.com/ipxe/ipxe/pull/136#event-7113216025

Which git repository (and which branch of it) has the commit that this merge request is based on?


Regards
Geert Stappers
-- 
Silence is hard to parse.

ajyoung-oracle:
Hi. I don't believe that I ever pushed the patchset to my public github repo. I only submitted the patches to the ipxe-devel email list. I never pursued it further than that.

stappersg:
FWIW
* https://lists.ipxe.org/pipermail/ipxe-devel/2019-February/thread.html
* https://lists.ipxe.org/pipermail/ipxe-devel/2019-April/006644.html

TomKranenburg:
Tried it and it works great. Thanks for the extra efimap command as well. You didn't need to put that in but it made things much easier.

Would love for this to be included in the main branch if possible.

sandynomad:
+1 for writing this update. 

Helped me understand the EFI disk layout.

mcb30:
This patch could not have been merged as-is: beyond the extra `efimap` command, it duplicates a substantial amount of logic from `efi_block.c` but with some modifications, which would create extra maintenance work in future.  There is simply no capacity to accommodate any change such as this which adds technical debt.

The equivalent functionality is now present upstream as of commit https://github.com/ipxe/ipxe/commit/b1c13cc43

TomKranenburg:
> This patch could not have been merged as-is: beyond the extra `efimap` command, it duplicates a substantial amount of logic from `efi_block.c` but with some modifications, which would create extra maintenance work in future. There is simply no capacity to accommodate any change such as this which adds technical debt.
> 
> The equivalent functionality is now present upstream as of commit [b1c13cc43](https://github.com/ipxe/ipxe/commit/b1c13cc43)

Totally appreciate this. I imagine the maintenance overhead for something like this is already pretty wild. I will be deploying this across a university as soon as I get back @mcb30.

Thank you once again.

TomKranenburg:
> > Totally appreciate this. I imagine the maintenance overhead for something like this is already pretty wild. I will be deploying this across a university as soon as I get back @mcb30.
> 
> Please let me know how it works out for you!
> 
> I'll also be adding a couple of extra small `sanboot` features soon to allow the boot partition to be selected by UUID, volume label, or existent filename (which might potentially be useful in your use case).

You've already improved it as far as I can see. Basically if you plug a USB drive into our iPXE systems currently it will fail to boot(I'm using a custom patch built from this pull request). As the San disk numbers change. If I'm reading this right you've assigned them static values that don't change? I am under the hoot just cycling through the first dozen or so number each boot until we get to one that is actually bootable. In the rare event someone boots the system with an external drive plugged in.

This package is the lynchpin in a system we use to admin hundreds of systems. Couldn't live without it. The migration to UEFI was a tough one but with the help of this package we did it @mcb30. Many thanks from the SDI at Abertay University.

mcb30:
> This package is the lynchpin in a system we use to admin hundreds of systems. Couldn't live without it. The migration to UEFI was a tough one but with the help of this package we did it @mcb30. Many thanks from the SDI at Abertay University.

@TomKranenburg You're very welcome!  I've just updated the documentation at https://ipxe.org/cmd/sanboot to include all the new features - it sounds as though using `--drive 0` and some filtering options could be very useful to you.

wapolinar:
@mcb30 thank you very much for b1c13cc43, this helped us a lot and also allowed us to get rid of this patch in our build cycle.

Needed some more time to test the changes, but they work fine with our setup, thank you again!

PR #626, opened 2022-03-26
[rng] Add more entropy

9001:
When running as efirom, the RNG is always seeded with the same value (0xc0, sometimes +/- 1).

Add some entropy to fix this:
* address of stack variable
* profile_timestamp()

This alleviates some separate issues:
* RNG seeded with `4` when running ipxe.efi inside VMwarePlayer or QEMU due to `currticks() called before initialisation`
* Servers getting confused by reused TCP sequence numbers and DHCP request IDs across reboots

The size of `random.o` has increased from 4260 to 5080 bytes

Tested on an i350-t4v2 in a Dell Optiplex 3090 (bios 2.2.0) using commit dd3547543811dd2e996b910bdadb609414904352, and QEMU/VMwarePlayer on f58b5109f46088bdbb5345a9d94b636c54345bdf

### DIFF
--- a/src/core/random.c
+++ b/src/core/random.c
@@ -8,6 +8,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 
 #include <stdlib.h>
 #include <ipxe/timer.h>
+#include <ipxe/profile.h>
 
 static int32_t rnd_seed = 0;
 
@@ -15,11 +16,21 @@ static int32_t rnd_seed = 0;
  * Seed the pseudo-random number generator
  *
  * @v seed		Seed value
+ *
+ * Will use the following entropy sources if @c seed is 0:
+ *  - system time ticks
+ *  - cpu profiling timestamp
+ *  - address of stack variable
+ *  - fair dice roll (ensures nonzero seed)
  */
 void srandom ( unsigned int seed ) {
-	rnd_seed = seed;
-	if ( ! rnd_seed )
-		rnd_seed = 4; /* Chosen by fair dice roll */
+	if ( ! ( rnd_seed = seed ) ) {
+		rnd_seed = ( currticks()
+			^ profile_timestamp()
+			^ ( size_t ) &seed
+		) | 4;
+	}
+	DBG ( "seed=%08x ", rnd_seed );
 }
 
 /**
@@ -31,7 +42,7 @@ long int random ( void ) {
 	int32_t q;
 
 	if ( ! rnd_seed ) /* Initialize linear congruential generator */
-		srandom ( currticks() );
+		srandom ( 0 );
 
 	/* simplified version of the LCG given in Bruce Schneier's
 	   "Applied Cryptography" */
--- a/src/core/random.c
+++ b/src/core/random.c
@@ -21,14 +21,16 @@ static int32_t rnd_seed = 0;
  *  - system time ticks
  *  - cpu profiling timestamp
  *  - address of stack variable
- *  - fair dice roll (ensures nonzero seed)
+ *
+ * The RNG implementation requires that the seed is non-zero;
+ * this function guarantees that with `| 4`
  */
 void srandom ( unsigned int seed ) {
 	if ( ! ( rnd_seed = seed ) ) {
 		rnd_seed = ( currticks()
 			^ profile_timestamp()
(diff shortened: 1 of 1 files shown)


src/core/random.c
```diff
@@ -8,18 +8,29 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
...
  */
 void srandom ( unsigned int seed ) {
-	rnd_seed = seed;
-	if ( ! rnd_seed )
-		rnd_seed = 4; /* Chosen by fair dice roll */
+	if ( ! ( rnd_seed = seed ) ) {
+		rnd_seed = ( currticks()
+			^ profile_timestamp()
+			^ ( size_t ) &seed
+		) | 4;
```
stappersg, line 31: Please keep the `/* Chosen by fair dice roll */`

Or add the `alt_text` from https://xkcd.com/221/
> "RFC 1149.5 specifies 4 as the standard IEEE-vetted random number."

src/core/random.c
```diff
@@ -31,7 +42,7 @@ long int random ( void ) {
 	int32_t q;
 
 	if ( ! rnd_seed ) /* Initialize linear congruential generator */
-		srandom ( currticks() );
+		srandom ( 0 );
```
stappersg, line 45: That `s/currticks()/0/` feels wrong in a commit message titled _Add more entropy_.

src/core/random.c
9001: Sure, I moved it into the doxygen comment instead, but keeping an additional inline copy is aight with me :+1:

src/core/random.c
9001: The idea was to make it obvious inside `srandom` whether we are given an exact seed value (for example from the tests), or if additional entropy needs to be added. So since `0` cannot be used as a seed, it seemed like the best approach. And, I figured it would make sense to keep all the entropy sources in one place -- unless there's a reason i'm missing to keep some of them separate :>

stappersg:
Regarding the _That `s/currticks()/0/` feels wrong in a commit message titled _Add more entropy_.

Thing I want to express is that a patch should only change 1 thing. This time:
* Only add more entropy
* Leave the _if statement_ inside function `srandom()` **un**touched
* Leave the calling `	srandom ( currticks() );` also **untouched**

```diff
--- a/src/core/random.c
+++ b/src/core/random.c
@@ -8,6 +8,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 
 #include <stdlib.h>
 #include <ipxe/timer.h>
+#include <ipxe/profile.h>
 
 static int32_t rnd_seed = 0;
 
@@ -15,11 +16,22 @@ static int32_t rnd_seed = 0;
  * Seed the pseudo-random number generator
  *
  * @v seed		Seed value
+ *
+ * * Will use the following entropy sources if @v seed is 0:
+ * - system time ticks
+ * - cpu profiling timestamp
+ * - address of stack variable
+ * - fair dice roll (ensures nonzero seed)
  */
 void srandom ( unsigned int seed ) {
 	rnd_seed = seed;
-	if ( ! rnd_seed )
-		rnd_seed = 4; /* Chosen by fair dice roll */
+	if ( ! rnd_seed ) {
+		rnd_seed = ( currticks()
+			^ profile_timestamp()
+			^ (size_t) &seed
+		) ^ 4; /* Chosen by fair dice roll */
+	}
+	DBG ( "seed=%08x ", rnd_seed );
 }
 
 /**
```

Do notice that I changed the _or 4_ into  _execlusive or 4_.

Euh, yes the
```c
+	DBG ( "seed=%08x ", rnd_seed );
```
is actually for a different patch.

NiKiZe:
I think this looks quite good as is, make sure that the fallback is documented, and also why it is 0

9001:
Alright, improved the comments :+1: 

The `OR 4` (instead of XOR) was intentional btw! Since there is a tiny chance that the other entropy sources could result in the value `4`, which would cause the RNG to break. Super convenient that `4` happens to only set a single bit :>

9001:
Closing this PR since it has most likely been superceded by #1052, with #1042 being very similar to the issues I encountered.

However I believe an xor with rdtsc (when available) would be an even better fix, as not all boards have a working cmos rtc

PR #639, opened 2022-05-01
Have the system use `env` to find echo

Krutonium:
On my (NixOS) system it was showing errors because the `echo` binary is not where it might normally be. Using `/usr/bin/env` should be universal and ensure that the echo binary can be found.

### DIFF
--- a/src/Makefile.housekeeping
+++ b/src/Makefile.housekeeping
@@ -10,8 +10,8 @@
 TAB 			:= $(shell $(PRINTF) '\t')
 ECHO_E_ECHO		:= $(ECHO)
 ECHO_E_ECHO_E		:= $(ECHO) -e
-ECHO_E_BIN_ECHO 	:= /bin/echo
-ECHO_E_BIN_ECHO_E 	:= /bin/echo -e
+ECHO_E_BIN_ECHO 	:= /usr/bin/env echo
+ECHO_E_BIN_ECHO_E 	:= /usr/bin/env echo -e
 ECHO_E_ECHO_TAB		:= $(shell $(ECHO_E_ECHO) '\t' | cat)
 ECHO_E_ECHO_E_TAB	:= $(shell $(ECHO_E_ECHO_E) '\t' | cat)
 ECHO_E_BIN_ECHO_TAB 	:= $(shell $(ECHO_E_BIN_ECHO) '\t')


NiKiZe:
This adds new dependency on `/usr/bin/env` in Makefile since it is not used anywhere before? How they are used in utils.

One could also argue that NixOS is the odd one here, but unless there is a good reason against this I think it should be valid to be merged. (after confirming to https://ipxe.org/admin, please update commit and PR, force push will be needed)

stappersg:
Idea:  Install `/bin/echo` as build dependency.

My websearch on `NixOS /bin/echo` yielded:
* [known problem](https://github.com/hmemcpy/milewski-ctfp-pdf/issues/208)
* [workaround](https://github.com/hmemcpy/milewski-ctfp-pdf/issues/209)

Krutonium:
> This adds new dependency on `/usr/bin/env` in Makefile since it is not used anywhere before? How they are used in utils.
> 
> One could also argue that NixOS is the odd one here, but unless there is a good reason against this I think it should be valid to be merged. (after confirming to https://ipxe.org/admin, please update commit and PR, force push will be needed)

 I suppose it does, but I have yet to see a system that doesn't have `/usr/bin/env`, including on systems like mine where echo is located at `/run/current-system/sw/bin/echo` which is actually a symlink to `/nix/store/x5f50z1kgdxmbjw65p84ifqybphc12c3-coreutils-full-9.0/bin/echo`
 
 

> Idea: Install `/bin/echo` as build dependency.
> 
> My websearch on `NixOS /bin/echo` yielded:
> 
>     * [known problem](https://github.com/hmemcpy/milewski-ctfp-pdf/issues/208)
> 
>     * [workaround](https://github.com/hmemcpy/milewski-ctfp-pdf/issues/209)

Perhaps using printf would indeed be cleaner.

felixsinger:
I'm using NixOS as well and having the same issue. Could this be fixed please?

felixsinger:
Hey @NiKiZe, could you have a look on this again? Anything I can do in order to make progress here?

NiKiZe:
If this PR was updated to confirm to iPXE commit standard and rebased, maybe.

Or just use a VM or docker image to build to work around your distros "brokenness"

PR #698, opened 2022-08-03
[nfs] Fix potential buffer overflow condition

ajyoung-oracle:
Fix potential buffer overflow condition in nfs_uri.c.

Signed-off-by: Aaron Young <aaron.young@oracle.com>


stappersg (approved):
Yeah, it makes sense to `malloc()` an extra `sizeof(char)`.

justinclift:
@mcb30 This seems like it should be merged?

mcb30:
> @mcb30 This seems like it should be merged?

It's a definite bugfix, but in all honesty the whole NFS and 802.11 code needs replacing because it's full of errors such as these.  (This is why those two code areas are not allowed in Secure Boot builds.)

justinclift:
No worries. :smile:

PR #723, opened 2022-08-26
[build] Respect $CC

dyeske:
On BSD systems 'gcc' is not the default compiler.
But $CC is honored.  This also makes it build with ccache.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -16,7 +16,7 @@ SYMBOL_PREFIX	:=
 #
 # Locations of tools
 #
-HOST_CC		:= gcc
+HOST_CC		:= $(CC)
 RM		:= rm -f
 TOUCH		:= touch
 MKDIR		:= mkdir
@@ -26,7 +26,7 @@ PRINTF		:= printf
 PERL		:= perl
 PYTHON		:= python
 TRUE		:= true
-CC		:= $(CROSS_COMPILE)gcc
+CC		:= $(CROSS_COMPILE)$(CC)
 CPP		:= $(CC) -E
 AS		:= $(CROSS_COMPILE)as
 LD		:= $(CROSS_COMPILE)ld


dyeske:
I did it again and hopefully this time everything is ready to merge.  Thanks again for your patience.

stappersg (approved):
Looks good to me.

dyeske:
Is this fix going to be merged in?

NiKiZe:
Thank you for your contribution, and attempts at making iPXE better. Despite the noise ;)

Official stance as I remember it is "iPXE builds on Linux, BSD not supported"
Adding more factors that could impact runtime is not desired.

stappersg:
afbeelding

habetsm-xilinx (approved):

dyeske:
Yes that is expected behavior on a default FreeBSD installation.  I have
tested this on FreeBSD 11, 12, and 13.  I have tested gcc 8,9,10,11, and
12.  I usually compile with ccache as well on Linux or FreeBSD.  When I
compiled on FreeBSD a couple of minutes ago I set
CC=/usr/local/libexec/ccache/world/gcc12.  I've been compiling ipxe on
FreeBSD with gcc for many years, and I don't see a compelling reason to
bother getting it to compile with clang.

mcb30:
The problem is that the patch seems to break compilation for someone with a default FreeBSD installation.  Currently it will work since it will explicitly use `gcc`.  With this patch, it will suddenly attempt to use `clang` and will fail.  Am I missing something?

dyeske:
I don't think it's a worthy goal to be concerned with this.  A default
FreeBSD installation doesn't have gcc, syslinux, qemu-img, bash, or
mkisofs.  Also when gcc is installed on FreeBSD it doesn't install a binary
called gcc.  The binary has the version in it because FreeBSD allows you to
have multiple version of gcc installed at the same time.  I think you might
get some insight from the FreeBSD port of ipxe here
https://github.com/freebsd/freebsd-ports/blob/main/net/ipxe/Makefile

mcb30:
OK, I'm happy to defer to your judgement on this since I'm not a FreeBSD user and don't know how people generally manage their FreeBSD systems.

However, we cannot merge this as-is because it fails the GitHub CI tests: there is no `aarch64-linux-gnu-cc` or `arm-none-eabi-cc` binary.  Somehow, it seems as though this PR managed to bypass running the CI tests. :shrug: 

I can't retrigger CI for this PR, but I've created #875 which is identical to this PR (with a reworded commit message).  You can see there that the ARM32 and ARM64 tests both fail.

dyeske:
Are you saying I need to fix the arm cross compile of ipxe on FreeBSD?

mcb30:
No, I'm saying that this patch breaks cross-compiling on Linux.  Since cross-compiling on Linux is part of the per-commit CI checks, that isn't allowed to be broken.

dyeske:
I made a small change that should fix CC being respected and also fix the cross compile and also work with ccache.  For example I can do this now.

make -j 4 -C src NO_WERROR=1 V=1 CROSS=arm-none-eabi- CC=/usr/lib/ccache/arm-none-eabi-gcc bin-arm32-efi/intel.efi bin-arm32-efi/intel.usb bin-arm32-efi/intel.iso

src/Makefile
```diff
@@ -16,7 +16,8 @@ SYMBOL_PREFIX	:=
 #
 # Locations of tools
 #
-HOST_CC		:= gcc
+CC		?= gcc
+HOST_CC		:= $(CC)
```
mcb30, line 20: This latest version of the change seems to affect only `HOST_CC`.  Since you still later have the line `CC := $(CROSS_COMPILE)gcc`, it looks as though this will always try to use a binary `gcc` for most of the build?

dyeske: See above comment.  The same concept works on FreeBSD and Linux by setting
CC=/usr/local/libexec/ccache/world/gcc12 or CC=/usr/lib/ccache/gcc.  I tested cross compiling on arm and arm64.

mcb30: With a test `Makefile` containing:
```makefile
$(warning CC=$(CC))
CC ?= gcc
$(warning CC=$(CC))
HOST_CC := $(CC)
$(warning HOST_CC=$(HOST_CC))
CC := $(CROSS_COMPILE)gcc
$(warning CC=$(CC))
all:
```
(i.e. matching this PR), I get the following:
```console
$ make
Makefile:1: CC=cc
Makefile:3: CC=cc
Makefile:5: HOST_CC=cc
Makefile:7: CC=gcc
make: Nothing to be done for 'all'.
```
but
```console
$ CC=some-other-compiler make
Makefile:1: CC=some-other-compiler
Makefile:3: CC=some-other-compiler
Makefile:5: HOST_CC=some-other-compiler
Makefile:7: CC=gcc
```
i.e. this definitely does **not** respect the value of CC set as an environment variable.

stappersg: > this definitely does not respect the value of CC set as an environment variable

```text
make CC=some-other-compiler
```
as in
```text
make -j 4 -C src NO_WERROR=1 V=1 \
 CROSS=arm-none-eabi- \
 CC=/usr/lib/ccache/arm-none-eabi-gcc \
   bin-arm32-efi/intel.efi \
   bin-arm32-efi/intel.usb \
   bin-arm32-efi/intel.iso
```
from above comment (fifty minutes ago).

mcb30: > ```
> make CC=some-other-compiler
> ```

That is irrelevant.  Passing variables as parameters on the `make` command line will always override whatever is set in the `Makefile`.  Even the current `Makefile` will "respect $CC" in that sense.  This PR is only relevant to a `CC` passed in as an environment variable.

dyeske:
I haven't gotten a reply for a year.  What do I need to do?

NiKiZe:
Superseeded by #875

PR #732, opened 2022-09-06
fix dhcpv6 with non /64 prefix

### LINKED ISSUE #449: ipv6_create_routes() - DHCPv6 addressing in non-/64 networks
Hi! 

I'm currently trying to use iPXE in an IPv6-only setup, where each system has their own little IPv6 subnet and is directly talking to its router for all communication.
All networks are /126 for this reason, there's no IPv4 or anything.

Router advertisements are sent and look like this: 
```php
interface br0
{
	AdvSendAdvert on;
	AdvManagedFlag on;
	AdvOtherConfigFlag on;
	AdvReachableTime 0;
	AdvRetransTimer 0;
	AdvCurHopLimit 64;
	AdvDefaultLifetime 30;
	AdvHomeAgentFlag off;
	AdvDefaultPreference medium;
	AdvSourceLLAddress on;

	prefix 2001:db8:0:3::4/126
	{
		AdvValidLifetime 2592000;
		AdvPreferredLifetime 604800;
		AdvOnLink on;
		AdvAutonomous on;
		AdvRouterAddr off;
	};
};
```

As you can see, Managed and OtherConfig flags are enabled and the network is smaller than /64, so iPXE cannot assign an IPv6 address via SLAAC.
After receiving this ICMPv6 RA packet, iPXE will start to query the DHCPv6 server and receive an IPv6 address. 

iPXE now has an IPv6-address (via DHCPv6) and knows the subnet length and default gateway (via ICMPv6 RA).
In the settings, net0 is filled correctly: 

net0.ndp.0 only has the gateway and subnet length, though: 

and net0.dhcpv6 has the IPv6 address, DNS server, etc.: 

This is the debug output and route table:
```
iPXE> ifconf -c ipv6
Configuring [ipv6] (net0 52:54:00:f2:4f:11)...NDP net0 found default router fe80::1e34:daff:fec8:1369 for on-link autonomous prefix 2001:db8:0:3::4/126
DHCPv6 net0 has XID ce980f
DHCPv6 net0 entering SOLICIT state
NDP net0 found default router fe80::1e34:daff:fec8:1369 for on-link autonomous prefix 2001:db8:0:3::4/126
DHCPv6 net0 received ADVERTISE from fe80::1e34:daff:fec8:1369
DHCPv6 net0 received ADVERTISE is for 2001:db8:0:3::6
DHCPv6 net0 entering REQUEST state
DHCPv6 net0 received REPLY from fe80::1e34:daff:fec8:1369
DHCPv6 net0 received REPLY is for 2001:db8:0:3::6
DHCPv6 net0 complete
 ok
iPXE> route
net0: 2001:db8:0:3::6/64
net0: fe80::5054:ff:fef2:4f11/64
```

The route on net0 does not have a gateway address attached to it.
Communication is therefore broken to all v6-targets outside of the Layer 2.

I've investigated this further and found the function ipv6_create_routes() in ipv6.c.
ipv6_create_routes() accepts a settings object (will get "net0" from the caller) and will run through the settings-tree recursively.
It needs 3 prerequisits to work properly:
- setting `ip6`
- setting `len6`
- setting `gateway6`

Here's the catch: 
All of these settings must be in the same layer of the settings tree. Also: The first layer (net0, which would have all 3 options) is ignored.
The way ipv6_create_routes() is built, requires all settings to be in the same layer and the `( origin != settings )`-check requires them to not be in `net0` but a further-down layer.

Just comment the `( origin != settings )` check in https://github.com/ipxe/ipxe/blob/e09e1142a3bd8bdb702efc92994c419a53e9933b/src/net/ipv6.c#L1292 out and it starts to work perfectly.

I'm not sure, which side effects this has, but I've tested a regular SLAAC /64 network with the changes and it worked fine as well.

How could this be fixed properly upstream?

nkraetzschmar:
ifconf is broken when using dhcpv6 on a non /64 prefix (tested with /127), such that the prefix used for the dhcpv6 address will default to /64 and no gateway gets set.

The origin of this issue seems to be two-fold:

1. If the auto flag is set in the prefix option of the router adv then `ndp_prefix_fetch_ip6` will attempt to use SLAAC, which is currently only implemented for a /64 prefix, so it will return `EINVAL` since the prefix length don't match.
This results in the setting `net0.ndp.0/ip6` not being set / fetched, so the values for `gateway6` and `len6` in `net0.ndp.0` will be ignored in `ipv6_create_routes`.

2. There appears to be some logic bug in `ipv6_match_len` such that a matching length of 127 incorrectly returns 120.
This results in the address received from dhcpv6 not being matched to the prefix and gateway options from ndp in `ipv6_add_miniroute`

This PR should fix the problem by ignoring the auto flag if the router adv has the managed flag set (i.e. when using dhcpv6 rather than SLAAC) and replacing the `ipv6_match_len` function with an (admittedly less performant) version that does not contain the same logic bug.

---

should fix #449

### DIFF
--- a/src/net/ndp.c
+++ b/src/net/ndp.c
@@ -612,6 +612,8 @@ struct ndp_settings {
 	struct in6_addr router;
 	/** Router lifetime */
 	unsigned int lifetime;
+	/** Router advertisement flags (used to check if using dhcpv6 in ndp_prefix_fetch_ip6) **/
+	uint8_t radv_flags;
 	/** Length of NDP options */
 	size_t len;
 	/** NDP options */
@@ -802,8 +804,14 @@ static int ndp_prefix_fetch_ip6 ( struct settings *settings, void *data,
 		prefix_len = ipv6_eui64 ( &ip6, netdev );
 		if ( prefix_len < 0 )
 			return prefix_len;
-		if ( prefix_len != prefix->prefix_len )
-			return -EINVAL;
+		if ( prefix_len != prefix->prefix_len ) {
+			if ( ndpset->radv_flags & NDP_ROUTER_MANAGED ) {
+				// if SLAAC fails, but dhcpv6 is in use keep the prefix address
+				memcpy ( &ip6, &prefix->prefix, sizeof ( ip6 ) );
+			}
+			else
+				return -EINVAL;
+		}
 	}
 
 	/* Fill in IPv6 address */
@@ -934,6 +942,7 @@ static struct settings_operations ndp_prefix_settings_operations = {
 static int ndp_register_settings ( struct net_device *netdev,
 				   struct in6_addr *router,
 				   unsigned int lifetime,
+				   uint8_t radv_flags,
 				   union ndp_option *options, size_t len ) {
 	struct settings *parent = netdev_settings ( netdev );
 	union ndp_option *option;
@@ -980,6 +989,7 @@ static int ndp_register_settings ( struct net_device *netdev,
 	ndpset->settings.order = order;
 	memcpy ( &ndpset->router, router, sizeof ( ndpset->router ) );
 	ndpset->lifetime = lifetime;
+	ndpset->radv_flags = radv_flags;
 	ndpset->len = len;
 	memcpy ( ndpset->options, options, len );
 	prefset = ( ( ( void * ) ndpset->options ) + len );
@@ -1200,7 +1210,7 @@ ipv6conf_rx_router_advertisement ( struct net_device *netdev,
 	/* Register NDP settings */
 	option_len = ( len - offsetof ( typeof ( *radv ), option ) );
 	if ( ( rc = ndp_register_settings ( netdev, router,
-					    ntohl ( radv->lifetime ),
+					    ntohl ( radv->lifetime ), radv->flags,
 					    radv->option, option_len ) ) != 0 )
 		return rc;
 


mcb30:
Thanks for the PR.

>     2. There appears to be some logic bug in `ipv6_match_len` such that a matching length of 127 incorrectly returns 120.
>        This results in the address received from dhcpv6 not being matched to the prefix and gateway options from ndp in `ipv6_add_miniroute`

A quick test suggests that `ipv6_match_len()` behaves perfectly correctly.  However, your PR prompted me to check for other possible causes for your symptoms, which lead me to find and [fix a bug in the code that constructs the prefix mask](https://github.com/ipxe/ipxe/commit/bc19aeca5).

Could you please rebase your PR on current master (deleting the unnecessary change to `ipv6_match_len()`), and recheck the logic of your other commit?

nkraetzschmar:
thanks for the quick response

yes, with the prefix mask fix, it now works without any changes to `ipv6_match_len`

mcb30:
> yes, with the prefix mask fix, it now works without any changes to `ipv6_match_len`

Great.  Could you please explain the logic of the remaining commit?  It is valid to have both SLAAC and DHCPv6 managed addresses at the same time, but your patch will break this.

nkraetzschmar:
The problem is when using a prefix of /127 for example SLAAC fails as `ipv6_eui64` returns 64 instead of 127, so the `prefix_len != prefix->prefix_len` check in `ndp_prefix_fetch_ip6` causes it to return `-EINVAL`. This results in the ndp setting being ignored in `ipv6_create_routes`.

I was not aware the combination of managed and auto flags was valid, so simply ignoring auto when the managed bit is set appeared to be the easiest solution...

I guess an alternative fix/workaround would be to just replace the `return -EINVAL` with `memcpy ( &ip6, &prefix->prefix, sizeof ( ip6 ) )` to reset the address from any changes done by `ipv6_eui64` and maintain the prefix address iff SLAAC fails. This may not be the desired behaviour when not using dhcpv6 though. :thinking:

So instead I've updated this PR to keep track of the router advertisement flags in the ndp settings and then only maintain the prefix address if SLAAC fails and the managed bit was set on the radv flags.

mcb30:
From my reading of https://www.rfc-editor.org/rfc/rfc4862#section-5.5.3 your router is sending invalid advertisements if it is setting the Autonomous bit for a /127 prefix, since Ethernet uses a 64-bit interface identifier.

If you fix the router to not set the Autonomous bit (e.g. via `AdvSendAutonomous off` for `radvd`), does the problem go away?

nkraetzschmar:
yes, when not sending the auto flag in the prefix options, the problem does go away.

the problem only occurs when using a combination of non /64 prefix and auto flag, but I guess if this combination is invalid anyways, then everything is fine now that the prefix mask stuff is fixed.

thanks for the help :)

MalteJ:
@mcb30 how would you set the onlink route? The address that is configured via DHCP is a /128. The RA tells the server which router provides a default route. So when we have for example a /112 subnet, how does this info get pushed to the server?

Afaik you need to announce the /112 prefix via the RA. And if you disable the `auto` config flag, the client would ignore this prefix.

MalteJ:
hmm, do you need to use the L flag instead of the A flag?

MalteJ:
> From my reading of https://www.rfc-editor.org/rfc/rfc4862#section-5.5.3 your router is sending invalid advertisements if it is setting the Autonomous bit for a /127 prefix, since Ethernet uses a 64-bit interface identifier.
> 
> If you fix the router to not set the Autonomous bit (e.g. via `AdvSendAutonomous off` for `radvd`), does the problem go away?

We see multiple RA implementations sending the auto flag also with prefix lengths != 64 (e.g. systemd-networkd and frr). Can we introduce a check in ipxe that uses the prefix only for SLAAC when it has a 64 bit length? Additionally, if the L flag is set, we should install the prefix as an onlink route.

mcb30:
> We see multiple RA implementations sending the auto flag also with prefix lengths != 64 (e.g. systemd-networkd and frr). Can we introduce a check in ipxe that uses the prefix only for SLAAC when it has a 64 bit length?

@MalteJ I've pushed https://github.com/ipxe/ipxe/commit/8f5fc1614 to cause iPXE to ignore SLAAC on misconfigured prefixes.

> Additionally, if the L flag is set, we should install the prefix as an onlink route.

This should already happen.

PR #755, opened 2022-09-25
[pxe_udp] Place an upper bound on number of queued packets

talshorer:
Currently, when pxe_udp is active, it continues to enqueue packets with no upper bound, which could lead to consuming all available memory if the queue is not consumed fast enough (or at all) while under heavy traffic. Once all memory is consumed by this queue, drivers are unable to allocate rx buffers for new packets, rendering all other sockets useless.

Place an upper bound on the length of this queue to limit the amount of memory it can consume.

Signed-off-by: Tal Shorer <shorer@amazon.com>
Reviewed-by: Itamar Katz <itamark@amazon.com>

### DIFF
--- a/src/arch/x86/interface/pxe/pxe_udp.c
+++ b/src/arch/x86/interface/pxe/pxe_udp.c
@@ -60,7 +60,12 @@ struct pxe_udp_connection {
 	struct sockaddr_in local;
 	/** List of received packets */
 	struct list_head list;
+	/** Current number of packets in list */
+	unsigned int npackets;
 };
+
+/** Number of packets to keep before dropping */
+#define PXE_UDP_QUEUE_LENGTH 48
 
 /**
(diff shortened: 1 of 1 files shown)


stappersg (approved):
Looks good to me.

talshorer:
@mcb30 Can you please review this PR? Thanks.

mcb30:
@talshorer Rather than a fixed 48-packet limit, a more generic approach would be to use a cache discarder as is done for the TCP receive queue.  This will be invoked whenever `malloc()` is unable to allocate a memory block.  I've implemented this in PR #859.  Could you please test and confirm whether or not that PR also solves your observed problem?

Thanks.

mcb30:
@talshorer I've tested and merged #859.  Please open a new issue if that does not solve the problem for you.  Thanks for highlighting the issue!

PR #770, opened 2022-10-10
[PATCH] [intel] Add PCI ID for I219-V and -LM 16,17

NiKiZe:


NiKiZe:
Only 0x1a1f was tested on Thinkpad 21AH0091MX, but only in EFI mode.
Windows driver used is the same e1d68x64.inf as previous NICs as well.

Non scientific observation was that the NII driver seemed slightly faster then iPXE native driver, I think I have noticed this on previous changes as well, but I still think it is better to have them in iPXE for USB boot or similar.

Experience also says that there is no reason to believe that there will be any issues in PCBIOS mode.

stappersg:
The _diff_ looks good, the _commit message_ should get the content from https://github.com/ipxe/ipxe/pull/770#issuecomment-1273648208.

chelmuth:
I was about to open a PR for e47175678e5f1600f9cfc844e73578782b64141a but discovered this issue. Thus, I can confirm that I219-LM (16) works reliably on Fujitsu LIFEBOOK U7412.

mcb30:
Merged: thanks @NiKiZe and @chelmuth

PR #798, opened 2022-11-08
[dhcp] Let DHCPNAK return to discovery state for request host only

AndersBlomdell:
Only consider DHCPNAK from the host we sent request to, this prohibits other hosts in a multi-dhcpd environment to spuriously return the state machine to discovery state.

### DIFF
--- a/src/net/udp/dhcp.c
+++ b/src/net/udp/dhcp.c
@@ -571,12 +571,6 @@ static void dhcp_request_rx ( struct dhcp_session *dhcp,
 	if ( peer->sin_port != htons ( BOOTPS_PORT ) )
 		return;
 
-	/* Handle DHCPNAK */
-	if ( msgtype == DHCPNAK ) {
-		dhcp_defer ( dhcp );
-		return;
-	}
-
 	/* Filter out unacceptable responses */
 	if ( msgtype /* BOOTP */ && ( msgtype != DHCPACK ) )
 		return;
@@ -584,6 +578,12 @@ static void dhcp_request_rx ( struct dhcp_session *dhcp,
 		return;
 	if ( ip.s_addr != dhcp->offer.s_addr )
 		return;
+
+	/* Handle DHCPNAK */
+	if ( msgtype == DHCPNAK ) {
+		dhcp_defer ( dhcp );
+		return;
+	}
 
 	/* Record assigned address */
 	dhcp->local.sin_addr = ip;
--- a/src/net/udp/dhcp.c
+++ b/src/net/udp/dhcp.c
@@ -571,6 +571,16 @@ static void dhcp_request_rx ( struct dhcp_session *dhcp,
 	if ( peer->sin_port != htons ( BOOTPS_PORT ) )
 		return;
 
+	/* Handle DHCPNAK */
+	if ( msgtype == DHCPNAK &&
+             server_id.s_addr == dhcp->server.s_addr &&
+             ip.s_addr == dhcp->offer.s_addr ) {
+                /* Only consider NAK's from the server we sent the
+                   request to, and with the same offer */ 
+		dhcp_defer ( dhcp );
+		return;
+	}
+
 	/* Filter out unacceptable responses */
 	if ( msgtype /* BOOTP */ && ( msgtype != DHCPACK ) )
 		return;
@@ -578,12 +588,6 @@ static void dhcp_request_rx ( struct dhcp_session *dhcp,
 		return;
 	if ( ip.s_addr != dhcp->offer.s_addr )
 		return;
-
-	/* Handle DHCPNAK */
-	if ( msgtype == DHCPNAK ) {
-		dhcp_defer ( dhcp );
-		return;
-	}
 
 	/* Record assigned address */
 	dhcp->local.sin_addr = ip;


AndersBlomdell:
https://bugzilla.redhat.com/show_bug.cgi?id=2140729

stappersg:
> https://bugzilla.redhat.com/show_bug.cgi?id=2140729

Title that issue is _"Qemu machines refuses to PXE-boot"_.

NiKiZe:
Has there been any update to RFCs in regards to dual DHCP servers?
If the server isn't authorative for the segment, it shouldn't be sending NAKs in the first place?

AndersBlomdell:
I did a bisect to find the culprit. It landed me on https://github.com/ipxe/ipxe/commit/1192edf394aaebde666c6aafda1c8b782b08a845

NiKiZe:
> I did a bisect to find the culprit. It landed me on [1192edf](https://github.com/ipxe/ipxe/commit/1192edf394aaebde666c6aafda1c8b782b08a845)

Awesome, thanks! Then I think this PR is appropriate. @mcb30 ?

AndersBlomdell:
RFC2131 3.1 §3 first paragraph covers our use case; poor mans failover by splitting address-range over multiple servers, i.e. each handles their own chunk of  that range. Has served us well for the last 20 yeats.

stappersg:
It was [dhcproxy config discussion](https://github.com/ipxe/ipxe/discussions/873) that triggered me to raise attention for this DHCP related merge request.

mcb30:
> Only consider DHCPNAK from the host we sent request to, this prohibits other hosts in a multi-dhcpd environment to spuriously return the state machine to discovery state.

But isn't that "spurious" behaviour precisely how the "authoritative" mode for several common DHCP servers is supposed to work?  A DHCP server that is authoritative for a network segment sees a (broadcast) DHCPREQUEST with a different selected DHCP server, and therefore sends a DHCPNAK to cause the client to return to the discovery state.

Merging this PR would break that behaviour.

I think @NiKiZe is correct:

> If the server isn't authorative for the segment, it shouldn't be sending NAKs in the first place?

It sounds as though you have misconfigured DHCP servers, where both of them believe themselves to be authoritative for the network but with non-overlapping address ranges.  Whichever server the client selects, the other server will send a DHCPNAK to try to force it back to the discovery state.  As far as I can tell, iPXE is doing exactly the right thing here.

I'm happy to be corrected, if there is any reference for how clients are supposed to handle a DHCPNAK coming from a server that was not the client's selected server.  RFC2131 seems to leave it undefined.

AndersBlomdell:
> > Only consider DHCPNAK from the host we sent request to, this prohibits other hosts in a multi-dhcpd environment to spuriously return the state machine to discovery state.
> 
> But isn't that "spurious" behaviour precisely how the "authoritative" mode for several common DHCP servers is supposed to work? A DHCP server that is authoritative for a network segment sees a (broadcast) DHCPREQUEST with a different selected DHCP server, and therefore sends a DHCPNAK to cause the client to return to the discovery state.
> 
> Merging this PR would break that behaviour.
> 
> I think @NiKiZe is correct:
> 
> > If the server isn't authorative for the segment, it shouldn't be sending NAKs in the first place?
> 
> It sounds as though you have misconfigured DHCP servers, where both of them believe themselves to be authoritative for the network but with non-overlapping address ranges. Whichever server the client selects, the other server will send a DHCPNAK to try to force it back to the discovery state. As far as I can tell, iPXE is doing exactly the right thing here.

I'm interpreting RFC2131 §1.6 (bulletpoint 6) as more permissive: _A DHCP client must be prepared to receive multiple responses  to a request for configuration parameters.  Some installations  may include multiple, overlapping DHCP servers to enhance reliability and increase performance._

> 
> I'm happy to be corrected, if there is any reference for how clients are supposed to handle a DHCPNAK coming from a server that was not the client's selected server. RFC2131 seems to leave it undefined.

I guess that you refer to RFC2131 §3.1.5:  _If the client receives a DHCPNAK message, the client restarts the configuration process._, but RFC2134 §3.1.4 previously states:  _The combination of 'client identifier' or  'chaddr' and assigned network address constitute a unique  identifier for the client's lease and are used by both the client  and server to identify a lease referred to in any DHCP messages._, that was my rationale for moving the NAK handling after the check for these.

AndersBlomdell:
> > I did a bisect to find the culprit. It landed me on [1192edf](https://github.com/ipxe/ipxe/commit/1192edf394aaebde666c6aafda1c8b782b08a845)
> 
> Awesome, thanks! Then I think this PR is appropriate. @mcb30 ?

It also seems like @NiKiZe approved of the patch after he posted very relevant question.

AndersBlomdell:
Sorry for the .txt extension, `authoritative` not set in my configuration
[dhcpd.conf.txt](https://github.com/ipxe/ipxe/files/10571054/dhcpd.conf.txt)

mcb30:
@AndersBlomdell Interesting!  Could you please track down the behaviour of the ISC reference client `dhcpcd` to see how it handles a DHCPNAK?

AndersBlomdell:
Since our clients get their adresses (served by a number of other DHCP-servers), I´m pretty sure it ignores them, as all the PC BIOSes we have does when it comes to PXE booting.
 
Side note: RFC2131 figure 5 has no transition from state REQESTING on a NAK.

mcb30:
> Since our clients get their adresses (served by a number of other DHCP-servers), I´m pretty sure it ignores them, as all the PC BIOSes we have does when it comes to PXE booting.

"Works for me" is not enough, sorry.  I want either a specification reference (which seems not to exist) or, failing that, either documented behaviour or relevant source code from the ISC reference `dhcpcd` client.

> Side note: RFC2131 figure 5 has no transition from state REQESTING on a NAK.

Yes it does: there's a "DHCPNAK / Discard offer" line from REQUESTING straight back to INIT (i.e. the exact path that iPXE currently takes).

mcb30:
@AndersBlomdell I do understand that this is frustrating for you: you have a single network environment in which your simple fix works perfectly, and you don't see why we can't just apply that fix upstream.

Unfortunately, I also have to take into account everybody else's network environment, and be sure that this change isn't going to break other valid setups.

AndersBlomdell:
> > Side note: RFC2131 figure 5 has no transition from state REQESTING on a NAK.
> 
> Yes it does: there's a "DHCPNAK / Discard offer" line from REQUESTING straight back to INIT (i.e. the exact path that iPXE currently takes).

So I interpret it differently the as the transitioning
 
  **SELECTING**  -> `Select offer/send DHCPREQUEST ` -> **REQUESTING**   -> `DHCPNAK/Discard offer`

to only discard the `offer`  that was selected in the `DHCPREQUEST`.

AndersBlomdell:
> @AndersBlomdell I do understand that this is frustrating for you: you have a single network environment in which your simple fix works perfectly, and you don't see why we can't just apply that fix upstream.

I'm not going to file file a bug report to ISC on their DHCP-server following specifications for non-authoritative DHCP -servers (which RFC2131 has no notion of BTW).

I'll leave it at this, even though if means that I will have to patch our systems indefinitely. Commit https://github.com/ipxe/ipxe/commit/1192edf394aaebde666c6aafda1c8b782b08a845 broke what I consider sane behaviour, and whoever wrote that patch has to decide what is sane...

mcb30:
> to only discard the `offer` that was selected in the `DHCPREQUEST`.

Yes, only the selected offer gets discarded.  The question is not which offer gets discarded: the question is how to respond to a DHCPNAK coming from a server that did not send the selected offer.  (RFC2131 does not specify the behaviour, as far as we both can tell.)

> I'm not going to file file a bug report to ISC on their DHCP-server not following specifications for non-authoritative DHCP -servers (which RFC2131 has no notion of BTW).

Of course not.  That was never the ask, and I'm not sure where you picked up that idea from.

The ask was for you to take the time to find the point in the `dhcpcd` (not `dhcpd`) code that handles DHCPNAKs and see if its behaviour corresponds to what your patch implements (i.e. ignore DHCPNAK unless it comes from the selected server).

You're not prepared to put in the time to do this, which is fair enough, but that just means that I have to do it instead.

mcb30:
Looking more closely at your patch, you have also moved the check for `msgtype == DHCPNAK` until after the code:
```
	if ( msgtype /* BOOTP */ && ( msgtype != DHCPACK ) )
		return;
```

Your patch therefore just disables DHCPNAK handling completely.

AndersBlomdell:
Pushed new version (untested, don't have time to chck until next thursday at the earliest). Hopefully taking into account your relevant critic, and clarifying the log entry.

mcb30:
I did a quick test.  My `dhcpd` instance does not seem to ever fill in `yiaddr` when sending a DHCPNAK.  (This makes some kind of twisted sense, since DHCP is supposed to be in some sense backwards-compatible with BOOTP, and in BOOTP filling in `yiaddr` would count as successfully issuing a lease for the address.)

I also took a look through some other DHCP client implementations.  None seemed to verify the `yiaddr` field in a received DHCPNAK.  This makes sense, given the above.

The upshot of this is that we definitely can't verify `yiaddr` (i.e. `ip.s_addr`) to identify a valid DHCPNAK since it is likely to be empty.

This leaves open the possibility of verifying `server_id.s_addr` - I haven't yet reached a definitive answer on that.

AndersBlomdell:
Wireshark capture of four DHCP servers responding to one client, servers configured as from previous config file.

[dhcp.zip](https://github.com/ipxe/ipxe/files/10581806/dhcp.zip)

mcb30:
> Wireshark capture of four DHCP servers responding to one client, servers configured as from previous config file.
> 
> [dhcp.zip](https://github.com/ipxe/ipxe/files/10581806/dhcp.zip)

That matches what I am seeing: the DHCPNAK responses do not have a `yiaddr` field set.

mcb30:
Having trawled through the `dhcpd` documentation, I think the `authoritative` declaration is not relevant here.  Quoting from `dhcpd.conf(5)`:

> If the server finds the address the client is requesting, and  that  address  is  available  to  the
> client,  the server will send a DHCPACK.  If the address is no longer available, or the client isn't
> permitted to have it, the server will send a DHCPNAK.  If the server knows  nothing  about  the
> address,  it  will remain silent, unless the address is incorrect for the network segment to which the
> client has been attached and the server is authoritative for that network segment, in which case the
> server will send a DHCPNAK even though it doesn't know about the address.

With your configuration, I'm pretty sure the servers are going down the "client isn't permitted to have it" branch of the above logic.  From each server's point of view, the client is requesting an address that lies within the subnet but outside the range of allocatable addresses.  The client is not permitted to have this address, and so as per the documentation the server will send a DHCPNAK without even considering the `authoritative` setting.

mcb30:
> The upshot of this is that we definitely can't verify `yiaddr` (i.e. `ip.s_addr`) to identify a valid DHCPNAK since it is likely to be empty.
> 
> This leaves open the possibility of verifying `server_id.s_addr` - I haven't yet reached a definitive answer on that.

This, combined with the fact that there exist multiple online sources suggesting the kind of split-range setup that you have, seems to be leading towards the conclusion that the only viable solution is to ignore a DHCPNAK if it comes from a server that we didn't select.

This will inevitably break the ability of a legitimate DHCP server to prevent a rogue DHCP server from assigning an address. However, since iPXE won't wait around for a potential DHCPNAK once it has received what it thinks is a valid DHCPACK, this would always have been unreliable anyway.

mcb30:
I've created and merged #879 which fixes this properly, based on all the information provided and discovered.

@AndersBlomdell Could you please test when you are able to (presumably next Thursday) and open a new issue if it still doesn't work as you expect?

Thanks for the contributions.

PR #852, opened 2023-01-16
[ena] Always allocate AENQ queues

agraf:
We currently don't allocate Asynchronous Event Notification Queues because we don't actually care about any of the events that may come in.

However, without allocating AENQ queues, ena internal memory checks fail and iPXE fails to run on Graviton systems.  So let's allocate the queues and not use them. That way, all target platforms will be happy.

With this patch applied, I can successfully run iPXE on c6g instances.

Signed-off-by: Alexander Graf <graf@amazon.com>

### DIFF
--- a/src/drivers/net/ena.c
+++ b/src/drivers/net/ena.c
@@ -351,6 +351,90 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
 }
 
 /**
+ * Set async event notification queue config
+ *
+ * @v ena		ENA device
+ * @v enabled		Bitmask of the groups to enable
+ * @ret rc		Return status code
+ */
+static int ena_set_aenq_config ( struct ena_nic *ena, uint32_t enabled ) {
+	union ena_aq_req *req;
+	union ena_acq_rsp *rsp;
+	union ena_feature *feature;
+	int rc;
+
+	/* Construct request */
+	req = ena_admin_req ( ena );
+	req->header.opcode = ENA_SET_FEATURE;
+	req->set_feature.id = ENA_AENQ_CONFIG;
+	feature = &req->set_feature.feature;
+	feature->aenq.enabled = cpu_to_le32 ( enabled );
+
+	/* Issue request */
+	if ( ( rc = ena_admin ( ena, req, &rsp ) ) != 0 )
+		return rc;
+
+	return 0;
+}
+
+/**
+ * Create async event notification queue
+ *
+ * @v ena		ENA device
+ * @ret rc		Return status code
+ */
+static int ena_create_async ( struct ena_nic *ena ) {
+	size_t aenq_len = ( ENA_AENQ_COUNT * sizeof ( ena->aenq.evt[0] ) );
+	int rc;
+
+	/* Allocate async event notification queue */
+	ena->aenq.evt = malloc_phys ( aenq_len, aenq_len );
+	if ( ! ena->aenq.evt ) {
+		rc = -ENOMEM;
+		goto err_alloc_aenq;
+	}
+	memset ( ena->aenq.evt, 0, aenq_len );
+
+	/* Program queue address and capabilities */
+	ena_set_base ( ena, ENA_AENQ_BASE, ena->aenq.evt );
+	ena_set_caps ( ena, ENA_AENQ_CAPS, ENA_AENQ_COUNT,
+		       sizeof ( ena->aenq.evt[0] ) );
+
+	DBGC ( ena, "ENA %p AENQ [%08lx,%08lx)\n",
+	       ena, virt_to_phys ( ena->aenq.evt ),
+	       ( virt_to_phys ( ena->aenq.evt ) + aenq_len ) );
+
+	/* Disable all events */
+	if ( ( rc = ena_set_aenq_config ( ena, 0 ) ) != 0 )
+		goto err_set_aenq_config;
+
+	return 0;
+
+ err_set_aenq_config:
+	ena_clear_caps ( ena, ENA_AENQ_CAPS );
+	free_phys ( ena->aenq.evt, aenq_len );
+ err_alloc_aenq:
+	return rc;
+}
+
+/**
+ * Destroy async event notification queue
+ *
+ * @v ena		ENA device
+ */
+static void ena_destroy_async ( struct ena_nic *ena ) {
+	size_t aenq_len = ( ENA_AENQ_COUNT * sizeof ( ena->aenq.evt[0] ) );
+
+	/* Clear queue capabilities */
+	ena_clear_caps ( ena, ENA_AENQ_CAPS );
+	wmb();
+
+	/* Free queue */
+	free_phys ( ena->aenq.evt, aenq_len );
+	DBGC ( ena, "ENA %p AENQ destroyed\n", ena );
+}
+
+/**
  * Create submission queue
  *
  * @v ena		ENA device
@@ -1098,6 +1182,10 @@ static int ena_probe ( struct pci_device *pci ) {
 	if ( ( rc = ena_create_admin ( ena ) ) != 0 )
 		goto err_create_admin;
 
+	/* Create async event notification queue */
+	if ( ( rc = ena_create_async ( ena ) ) != 0 )
+		goto err_create_async;
+
 	/* Set host attributes */
 	if ( ( rc = ena_set_host_attributes ( ena ) ) != 0 )
 		goto err_set_host_attributes;
@@ -1121,6 +1209,8 @@ static int ena_probe ( struct pci_device *pci ) {
  err_register_netdev:
  err_get_device_attributes:
  err_set_host_attributes:
+	ena_destroy_async ( ena );
+ err_create_async:
 	ena_destroy_admin ( ena );
  err_create_admin:
 	ena_reset ( ena );
@@ -1147,6 +1237,9 @@ static void ena_remove ( struct pci_device *pci ) {
 
 	/* Unregister network device */
 	unregister_netdev ( netdev );
+
+	/* Destroy async event notification queue */
+	ena_destroy_async ( ena );
 
 	/* Destroy admin queues */
 	ena_destroy_admin ( ena );
--- a/src/drivers/net/ena.h
+++ b/src/drivers/net/ena.h
@@ -24,6 +24,9 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 /** Number of admin completion queue entries */
 #define ENA_ACQ_COUNT 2
 
+/** Number of async event notification queue entries */
+#define ENA_AENQ_COUNT 2
+
 /** Number of transmit queue entries */
 #define ENA_TX_COUNT 16
 
@@ -59,6 +62,12 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 
 /** Maximum time to wait for admin requests */
 #define ENA_ADMIN_MAX_WAIT_MS 5000
+
+/** Async event notification queue capabilities register */
+#define ENA_AENQ_CAPS 0x34
+
+/** Async event notification queue base address register */
+#define ENA_AENQ_BASE 0x38
 
 /** Device control register */
 #define ENA_CTRL 0x54
@@ -128,6 +137,17 @@ struct ena_device_attributes {
 	uint8_t reserved_b[2];
 	/** Maximum MTU */
 	uint32_t mtu;
+} __attribute__ (( packed ));
+
+/** Async event notification queue config */
+#define ENA_AENQ_CONFIG 26
+
+/** Async event notification queue config */
+struct ena_aenq_config {
+	/** Bitmask of supported AENQ groups (device -> host) */
+	uint32_t supported;
+	/** Bitmask of enabled AENQ groups (host -> device) */
+	uint32_t enabled;
 } __attribute__ (( packed ));
 
 /** Host attributes */
@@ -208,6 +228,8 @@ struct ena_host_info {
 union ena_feature {
 	/** Device attributes */
 	struct ena_device_attributes device;
+	/** Async event notification queue config */
+	struct ena_aenq_config aenq;
 	/** Host attributes */
 	struct ena_host_attributes host;
 };
@@ -506,6 +528,28 @@ struct ena_acq {
 	unsigned int phase;
 };
 
+/** Async event notification queue event */
+struct ena_aenq_event {
+	/** Type of event */
+	uint16_t group;
+	/** ID of event */
+	uint16_t syndrome;
+	/** Phase */
+	uint8_t flags;
+	/** Reserved */
+	uint8_t reserved[3];
+	/** Timestamp */
+	uint64_t timestamp;
+	/** Additional event data */
+	uint8_t data[48];
+} __attribute__ (( packed ));
+
+/** Async event notification queue */
+struct ena_aenq {
+	/** Events */
+	struct ena_aenq_event *evt;
+};
+
 /** Transmit submission queue entry */
 struct ena_tx_sqe {
 	/** Length */
@@ -702,6 +746,8 @@ struct ena_nic {
 	struct ena_aq aq;
 	/** Admin completion queue */
 	struct ena_acq acq;
+	/** Async event notification queue */
+	struct ena_aenq aenq;
 	/** Transmit queue */
 	struct ena_qp tx;
 	/** Receive queue */


src/drivers/net/ena.c
```diff
@@ -206,23 +207,39 @@ static int ena_create_admin ( struct ena_nic *ena ) {
...
 		       sizeof ( ena->acq.rsp[0] ) );
 	ena_set_base ( ena, ENA_AQ_BASE, ena->aq.req );
 	ena_set_caps ( ena, ENA_AQ_CAPS, ENA_AQ_COUNT,
 		       sizeof ( ena->aq.req[0] ) );
+	ena_set_base ( ena, ENA_AENQ_BASE, ena->aenq.events );
+	ena_set_caps ( ena, ENA_AENQ_CAPS, ENA_AENQ_COUNT,
+		       sizeof ( ena->aenq.events[0] ) );
 
-	DBGC ( ena, "ENA %p AQ [%08lx,%08lx) ACQ [%08lx,%08lx)\n",
+	DBGC ( ena, "ENA %p AQ [%08lx,%08lx) ACQ [%08lx,%08lx) AENQ [%08lx,%08lx)\n",
```
mcb30, line 229: Please wrap to 80 characters (reflowing onto the next line immediately before `ena, `).  Please also check that the printed debug message will (with typical values) fit within 80 characters and, if not, split to two debug messages.

src/drivers/net/ena.c
```diff
@@ -206,23 +207,39 @@ static int ena_create_admin ( struct ena_nic *ena ) {
...
 	       ena, virt_to_phys ( ena->aq.req ),
 	       ( virt_to_phys ( ena->aq.req ) + aq_len ),
 	       virt_to_phys ( ena->acq.rsp ),
-	       ( virt_to_phys ( ena->acq.rsp ) + acq_len ) );
+	       ( virt_to_phys ( ena->acq.rsp ) + acq_len ),
+	       virt_to_phys ( ena->aenq.events ),
+	       ( virt_to_phys ( ena->aenq.events ) + aenq_len ));
```
mcb30, line 235: Please include a whitespace at the end: `) );` rather than `));`

src/drivers/net/ena.c
```diff
@@ -206,23 +207,39 @@ static int ena_create_admin ( struct ena_nic *ena ) {
...
 	/* Program queue addresses and capabilities */
 	ena_set_base ( ena, ENA_ACQ_BASE, ena->acq.rsp );
 	ena_set_caps ( ena, ENA_ACQ_CAPS, ENA_ACQ_COUNT,
 		       sizeof ( ena->acq.rsp[0] ) );
 	ena_set_base ( ena, ENA_AQ_BASE, ena->aq.req );
 	ena_set_caps ( ena, ENA_AQ_CAPS, ENA_AQ_COUNT,
 		       sizeof ( ena->aq.req[0] ) );
+	ena_set_base ( ena, ENA_AENQ_BASE, ena->aenq.events );
+	ena_set_caps ( ena, ENA_AENQ_CAPS, ENA_AENQ_COUNT,
+		       sizeof ( ena->aenq.events[0] ) );
```
mcb30, line 227: Is the event queue supposed to be in place before the command queue is set up?

src/drivers/net/ena.c
```diff
@@ -206,23 +207,39 @@ static int ena_create_admin ( struct ena_nic *ena ) {
...
 	       virt_to_phys ( ena->acq.rsp ),
-	       ( virt_to_phys ( ena->acq.rsp ) + acq_len ) );
+	       ( virt_to_phys ( ena->acq.rsp ) + acq_len ),
+	       virt_to_phys ( ena->aenq.events ),
+	       ( virt_to_phys ( ena->aenq.events ) + aenq_len ));
 	return 0;
 
 	ena_clear_caps ( ena, ENA_AQ_CAPS );
 	ena_clear_caps ( ena, ENA_ACQ_CAPS );
+	ena_clear_caps ( ena, ENA_AENQ_CAPS );
```
mcb30, line 240: The order of `ena_clear_caps()` calls should mirror the order of `ena_set_caps()` calls above.

mcb30, line 240: The corresponding call to `ena_clear_caps()` in `ena_destroy_admin()` is missing.

src/drivers/net/ena.c
```diff
@@ -206,23 +207,39 @@ static int ena_create_admin ( struct ena_nic *ena ) {
 	}
 	memset ( ena->aq.req, 0, aq_len );
 
+	/* Allocate AENQ queue */
```
mcb30, line 210: "Allocate admin event queue" for consistency with the other comments?

src/drivers/net/ena.h
```diff
@@ -24,6 +24,9 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 /** Number of admin completion queue entries */
 #define ENA_ACQ_COUNT 2
 
+/** Number of async event notification queue entries */
```
mcb30, line 27: "Number of admin event queue entries", for consistency with the other comments?

src/drivers/net/ena.h
```diff
@@ -57,6 +60,11 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 /** Admin queue doorbell register */
 #define ENA_AQ_DB 0x2c
 
+#define ENA_AENQ_CAPS 0x34
+#define ENA_AENQ_BASE 0x38
+#define ENA_AENQ_HEAD_DB 0x40
+#define ENA_AENQ_TAIL 0x44
```
mcb30, line 66: Please move these after `ENA_ADMIN_MAX_WAIT_MS` (which logically belongs with `ENA_AQ_DB`)

mcb30, line 66: All of these register definitions need comments.

src/drivers/net/ena.h
```diff
@@ -506,6 +514,22 @@ struct ena_acq {
...
 
+struct ena_aenq_event {
+	uint16_t group;
+	uint16_t syndrom;
+	uint8_t flags;
+	uint8_t reserved[3];
+	uint32_t timestamp_low;
+	uint32_t timestamp_high;
+	uint32_t data[12];
+};
```
mcb30, line 525: The structure and every field within the structure need comments.

mcb30, line 525: The structure should be marked `__attribute__ (( packed ))`

src/drivers/net/ena.h
```diff
@@ -506,6 +514,22 @@ struct ena_acq {
...
+
+/** Async event notification queue */
```
mcb30, line 527: "Admin event queue" for consistency with other comments.

src/drivers/net/ena.h
```diff
@@ -506,6 +514,22 @@ struct ena_acq {
...
+struct ena_aenq {
+	/** Notifications */
```
mcb30, line 529: "Events" seems more natural, given the field name

src/drivers/net/ena.h
```diff
@@ -702,6 +726,8 @@ struct ena_nic {
 	struct ena_aq aq;
 	/** Admin completion queue */
 	struct ena_acq acq;
+	/** Async event notification queue */
```
mcb30, line 729: "Admin event queue", for consistency.

src/drivers/net/ena.h
```diff
@@ -506,6 +514,22 @@ struct ena_acq {
...
+	uint8_t reserved[3];
+	uint32_t timestamp_low;
+	uint32_t timestamp_high;
+	uint32_t data[12];
+};
+
+/** Async event notification queue */
+struct ena_aenq {
+	/** Notifications */
+	struct ena_aenq_event *events;
```
mcb30, line 530: Since the array fields within the other structures are `req` and `rsp`, both of which are abbreviated and neither of which are pluralised, `evt` would seem more consistent than `events`.

mcb30 (changes requested):
Thank you for the contribution!

I've added review comments to address the line-by-line code issues, most of which relate to naming or other consistency.

One major question: does `ena_poll()` needs to be extended to poll the event queue and to mark the events as consumed (even if the events themselves are ignored)?  What will happen if the event queue overflows?

The commit message should also have a comment as to why the Graviton version of ENA requires the existence of the AENQ, given that no other versions have done so.

src/drivers/net/ena.c
agraf: Setting up the command queue doesn't initialize it yet. That happens lazily on the first doorbell. So the order here does not matter.

src/drivers/net/ena.c
agraf: I'll put in "async event queue". This is not an admin queue :)

agraf:
> Thank you for the contribution!
> 
> I've added review comments to address the line-by-line code issues, most of which relate to naming or other consistency.

Thanks a bunch for the super fast thorough review!

> One major question: does `ena_poll()` need to be extended to poll the event queue and to mark the events as consumed (even if the events themselves are ignored)? What will happen if the event queue overflows?

When the queue overflows, we issue internal errors into logs, but the adapter should survive. That said, I'll give it a stab to implement event retrieval as well. That way we'll be more future proof in case that behavior changes.

> The commit message should also have a comment as to why the Graviton version of ENA requires the existence of the AENQ, given that no other versions have done so.

That's a fun one. Ena checks whether the AENQ queues are at a valid physical address. When AENQ is disabled, it tracks aenq address = 0, depth = 0. The check however still runs, even with depth = 0. And because on Graviton RAM starts at 1G, that check fails and then brings down the whole adapter. On x86 systems, there is always RAM at 0, so we don't run into issues.

agraf:
I pushed v2 that should hopefully address all comments. Instead of handling AENQ events I went with just disabling all of them :)

mcb30 (approved):

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
 	return rc;
 }
 
+/**
+ * Set async event notification queue config
+ *
+ * @v ena		ENA device
+ * @ret rc		Return status code
```
mcb30, line 357: Missing parameter description for `enabled_groups`

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+ */
+static int ena_set_aenq_config ( struct ena_nic *ena,
+				 uint32_t enabled_groups ) {
+	union ena_aq_req *req;
+	union ena_acq_rsp *rsp;
+	union ena_feature *feature;
+	int rc = 0;
```
mcb30, line 364: Please do not initialise `rc = 0`.  Initialising variables unnecessarily prevents the compiler from detecting situations in which a code path has genuinely forgotten to set a variable's value.

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+	req->header.opcode = ENA_SET_FEATURE;
+	req->set_feature.id = ENA_AENQ_CONFIG;
+	feature = &req->set_feature.feature;
+	feature->aenq.enabled_groups = cpu_to_le32 ( enabled_groups );
+
+	/* Issue request */
+	if ( ( rc = ena_admin ( ena, req, &rsp ) ) != 0 )
+		DBGC ( ena, "ENA %p failed to set AENQ config\n", ena );
+
+	return rc;
```
mcb30, line 377: As per existing code, this should be a `return rc` within the `( ( rc = ... ) != 0 )` failure path, and a `return 0` if the flow of execution successfully reaches the end of the function.

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+}
+
+
```
mcb30, line 380: Stray blank line

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+/**
+ * Create async queues
```
mcb30, line 382: "Create async event queue"?  Seems to be handling only a single queue, as far as I can tell.

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+	/* Allocate async event queue */
+	ena->aenq.evt = malloc_phys ( aenq_len, aenq_len );
+	if ( ! ena->aenq.evt ) {
+		rc = -ENOMEM;
+		goto err_alloc_aenq;
+	}
+	memset ( ena->aenq.evt, 0, aenq_len );
+
+	/* Program queue addresses and capabilities */
+	ena_set_base ( ena, ENA_AENQ_BASE, ena->aenq.evt );
```
mcb30, line 400: "address" singular

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+		       sizeof ( ena->aenq.evt[0] ) );
+
+	DBGC ( ena, "ENA %p AENQ [%08lx,%08lx)\n",
+	       ena, virt_to_phys ( ena->aenq.evt ),
+	       ( virt_to_phys ( ena->aenq.evt ) + aenq_len ) );
+
+	/* Disable all events. That way we don't need to handle the queue. */
+	rc = ena_set_aenq_config ( ena, 0 );
+	if ( !rc )
+		return 0;
```
mcb30, line 411: Please follow the general pattern:

```
if ( ( rc = try_thing() ) != 0 )
   goto err_thing;
```

followed by a standalone

```
return 0;
```

when the flow of execution successfully reaches the end of the function, followed by

```
    undo_thing();
err_thing:
```

(In this case, I'm guessing that there won't be any `undo_thing()`, since there is nothing to undo from an attempt to disable events.)

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+
+	ena_clear_caps ( ena, ENA_AENQ_CAPS );
+	free_phys ( ena->aenq.evt, aenq_len );
+ err_alloc_aenq:
+	return rc;
+}
+
+/**
+ * Destroy async queues
+ *
```
mcb30, line 421: "Destroy async event queue"

src/drivers/net/ena.c
```diff
@@ -350,6 +350,89 @@ static int ena_admin ( struct ena_nic *ena, union ena_aq_req *req,
...
+ */
+static void ena_destroy_async ( struct ena_nic *ena ) {
+	size_t aenq_len = ( ENA_AENQ_COUNT * sizeof ( ena->aenq.evt[0] ) );
+
+	/* Clear queue capabilities */
+	ena_clear_caps ( ena, ENA_AENQ_CAPS );
+	wmb();
+
+	/* Free queues */
+	free_phys ( ena->aenq.evt, aenq_len );
```
mcb30, line 432: "Free queue"

src/drivers/net/ena.h
```diff
@@ -506,6 +534,29 @@ struct ena_acq {
 	unsigned int phase;
 };
 
+/** Async event notification queue event */
+struct ena_aenq_event {
+	/** Type of event */
+	uint16_t group;
+	/** Id of event */
+	uint16_t syndrom;
```
mcb30, line 542: Missed this last time, sorry: is this a typo of "syndrome"?

src/drivers/net/ena.h
```diff
@@ -506,6 +534,29 @@ struct ena_acq {
...
+	/** Phase */
+	uint8_t flags;
+	uint8_t reserved[3];
```
mcb30, line 545: Missing comment `/** Reserved */`

mcb30 (changes requested):
Thanks for the quick v2!  One more round of comments and then I think we're good to go.

agraf:
Alright, let's see if v3 looks good :)

mcb30:
Thanks you!  :slightly_smiling_face:

I've made some minor changes:
* consistently using the phrase "async event notification queue", since that seemed to be the variant used most often in the code
* shortening `enabled_groups` and `supported_groups` to `enabled` and `supported`, since there are no other fields within the structure
* reordering structures in the header file to maintain numeric sort ordering of feature IDs
* using `if ( ( rc = ..... ) != 0 )` rather than the equivalent two-line form, for consistency

I've also changed the commit message.  Some experimentation on `c6g.medium` in `eu-west-2` indicated that it's not just having a valid physical address as the queue base: it will also fail if the queue count is left as zero.  I've therefore changed the commit message to state that "The ENA firmware found on Graviton instances requires the AENQ to exist, otherwise all admin queue commands will fail."

It occurs to me that there's a theoretical race condition: it's possible for the firmware to fill the AENQ with events before we get a chance to disable all events.  We can't disable all events before enabling the AENQ, since the act of disabling events requires a functional admin queue, and the admin queue won't function until the AENQ exists.  I don't think there's anything we can do about this.  I'm going to therefore presume that the firmware cannot complain if the AENQ fills up, since it gives us no way to guarantee that this will not happen.

The updated commit is at https://github.com/ipxe/ipxe/commit/6b977d125.  If you're happy to accept these changes, could you force-push your branch to that commit, and I'll then merge it in?

agraf:
Also yes, the race condition exists in theory, but in practice won't be a problem for iPXE. Even if we were to exceed the number of queue entries for AENQ, the adapter would still function correctly.

mcb30:
@agraf Thank you for all of your work on this.  I've merged it in and published updated AMIs to all regions.  The links at https://ipxe.org/howto/ec2 now lead to AMIs with this commit included.

Thanks again!

Issue #856, opened 2023-01-19
cross-compilation broken on host architectures not supported by EDK2


hramrach:
When cross-compiling the util/elf2efi.c is compiled with host compiler but includes the target-specific header src/include/ipxe/efi/ProcessorBind.h

When binary for different supported architecture is built on another supported architecture the host settings are applied potentially silently building wrong binaries.

When building on unsupported architecture such as powerpc or s390 no settings are applied and the build fails.

mcb30:
> When binary for different supported architecture is built on another supported architecture the host settings are applied potentially silently building wrong binaries.

Not actually a problem, since the only parts of the EDK2 headers that we are relying upon are those that define the PE format headers, and the layout of those does not vary by CPU architecture.  (It does vary between 32-bit and 64-bit architectures but we already handle that, hence the existence of `elf2efi32` and `elf2efi64`.)

> When building on unsupported architecture such as powerpc or s390 no settings are applied and the build fails.

This is tricky.  For any host CPU not supported by EDK2 there will not be a `ProcessorBind.h` that we could use.  I've created #857 which takes the alternative approach of using the host's `<stdint.h>` and defining the EFI integer types `UINT16` etc in terms of the standard types `uint16_t` etc.

hramrach:
Thanks for the fix. 

I came up with a hack using the arm64 include as fallback when nothing else applied but not using the arch-specific header in the first place is much better.

hramrach:
It still does not build on bigendian platforms but whatever

```
[  140s] x86_64-suse-linux-objdump -ht bin-i386-efi/ipxe.efi.tmp | perl ./util/sortobjdump.pl >> bin-i386-efi/ipxe.efi.tmp.map
[  140s] ./util/elf2efi32 --subsystem=10 bin-i386-efi/ipxe.efi.tmp bin-i386-efi/ipxe.efi
[  140s] ELF section header outside file in bin-i386-efi/ipxe.efi.tmp
[  140s] make: *** [Makefile.efi:34: bin-i386-efi/ipxe.efi] Error 1
[  140s] rm bin-i386-efi/version.ipxe.efi.o
[  140s] error: Bad exit status from /var/tmp/rpm-tmp.nz0LrT (%build)

```

mcb30:
> It still does not build on bigendian platforms but whatever

[...]

PR #857, opened 2023-01-20
[efi] Do not rely on ProcessorBind.h when building host binaries

mcb30:
We cannot rely on the EDK2 ProcessorBind.h headers when compiling a binary for execution on the build host itself (e.g. elf2efi), since the host's CPU architecture may not even be supported by EDK2.

Fix by skipping ProcessorBind.h when building a host binary, and defining the bare minimum required to allow other EDK2 headers to compile cleanly.

Fixes: #856 

Reported-by: Michal Suchánek <msuchanek@suse.de>

### DIFF
--- a/src/include/ipxe/efi/ProcessorBind.h
+++ b/src/include/ipxe/efi/ProcessorBind.h
@@ -10,20 +10,54 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
  *  - mcb30
  */
 
-#if __i386__
+#ifdef EFI_HOSTONLY
+
+/*
+ * We cannot rely on the EDK2 ProcessorBind.h headers when compiling a
+ * binary for execution on the build host itself, since the host's CPU
+ * architecture may not even be supported by EDK2.
+ */
+
+/* Define the basic integer types in terms of the host's <stdint.h> */
+#include <stdint.h>
+typedef int8_t INT8;
+typedef int16_t INT16;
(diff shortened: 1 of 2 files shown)


src/include/ipxe/efi/ProcessorBind.h
```diff
@@ -10,20 +10,54 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
...
 #include <ipxe/efi/Ia32/ProcessorBind.h>
 #endif
 
-#if __x86_64__
+#ifdef __x86_64__
 #include <ipxe/efi/X64/ProcessorBind.h>
 #endif
 
-#if __arm__
+#ifdef __arm__
```
hramrach, line 53: I think that the #if was intentional here, there used to be an Apple compiler that defined one of these to 0 for an unsupported architecture.

May not be very relevant anymore today.

mcb30: We have several other points within the codebase that use `#if defined(__arm__)`, so this change can't introduce any additional problems.  I changed it here so that there was a consistent use of `#ifdef` within `ProcessorBind.h` rather than a mix of `#ifdef` and `#if`.

(The alternative would have been to use an explicit `#define EFI_HOSTONLY 1` instead of just `#define EFI_HOSTONLY` in `elf2efi.c`, but that would itself be inconsistent with the way that `#define` flags are generally used in iPXE.)

PR #882, opened 2023-02-05
[lldp] Add support for the Link Layer Discovery Protocol

mcb30:
Add support for recording LLDP packets and exposing TLV values via the settings mechanism.  LLDP settings are encoded as

  `${netX.lldp/<prefix>.<type>.<index>.<offset>.<length>}`

where

  `<type>` is the TLV type

  `<offset>` is the starting offset within the TLV value

  `<length>` is the length (or zero to read the from `<offset>` to the end)

  `<prefix>`, if it has a non-zero value, is the subtype byte string of
  length <offset> to match at the start of the TLV value, up to a
  maximum matched length of 4 bytes

  `<index>` is the index of the entry matching `<type>` and `<prefix>` to be
  accessed, with zero indicating the first matching entry

The `<prefix>` is designed to accommodate both matching of the OUI within an organization-specific TLV (e.g. 0x0080c2 for IEEE 802.1 TLVs) and of a subtype byte as found within many TLVs.

This encoding allows most LLDP values to be extracted easily.  For example

  System name: `${netX.lldp/5.0.0.0:string}`

  System description: `${netX.lldp/6.0.0.0:string}`

  Port description: `${netX.lldp/4.0.0.0:string}`

  Port interface name: `${netX.lldp/5.2.0.1.0:string}`

  Chassis MAC address: `${netX.lldp/4.1.0.1.0:hex}`

  Management IPv4 address: `${netX.lldp/5.1.8.0.2.4:ipv4}`

  Port VLAN ID: `${netX.lldp/0x0080c2.1.127.0.4.2:int16}`

  Port VLAN name: `${netX.lldp/0x0080c2.3.127.0.7.0:string}`

  Maximum frame size: `${netX.lldp/0x00120f.4.127.0.4.2:uint16}`

Fixes: #413 

Originally-implemented-by: Marin Hannache <git@mareo.fr>

### DIFF
--- a/src/config/config_ethernet.c
+++ b/src/config/config_ethernet.c
@@ -49,3 +49,6 @@ REQUIRE_OBJECT ( eth_slow );
 #ifdef NET_PROTO_EAPOL
 REQUIRE_OBJECT ( eapol );
 #endif
+#ifdef NET_PROTO_LLDP
+REQUIRE_OBJECT ( lldp );
+#endif
--- a/src/config/general.h
+++ b/src/config/general.h
@@ -40,6 +40,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #define	NET_PROTO_STP		/* Spanning Tree protocol */
 #define	NET_PROTO_LACP		/* Link Aggregation control protocol */
 #define	NET_PROTO_EAPOL		/* EAP over LAN protocol */
+#undef	NET_PROTO_LLDP		/* Link Layer Discovery protocol */
 
 /*
  * PXE support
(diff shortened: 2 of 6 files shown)


NiKiZe:
Thank you!
Could we have it enabled by default in efi builds? ❤️

mcb30:
> Thank you! Could we have it enabled by default in efi builds? heart

Here you go: https://github.com/ipxe/ipxe/commit/7cc305f7b4fb3508e64afa8c4c54bee1fff1f405

antonym:
Huge! Thank you!

PR #907, opened 2023-03-02
Add workaround for I210 reset hardware bugs

BigSoupFan42:
Hello, this fixes a weird hardware bug with I210s that I encountered. Basically not all the registers return to their default values on reset, so depending on how an OS is using the device it may become unable to pass traffic. I encountered this by using the QAV features on the I210 and then failing to get a DHCP address on the next boot after a reset signal was asserted.

More than happy to add explanation, testing logs, change up code, etc.

Using the current master branch, using the bin-x86_64-efi/ipxe.efi target:
```
FS0:\> master_ipxe.efi
iPXE initialising devices...ok
iPXE 1.21.1+ (g96bb6) -- Open Source Network Boot Firmware -- https://ipxe.org
Features: DNS HTTP iSCSI TFTP VLAN SRP AoE EFI Menu
iPXE> dhcp net1
Configuring (net1 00:17:3c:02:d6:7e).................. No configuration methods succeeded (https://ipxe.org/040ee186)
iPXE> ifstat net1
net1: 00:17:3c:02:d6:7e using i210-2 on 0000:0f:00.0 (Ethernet) [closed]
[Link:up, TX:0 TXE:9 RX:107 RXE:20]
[TXE: 9 x "Operation canceled (https://ipxe.org/0b086095)"]
[RXE: 4 x "The socket is not connected (https://ipxe.org/380f6093)"]
[RXE: 16 x "Error 0x440e6083 (https://ipxe.org/440e6083)"]
iPXE> 
```

With this fix, same target:
```
FS0:\> my_ipxe.efi
iPXE initialising devices...ok
iPXE 1.21.1+ (g4366e) -- Open Source Network Boot Firmware -- https://ipxe.org
Features: DNS HTTP iSCSI TFTP VLAN SRP AoE EFI Menu
iPXE> dhcp net1
Configuring (net1 00:17:3c:02:d6:7e)...... ok
iPXE> ifstat net1
net1: 00:17:3c:02:d6:7e using i210-2 on 0000:0f:00.0 (Ethernet) [open]
[Link:up, TX:9 TXE:0 RX:57 RXE:26]
[RXE: 5 x "Error 0x440e6083 (https://ipxe.org/440e6083)"]
[RXE: 21 x "The socket is not connected (https://ipxe.org/380f6093)"]
```

### DIFF
--- a/src/drivers/net/intel.c
+++ b/src/drivers/net/intel.c
@@ -290,6 +290,18 @@ static int intel_reset ( struct intel_nic *intel ) {
 		       pba, readl ( intel->regs + INTEL_PBA ) );
 	}
 
+	/* The Intel I210's packet buffer size registers reset only on
+	 * power up.  If an operating system changes these but then
+	 * the computer recieves a reset signal without losing power,
+	 * the registers will stay the same (but be incompatible with
+	 * other register defaults), thus making the device unable to
+	 * pass traffic.
+	 */
+	if ( intel->flags & INTEL_PBSIZE_RST ) {
+		writel ( INTEL_RXPBS_I210, intel->regs + INTEL_RXPBS );
+		writel ( INTEL_TXPBS_I210, intel->regs + INTEL_TXPBS );
+	}
+
 	/* Always reset MAC.  Required to reset the TX and RX rings. */
 	writel ( ( ctrl | INTEL_CTRL_RST ), intel->regs + INTEL_CTRL );
 	mdelay ( INTEL_RESET_DELAY_MS );
@@ -1139,7 +1151,7 @@ static struct pci_device_id intel_nics[] = {
 	PCI_ROM ( 0x8086, 0x1525, "82567v-4", "82567V-4", 0 ),
 	PCI_ROM ( 0x8086, 0x1526, "82576-5", "82576", 0 ),
 	PCI_ROM ( 0x8086, 0x1527, "82580-f2", "82580 Fiber", 0 ),
-	PCI_ROM ( 0x8086, 0x1533, "i210", "I210", 0 ),
+	PCI_ROM ( 0x8086, 0x1533, "i210", "I210", INTEL_PBSIZE_RST ),
 	PCI_ROM ( 0x8086, 0x1539, "i211", "I211", 0 ),
 	PCI_ROM ( 0x8086, 0x153a, "i217lm", "I217-LM", INTEL_NO_PHY_RST ),
 	PCI_ROM ( 0x8086, 0x153b, "i217v", "I217-V", 0 ),
@@ -1147,7 +1159,7 @@ static struct pci_device_id intel_nics[] = {
 	PCI_ROM ( 0x8086, 0x155a, "i218lm", "I218-LM", INTEL_NO_PHY_RST ),
 	PCI_ROM ( 0x8086, 0x156f, "i219lm", "I219-LM", INTEL_I219 ),
 	PCI_ROM ( 0x8086, 0x1570, "i219v", "I219-V", INTEL_I219 ),
-	PCI_ROM ( 0x8086, 0x157b, "i210-2", "I210", 0 ),
+	PCI_ROM ( 0x8086, 0x157b, "i210-2", "I210", INTEL_PBSIZE_RST ),
 	PCI_ROM ( 0x8086, 0x15a0, "i218lm-2", "I218-LM", INTEL_NO_PHY_RST ),
 	PCI_ROM ( 0x8086, 0x15a1, "i218v-2", "I218-V", 0 ),
 	PCI_ROM ( 0x8086, 0x15a2, "i218lm-3", "I218-LM", INTEL_NO_PHY_RST ),
--- a/src/drivers/net/intel.h
+++ b/src/drivers/net/intel.h
@@ -138,6 +138,10 @@ struct intel_descriptor {
 /** Packet Buffer Size */
 #define INTEL_PBS 0x01008UL
 
+/** Receive packet buffer size */
+#define INTEL_RXPBS 0x02404UL
+#define INTEL_RXPBS_I210	0x000000a2UL	/**< I210 power-up default */
+
 /** Receive Descriptor register block */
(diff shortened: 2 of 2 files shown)


src/drivers/net/intel.h
```diff
@@ -319,6 +327,8 @@ enum intel_flags {
 	INTEL_NO_ASDE = 0x0008,
 	/** Reset may cause a complete device hang */
 	INTEL_RST_HANG = 0x0010,
+	/** Reset doesn't affect Packet Buffer Size registers */
+	INTEL_NO_PBSIZE_RST = 0x0020,
```
mcb30, line 331: We have the existing flag `INTEL_NO_PHY_RST` which effectively means "please do not perform a PHY reset in software".  This new flag effectively means "please do perform a PBSIZE reset in software".  The naming is therefore potentially confusing.

The most similar existing flag seems to be `INTEL_PBS_ERRATA` (which looks like a very similar bug for some of the ICH chips).

I would suggest changing this name to minimise potential confusion.

src/drivers/net/intel.h
```diff
@@ -200,6 +200,14 @@ struct intel_descriptor {
 #define INTEL_FEXTNVM11 0x05bbcUL
 #define INTEL_FEXTNVM11_WTF	0x00002000UL	/**< Don't ask */
 
+/** I210 Packet Buffer Sizes */
+#define INTEL_E1000_RXPBS	0x02404
+#define INTEL_E1000_TXPBS	0x03404
```
mcb30, line 205: These registers do not seem to exist in the E1000.  They do exist with the same name (but different bitfield allocations) in some E1000 variants such as i350.  Also, iPXE generally does not use `E1000` within its register definitions.

I would suggest using `INTEL_RXPBS` and `INTEL_TXPBS` here.

src/drivers/net/intel.c
```diff
@@ -674,6 +674,19 @@ static int intel_open ( struct net_device *netdev ) {
...
+	 * (but be incompatible with other register defaults), thus making the
+	 * device unable to pass traffic.
+	 */
+	if ( intel->flags & INTEL_NO_PBSIZE_RST ) {
+		writel ( INTEL_I210_RXPBSIZE_DEFAULT, intel->regs + INTEL_E1000_RXPBS );
+		writel ( INTEL_I210_TXPBSIZE_DEFAULT, intel->regs + INTEL_E1000_TXPBS );
+		writel ( INTEL_I210_DTXMXPKTSZ,
+		         intel->regs + INTEL_I210_DTXMXPKTSZ_DEFAULT );
+	}
+
```
mcb30, line 689: I think this logic probably belongs in `intel_reset()`, alongside the very similar logic for `INTEL_PBS_ERRATA`.  There is no need to reapply these register values on every device open, as far as I can tell.

src/drivers/net/intel.h
```diff
@@ -200,6 +200,14 @@ struct intel_descriptor {
...
+#define INTEL_I210_DTXMXPKTSZ	0x355C
```
mcb30, line 206: Same comment: `DTXMXPKTSZ` exists for some other E1000 variants, so let's just use `INTEL_DTXMXPKTSZ`.

src/drivers/net/intel.h
```diff
@@ -200,6 +200,14 @@ struct intel_descriptor {
...
+#define INTEL_I210_RXPBSIZE_DEFAULT	0x000000A2
+#define INTEL_I210_TXPBSIZE_DEFAULT	0x04000014
+#define INTEL_I210_DTXMXPKTSZ_DEFAULT	0x00000098
```
mcb30, line 209: Happy for these to retain `I210` within the name, since the values are I210-specific.

src/drivers/net/intel.c
mcb30, line 689: The commit log message states that the `DTXMXPKTSZ` register **does** get reset as expected by a software reset.  Do we need to include code to set its value?

mcb30 (changes requested):

src/drivers/net/intel.h
BigSoupFan42: I've changed it to `INTEL_SET_PBSIZE` to hopefully be more literal, although I am still open to suggestions there.

src/drivers/net/intel.h
BigSoupFan42: I've changed them to your suggestions in the most recent revision

src/drivers/net/intel.c
BigSoupFan42: Yes, you're right; I've moved it just under the `INTEL_PBS_ERRATA` block in the latest revision

BigSoupFan42: No, not necessarily. I just included it since the incompatibility between those three registers were what was causing issues. In the latest revision I've removed the `DTXMXPKTSZ` defines

src/drivers/net/intel.h
BigSoupFan42: Just circling back to this, would you prefer `INTEL_PBSIZE_RST`, as in "please reset the PBSIZE registers"? I initially didn't want to add `ERRATA` since the hardware bug is not listed as a errata, sighting or bug. The register chapter just says "these only reset on power up" at the beginning.

mcb30: Good point on it not being technically an errata.  I think `INTEL_PBSIZE_RST` with the comment "PBSIZE registers must be explicitly reset" or similar is a good call.

BigSoupFan42: I've swapped to `INTEL_PBSIZE_RST` and added your comment suggestion in the most recent revision

mcb30:
Thank you.  I've rebased and fixed some minor stylistic consistency issues (e.g. line lengths, ordering within the `intel.h` file by register address, etc).  This should be functionally unchanged from your code.  Are you happy for me to go ahead and merge this version?

mcb30 (approved):

BigSoupFan42:
Yep, looks great to me. I just re-tested and it still works

Issue #913, opened 2023-03-08
Multiple PXE options in dnsmasq (proxyDHCP) causes iPXE self-loading loop (instead of loading boot script)


mcb30:
### Discussed in https://github.com/ipxe/ipxe/discussions/909

<div type='discussions-op-text'>

<sup>Originally posted by **henryptung** March  4, 2023</sup>
I have a proxyDHCP setup using dnsmasq, where it provides both iPXE (via boot/snponly.efi) and grub2 as PXE options. I also have the iPXE script being loaded via a separate DHCP tag, as follows:

```
dhcp-match=set:iPXE,175

pxe-service=tag:!iPXE,x86PC,iPXE (legacy),boot/undionly.kpxe
pxe-service=tag:!iPXE,IA32_EFI,iPXE (EFI),boot/snponly.efi
pxe-service=tag:!iPXE,X86-64_EFI,iPXE (EFI),boot/snponly.efi
pxe-service=tag:!iPXE,BC_EFI,iPXE (EFI),boot/snponly.efi

pxe-service=tag:!iPXE,x86PC,GRUB 2 (i386-pc),boot/grub/i386-pc/core
pxe-service=tag:!iPXE,IA32_EFI,GRUB 2 (i386-efi),boot/grub/i386-efi/core.efi
pxe-service=tag:!iPXE,X86-64_EFI,GRUB 2 (x86_64-efi),boot/grub/x86_64-efi/core.efi
pxe-service=tag:!iPXE,BC_EFI,GRUB 2 (x86_64-efi),boot/grub/x86_64-efi/core.efi

dhcp-boot=tag:iPXE,boot/winpe.ipxe

log-dhcp
dhcp-no-override
dhcp-range=192.168.2.10,proxy
```

If I comment out the GRUB options and leave only a single PXE item, this works (i.e. iPXE loads, and then loads `boot/winpe.ipxe`. However, if I leave the GRUB options and choose iPXE from the PXE menu, iPXE instead detects `boot/snponly.efi` as its "boot script", and loops loading itself infinitely.

[captures.zip](https://github.com/ipxe/ipxe/files/10889771/captures.zip)

Note that in the attached packet captures, the DHCP options contain the correct script name in both cases - but for some reason, it is picked up in the one-option case, but ignored in the two-option case. As far as I can tell, the second (iPXE-driven) DHCP flow is identical in the two cases, so the behavior change seems to come from iPXE itself.

[...]

PR #914, opened 2023-03-08
[dhcp] Unregister ProxyDHCP and PXEBS settings on a successful DHCPACK

mcb30:
When a DHCP transaction does not result in the registration of a new "proxydhcp" or "pxebs" settings block, any existing settings blocks are currently left unaltered.

This can cause surprising behaviour.  For example: when chainloading iPXE, the "proxydhcp" and "pxebs" settings blocks may be prepopulated using cached values from the previous PXE bootloader.  If iPXE performs a subsequent DHCP request, then the DHCP or ProxyDHCP servers may choose to respond differently to iPXE.  The response may choose to omit the ProxyDHCP or PXEBS stages, in which case no new "proxydhcp" or "pxebs" settings blocks may be registered.  This will result in iPXE using a combination of both old and new DHCP responses.

Fix by assuming that a successful DHCPACK effectively acquires ownership of the "proxydhcp" and "pxebs" settings blocks, and that any existing settings blocks should therefore be unregistered.

Fixes: #913 

Reported-by: Henry Tung <htung@palantir.com>

### DIFF
--- a/src/net/udp/dhcp.c
+++ b/src/net/udp/dhcp.c
@@ -601,6 +601,12 @@ static void dhcp_request_rx ( struct dhcp_session *dhcp,
 		return;
 	}
 
+	/* Unregister any existing ProxyDHCP or PXEBS settings */
+	if ( ( settings = find_settings ( PROXYDHCP_SETTINGS_NAME ) ) != NULL )
+		unregister_settings ( settings );
+	if ( ( settings = find_settings ( PXEBS_SETTINGS_NAME ) ) != NULL )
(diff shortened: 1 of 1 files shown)


henryptung:
Tested the same proxy config (in a slightly different environment) and it seems to work properly with this fix.

I don't know if there should be any tests associated with the fix - think there are DNS-related tests but not DHCP-related tests in the repo? But agree with the description's analysis and the fix LGTM.

mcb30:
Thanks for testing.  I'll merge this in now.

PR #924, opened 2023-03-21
Correct memcpy issue resulting in no SNI for TLS connections

jmyoung:
Line 1199 in tls.c determines the size of the buffer to copy for the server name for TLS using `sizeof()`.  However `sizeof()` is determined at compile time, and the length of the buffer is determined at runtime, resulting in the length copied always being 0 bytes.

This pull request changes the memcpy to use the already existing `name_len` variable, which contains the length determined at runtime, resulting in the TLS server_name extension having the name in it as expected.

### DIFF
--- a/src/net/tls.c
+++ b/src/net/tls.c
@@ -1196,7 +1196,7 @@ static int tls_client_hello ( struct tls_connection *tls,
 	hello.extensions.server_name.list[0].len
 		= htons ( sizeof ( hello.extensions.server_name.list[0].name ));
 	memcpy ( hello.extensions.server_name.list[0].name, session->name,
-		 sizeof ( hello.extensions.server_name.list[0].name ) );
+		 name_len );
 	hello.extensions.max_fragment_length_type
 		= htons ( TLS_MAX_FRAGMENT_LENGTH );
 	hello.extensions.max_fragment_length_len


mcb30:
As per the `gcc` documentation at https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html:
> The length of an array is computed once when the storage is allocated and is remembered for the scope of the array in case you access it with `sizeof`.

There are multiple places in iPXE that rely on `sizeof` working on a variable length array.  It's a compiler extension that allows for simpler and safer code, since it allows the length of a structure to be expressed only once and so eliminates potential inconsistencies.

Do you have an example of a compiler actually producing code that results in a zero-length copy here?

jmyoung:
> Do you have an example of a compiler actually producing code that results in a zero-length copy here?

Yes.  `gcc version 11.3.0 (Ubuntu 11.3.0-1ubuntu1~22.04)`, compiling snponly64.efi produces a binary that outputs no name_extension in the TLS handshake.

With the fix I applied there, it does output a name_extension in the TLS handshake.  Note that the fix uses `name_len`, which was defined earlier using strlen, and that value is used when defining the length of that variable array in the struct anyway.

mcb30:
For reference, here's the wireshark dissection of the TLS Client Hello packet that I get when building with gcc 11.3.0-1ubuntu1~22.04:

```
Frame 37: 221 bytes on wire (1768 bits), 221 bytes captured (1768 bits)
Ethernet II, Src: RealtekU_12:34:56 (52:54:00:12:34:56), Dst: 26:f5:a2:c0:35:58 (26:f5:a2:c0:35:58)
Internet Protocol Version 4, Src: 10.0.0.139, Dst: 142.250.180.14
Transmission Control Protocol, Src Port: 49272 (49272), Dst Port: https (443), Seq: 1, Ack: 1, Len: 155
Transport Layer Security
    TLSv1.2 Record Layer: Handshake Protocol: Client Hello
        Content Type: Handshake (22)
        Version: TLS 1.2 (0x0303)
        Length: 150
        Handshake Protocol: Client Hello
            Handshake Type: Client Hello (1)
            Length: 146
            Version: TLS 1.2 (0x0303)
            Random: 2b361a64f4c16b48ffc1e48261f51f018fb9818224e54787b09d878e0ff329ba
            Session ID Length: 32
            Session ID: 2b361a64f4c16b48ffc1e48261f51f018fb9818224e54787b09d878e0ff329ba
            Cipher Suites Length: 24
            Cipher Suites (12 suites)
            Compression Methods Length: 1
            Compression Methods (1 method)
            Extensions Length: 49
            Extension: server_name (len=15)
                Type: server_name (0)
                Length: 15
                Server Name Indication extension
                    Server Name list length: 13
                    Server Name Type: host_name (0)
                    Server Name length: 10
                    Server Name: google.com
            Extension: max_fragment_length (len=1)
                Type: max_fragment_length (1)
                Length: 1
                Maximum Fragment Length: 4096 (4)
            Extension: signature_algorithms (len=12)
                Type: signature_algorithms (13)
                Length: 12
                Signature Hash Algorithms Length: 10
                Signature Hash Algorithms (5 algorithms)
            Extension: renegotiation_info (len=1)
                Type: renegotiation_info (65281)
                Length: 1
                Renegotiation Info extension
            Extension: session_ticket (len=0)
                Type: session_ticket (35)
                Length: 0
                Data (0 bytes)
            [JA3 Fullstring: 771,158-159-103-107-51-57-156-157-60-61-47-53,0-1-13-65281-35,,]
            [JA3: 2083b9d6b38ae84dad89dcf2d0feb318]
```
The server name extension is very definitely present and correct :slightly_smiling_face:

jmyoung:
I'm in the process of regenerating the problem now using your master and my patch, I'll send you through the wiresharks when I have them.  I have older wiresharks, but since they were generated beforehand, I'd rather get completely fresh ones to demonstrate this.  Hopefully our timezones will continue to line up.

jmyoung:
Well, I think you're right.  I'm not able to duplicate the issue with the current master, however I _can_ duplicate the issue with the iPXE that VMware supplies as part of AutoDeploy (that's my original issue).  I suspect what's happened here is that the captures I was using to see what was going on got sent through twice, so I was looking at what I thought was a capture from the iPXE master, whereas it was actually a capture from the VMware iPXE instead, hence me starting to look at the SNI code.

Obviously that's not your problem.  The current master appears to output SNI as one would expect without problem, and the packets I get from your master and from my pull request have identical SNI extension data and lengths etc.

At any rate, I'll follow up with VMware about it and see what they have to say, since it seems to be quite specifically with the fork that they distribute.

mcb30:
That's good to know: thanks for reporting back.

PR #963, opened 2023-06-02
Create PKCS#1 formatted private key when using openssl 3.0

jpy794:
Since openssl 3.0, `openssl-rsa` will create PKCS#8 formatted private key by default. But ipxe seems to only support PKCS#1. Pass a `-traditional` flag to `openssl-rsa` will fix this issue.

see https://www.openssl.org/docs/man3.0/man1/openssl-rsa.html

### DIFF
--- a/src/Makefile.housekeeping
+++ b/src/Makefile.housekeeping
@@ -686,9 +686,14 @@ VERYCLEANUP += $(PRIVKEY_LIST)
 #
 PRIVKEY_INC := $(BIN)/.private_key.der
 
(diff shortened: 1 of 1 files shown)


mcb30:
I would prefer to add the support for PKCS#8 so that it does not matter which version of OpenSSL is used.  This will also allow PKCS#8 keys to be set dynamically at runtime via the [`privkey`](https://ipxe.org/cfg/privkey) setting.

I have implemented PKCS#8 support in #964

mcb30:
Closed in favour of #964

PR #969, opened 2023-06-09
[efi] Provide read-only access to EFI variables via settings mechanism

mcb30:
EFI variables do not map neatly to the iPXE settings mechanism, since the EFI variable identifier includes a namespace GUID that cannot cleanly be supplied as part of a setting name.  Creating a new EFI variable requires the variable's attributes to be specified, which does not fit within iPXE's settings concept.

However, EFI variable names are generally unique even without the namespace GUID, and EFI does provide a mechanism to iterate over all existent variables.  We can therefore provide read-only access to EFI variables by comparing only the names and ignoring the namespace GUIDs.

Provide an "efi" settings block that implements this mechanism. Settings are returned as raw binary values by default since an EFI variable may contain boolean flags, integer values, ASCII strings, UCS-2 strings, EFI device paths, X.509 certificates, or any other arbitrary blob of data.

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -355,6 +355,9 @@ REQUIRE_OBJECT ( vram_settings );
 #ifdef ACPI_SETTINGS
 REQUIRE_OBJECT ( acpi_settings );
 #endif
+#ifdef EFI_SETTINGS
+REQUIRE_OBJECT ( efi_settings );
+#endif
 
 /*
  * Drag in selected keyboard map
--- a/src/config/defaults/efi.h
+++ b/src/config/defaults/efi.h
@@ -48,6 +48,8 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 
 #define	REBOOT_CMD		/* Reboot command */
 
+#define EFI_SETTINGS		/* EFI variable settings */
+
 #if defined ( __i386__ ) || defined ( __x86_64__ )
 #define IOAPI_X86
 #define NAP_EFIX86
--- a/src/config/settings.h
+++ b/src/config/settings.h
@@ -9,6 +9,8 @@
 
 FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 
+#include <config/defaults.h>
+
 #define	PCI_SETTINGS	/* PCI device settings */
 //#define	CPUID_SETTINGS	/* CPUID settings */
 //#define	MEMMAP_SETTINGS	/* Memory map settings */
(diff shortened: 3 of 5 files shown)


miczyg1:
This looks very promising. What would be most interesting to see is the write access. From what I see it ought to be as "simple" as adding `store` operation. Unless I am missing something what we would need is:

1. Own iPXE EFI varstore GUID under which iPXE would store the settings. As GUID stands for globally unique, there should be no major problem with generating one for iPXE and worrying about duplicates
2. Implementing `store` operation that would use the Runtime Services to write the variables

Anything else? I am willing to contribute the above implementation, if the approach is correct.

@mcb30 @Cornelicorn @Hammarskjold

mcb30:
> This looks very promising. What would be most interesting to see is the write access. From what I see it ought to be as "simple" as adding `store` operation. Unless I am missing something what we would need is:
> 
>     1. Own iPXE EFI varstore GUID under which iPXE would store the settings. As GUID stands for globally unique, there should be no major problem with generating one for iPXE and worrying about duplicates
> 
>     2. Implementing `store` operation that would use the Runtime Services to write the variables

@miczyg1 This needs some thought to avoid creating surprising effects.  For example, an attempt to set an existing EFI variable outside of the iPXE GUID namespace would be expected to modify that existing EFI variable, not create a new variable in the "wrong" namespace.

What is your use case for writing variables to an iPXE-specific GUID namespace (where they will be ignored by all other software)?  Is it just to allow for some information to be persisted between iPXE invocations across reboots?

miczyg1:
@mcb30 I thought about iPXE GUID namespace as an exclusive namespace for iPXE. The EFI settings backend shall only modify variables in this namespace.

The use case is to read/write URL(s) to chain from in iPXE scripts. I would like the URL to be persistent between boots, which is AFAIk not possible with currently available settings backends, right?

mcb30:
> The EFI settings backend shall only modify/access variables in this namespace.

If taken literally, this would break the existing feature, which provides read access to variables in any namespace.

> The use case is to read/write URL(s) to chain from in iPXE scripts. I would like the URL to be persistent between boots, which is AFAIK not possible with currently available settings backends, right?

Correct: there is no existing all-purpose settings backend that will allow for that.  There are a few NIC-specific settings backends that can preserve settings within flash storage on the NIC itself, but those are rare.

I think there is value in the feature idea.  It would need to be thought through carefully to avoid surprising outcomes.

NiKiZe:
Any chance we can consider an official proof of concept branch that can be tested before going main/master?

miczyg1:
> > The EFI settings backend shall only modify/access variables in this namespace.
> 
> If taken literally, this would break the existing feature, which provides read access to variables in any namespace.

I see. Well, we may do something like that for example:

`set efi./<namespace_guid>/<setting_name> <value>`

I.e. have the EFI settings backend populate multiple instances of the EFI settings, one per GUID namespace, in similar fashion like iPXE populates netN settings instances (one per NIC). That way we could preserve the original functionality. I don't know how incompatible it may be with the settings API though.

> I think there is value in the feature idea. It would need to be thought through carefully to avoid surprising outcomes.

It is easy to abuse the use of EFI variables, if that's what is on your mind. Or do you foresee other problems?
I wonder if the `store` operation should only allow a predefined set of setting names to be settable, like a set or subset from here: https://ipxe.org/cfg

PR #970, opened 2023-06-12
[rv64] Import RISC-V 64 ProcessorBind.h from EDK2 headers

andreas-schwab:
Signed-off-by: Andreas Schwab <schwab@suse.de>


stappersg:
Looks good to me, looks like #820

mcb30:
> After commit [a99e435](https://github.com/ipxe/ipxe/commit/a99e435c8e24887ce80c322029ba23103e00d1c2) this is no longer needed for my purpose, which is cross compiling the ipxe firmware on RISC-V as part of QEmu.

Closing since this is no longer needed, at least until iPXE gains support for building as a RISC-V binary.

PR #982, opened 2023-06-29
[interface] Initialize temporary interface in intf_shutdown()

vfazio:
Initialize the temporary interface to a null interface.

This prevents GPFs due to dereferencing garbage in intf_close() and intf_unplug() when interface.c is built with debug messages enabled.

### DIFF
--- a/src/core/interface.c
+++ b/src/core/interface.c
@@ -276,7 +276,7 @@ void intf_close ( struct interface *intf, int rc ) {
  * unplugs the interface.
  */
 void intf_shutdown ( struct interface *intf, int rc ) {
-	struct interface tmp;
+	struct interface tmp = INTF_INIT ( null_intf_desc );
 
 	DBGC ( INTF_COL ( intf ), "INTF " INTF_FMT " shutting down (%s)\n",
 	       INTF_DBG ( intf ), strerror ( rc ) );


vfazio:
This is an example of the exception raised when debug is enabled but the temporary interface is not properly initialized:

```
iPXE> ifconf -c dhcp && echo SUCCESS || echo FAILURE (${errno})
NETDEV net0 opening
NETDEV net0 link is up
INTF 0x5c65b658+18 shutting down (Operation canceled (https://ipxe.org/0b086095))
!!!! X64 Exception Type - 0D(#GP - General Protection)  CPU Apic ID - 00000000 !!!!
ExceptionData - 0000000000000000
RIP  - 000000005C62D96E, CS  - 0000000000000038, RFLAGS - 0000000000010246
RAX  - 000000005A5FBD58, RCX - 000000000000006D, RDX - 5DE90691FBA95D00
RBX  - 000000005A5FBD58, RSP - 000000005A5FBD08, RBP - 000000005C6562E0
RSI  - 000000005C62DB93, RDI - 000000005A5FBD58
R8   - 000000006AEA5420, R9  - 0000000000000000, R10 - 000000000000006D
R11  - 000000005A5FBB40, R12 - 00000000F4F79F6B, R13 - 000000005C62D8CD
R14  - 000000005C6562E0, R15 - 000000005A5FBE30
DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
GS   - 0000000000000030, SS  - 0000000000000030
CR0  - 0000000080010013, CR2 - 0000000000000000, CR3 - 000000005A201000
CR4  - 0000000000000668, CR8 - 0000000000000000
DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 0000000000000400
GDTR - 000000006B5BB000 0000000000000047, LDTR - 0000000000000000
IDTR - 0000000058A30018 0000000000000FFF,   TR - 0000000000000000
FXSAVE_STATE - 000000005A5FB960
!!!! Can't find image information. !!!!

```

mcb30:
This debug-message bugfix has made me realise that the debug message **values** will be incorrect for any use of a temporary interface (as used in `intf_shutdown()` and `xfer_vredirect()`) since the debug messages rely on the interface struct being present within the originating object, and this assumption is no longer valid since the commits that first introduced temporary interfaces located on the stack.

The issue affects only the debug messages, since nothing else ever sees the pointer to the temporary interface.  It is undesirable to add code that will always be present merely to allow for a very rarely used debug option.

I am not sure what the best solution to this problem may be.

vfazio:
> This debug-message bugfix has made me realise that the debug message **values** will be incorrect for any use of a temporary interface (as used in `intf_shutdown()` and `xfer_vredirect()`) since the debug messages rely on the interface struct being present within the originating object, and this assumption is no longer valid since the commits that first introduced temporary interfaces located on the stack.
> 
> The issue affects only the debug messages, since nothing else ever sees the pointer to the temporary interface. It is undesirable to add code that will always be present merely to allow for a very rarely used debug option.
> 
> I am not sure what the best solution to this problem may be.

I realize it's probably not ideal, but it does at least address a potential crash and save someone time from having to debug the issue. Even if the values in some of the debug messages themselves are not of meaning, them being emitted at least shows code path execution.

Maybe this is a temporary fix until a better over-arching solution can be determined.

I'd just prefer to not have to maintain this bugfix on a private fork if possible, but open to suggestions on a better path forward.

mcb30:
I've created #989 (untested) as an alternative approach

vfazio:
> I've created #989 (untested) as an alternative approach

Yea, that's a better fix, shall I close this PR? Or do you want to test it first? I can try to test it tomorrow morning

PR #992, opened 2023-07-10
[build] Allow for building EFI ISOs without syslinux

mcb30:
The isohybrid tool (that creates a partition table to describe the ESP embedded within the ISO image) relies upon the existence of the x86-only isolinux.bin file.  This file may not be available on non-x86 build hosts.

We already allow for the possibility that isohybrid may not be present on the build host, in which case we fall back to creating a non-hybrid ISO image.  Make a similar allowance for the possibility that the isolinux.bin file may not be present: require its existence only if we are attempting to build a BIOS-bootable image, or if we will be attempting to use isohybrid to create a hybrid ISO image.

Fixes: #986 

Reported-by: Xiaotian Wu <wuxiaotian@loongson.cn>

### DIFF
--- a/src/util/genfsimg
+++ b/src/util/genfsimg
@@ -229,17 +229,27 @@ done
 
 # Configure ISO image, if applicable
 #
-# Note that the BIOS boot files are required even for an EFI-only ISO,
-# since isohybrid will refuse to work without them.
+# Note that the BIOS boot files are required even for an EFI-only
+# hybrid ISO, since isohybrid will refuse to work without them.
 #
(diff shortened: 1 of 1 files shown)


yetist:
Thanks.

On the LoongArch platform, `make bin-loong64-efi/ipxe.iso` works fine, but `make bin-loong64-efi/ipxe.usb`, I get the error:
`util/genfsimg: line 286: /usr/bin/syslinux: No such file or directory`
Is it possible to fix this bug? Thanks.

yetist:
I think this PR can be merged and it does what it promises.

Issue #997, opened 2023-07-14
make target bin-x86_64-efi/ipxe.usb fails with Error: operand size mismatch for `push'


jstncno:
Hello,

I have been building ipxe from source using `make bin-x86_64-efi/ipxe.usb`. Lately I've been receiving the following errors:

```
$ make bin-x86_64-efi/ipxe.usb
  ...
  [BUILD] bin-x86_64-efi/bootsector.o
arch/x86/image/bootsector.c: Assembler messages:
arch/x86/image/bootsector.c:85: Error: operand size mismatch for `push'
make: *** [Makefile.housekeeping:964: bin-x86_64-efi/bootsector.o] Error 1
```

Could anybody help me troubleshoot this issue? Thanks

NiKiZe:
If this has worked previously, please use instructions at https://ipxe.org/howto/bisect to identify commits introducing issues.
Also make sure you don't have any local modifications.

If you can't find a working commit at all, please make sure to include versions of your toolchain, especially gcc.

boot.ipxe.org produces builds every hour, so there shouldn't be anything that could break for everyone being commited to the tree.

jstncno:
Thanks @NiKiZe for the quick response. It seems it has to do with my toolchain as I was able to successfully build in a vanilla Ubuntu environment.

My working env has gcc v10.2.1 while the non-working env has gcc v10.4.0. Is there a way to check for the correct toolchain?

mcb30:
This looks like a gcc bug.  There are precisely three `push` instructions within that inline assembly (including the implicit `push $1f` created by the `REAL_CODE()` macro), and none of them are problematic.

From your description of this as being fixed by using a "vanilla Ubuntu environment", I'm guessing that you may have a locally built `gcc` in your original environment, in which case that is most likely to be the source of the problem.

Closing: please reopen if you are able to reproduce the error in some kind of "standard" environment.

jstncno:
@mcb30 thanks for your response.

With some help from some colleagues, we were able to determine the root cause.

It seems that this is likely a change in the `binutils` package (specifically the assembler, `as`) rather than `gcc`.

In my working environment (I mistakenly mentioned "vanilla Ubuntu" instead of "vanilla Debian"), running `as --version` shows version `2.35.2`:
```
GNU assembler (GNU Binutils for Debian) 2.35.2
Copyright (C) 2020 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
[...]

PR #1011, opened 2023-08-03
Use the right sized register for the push operand

jstncno:
`push` operand based on the size of the value being pushed

Fixes https://github.com/ipxe/ipxe/issues/997, which is caused by `as` version 2.40+


eworm-de:
Why was this closed?

mcb30:
As per https://github.com/ipxe/ipxe/issues/997#issuecomment-1705065435 I am unable to reproduce the problem described.

mcb30:
Modified and merged as #1036

PR #1024, opened 2023-08-15
TPM presence information

mr-const:
This PR provides built-in setting uint8_t variable with TPM presence information, 0 if none and 1 if it's present and working.


NiKiZe:
I like this, but that is just an opinion from a long time iPXE user.

I can however provide some input on iPXE commits and PRs
Please rebase and forcepush to include any fixes in original commit.
Ensure correct whitespace.

See https://ipxe.org/admin#commit_log_messages for commit message guidelines.

This does not guarantee a merge, but it does reduce the work needed, and thus the likelihood of it being merged.

src/include/ipxe/errfile.h
stappersg, line 1: `s/ERRFILE_tpm/ERRFILE_efi_tpm/`

PR #1028, opened 2023-08-18
virtio-pci.c: Fix implementation of vpm_ioread32()

AlexanderEichner:
vpm_ioread32() only reads 16-bit of data due to a copy and paste bug (using readw/inw like vpm_ioread16() instead of readl/inl). This causes the virtio-net driver of ipxe to fail with the virtio-net device emulation of VirtualBox because it is quite strict regarding register access width and alignment which must be the natural register width.


mcb30:
The code change looks good to me.  I'll fix up the commit log message format.

PR #1082, opened 2023-11-19
[build] Use a deterministic serial number in sdsk

bmwiedemann:
The new deterministic serial number is based on the content hash.

Without this patch, `ipxe.sdsk` varied in the random 4-byte serial number stored by `mformat` at offset 0x27.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

### DIFF
--- a/src/util/genfsimg
+++ b/src/util/genfsimg
@@ -269,6 +269,7 @@ if [ -n "${FATIMG}" ] ; then
 	FATSIZE=$(( FATCYLS * 504 ))
 	FATARGS="-s 63 -h 16 -t ${FATCYLS}"
     fi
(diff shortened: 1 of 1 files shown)


stappersg (commented):
Looks good to me and https://reproducible-builds.org/ is a good thing.

mcb30:
I think this patch would unfortunately break building on FreeBSD, where `sha1sum` is (AFAIK) just `sha1`.

I've pushed https://github.com/ipxe/ipxe/commit/337880dea as an alternative approach, which uses `SOURCE_DATE_EPOCH` directly (as is done elsewhere in `genfsimg`) to derive a volume serial number.  Thanks for catching this!

PR #1117, opened 2024-01-08
Fix .gitignore files so bin and local dirs are not lost

markrogalski:
I ran into a problem after I cloned ipxe, then committed that tree to another repo, iPXE would not build in the new repo. I found it was because the bin and config/local directories were not committed in the new repo due to .gitignore rules. This PR updates the .gitignore rules so the required directories are not excluded when committing the tree.

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,4 +1,4 @@
 .toolcheck
 .echocheck
 TAGS*
-bin*
+bin-*
--- a/src/bin/.gitignore
+++ b/src/bin/.gitignore
@@ -1,1 +1,2 @@
 *
+!.gitignore
--- a/src/config/local/.gitignore
+++ b/src/config/local/.gitignore
@@ -1,1 +1,2 @@
 *
+!.gitignore


mcb30:
I'm not sure what problem this is trying to solve.  Both the `bin` and `config/local` directories will be created as needed by the build system.

Your patch also erroneously causes directories such as `bin-x86_64-efi` to no longer be ignored.

markrogalski:
A minor issue is the non-symetry in that committing what I just cloned does not produce equivalent trees.
The major issue is that not having config/local in the build tree produces build errors like this:
```
  [DEPS] libgcc/__umoddi3.c
touch: cannot touch 'config/local/general.h': No such file or directory
touch: cannot touch 'config/local/console.h': No such file or directory
touch: cannot touch 'config/local/ioapi.h': No such file or directory
touch: cannot touch 'config/local/time.h': No such file or directory
touch: cannot touch 'config/local/nap.h': No such file or directory
touch: cannot touch 'config/local/reboot.h': No such file or directory
touch: cannot touch 'config/local/entropy.h': No such file or directory
touch: cannot touch 'config/local/fdt.h': No such file or directory
touch: cannot touch 'config/local/crypto.h': No such file or directory
touch: cannot touch 'config/local/usb.h': No such file or directory
touch: cannot touch 'config/local/timer.h': No such file or directory
touch: cannot touch 'config/local/branding.h': No such file or directory
touch: cannot touch 'config/local/umalloc.h': No such file or directory
touch: cannot touch 'config/local/fault.h': No such file or directory
touch: cannot touch 'config/local/colour.h': No such file or directory
make: *** No rule to make target 'config/local/nap.h', needed by 'bin/null_nap.o'.  Stop.
```
I.e. Running 'make' does not produce the config/local directory as needed.

Feel free to close my PR and resolve as you think appropriate. Thanks.

markrogalski:
A fresh clone includes the config/local directory. No problem then. The problem is when you don't have config/local. Make does not create the directory. I've tried on Ubuntu 22.04 and UBI 9.3 and build fails every time config/local is not present. Manually make the 'local' dir and build works fine.

mcb30:
I've modified and pushed this, since it is generally cleaner to have the `.gitignore` rules not cover files that are present in the repository.

You may want to consider using git submodules instead, since that's generally a cleaner way to embed external projects in your own git repository.  Using submodules means that you don't lose the whole of the subproject's history, makes it trivial to update to a newer (or older) version of the subproject (even if files have been added, deleted, or moved around), and gives you a guarantee that this kind of issue can't arise.

markrogalski:
Thanks for the advice. We'll try submodules for any future update.

PR #1128, opened 2024-01-19
[bnxt] Add support for additional chip IDs

jw14812:
Add additional chip IDs that can be recognized as part of the thor family.


mcb30:
@jw14812 I've updated the commit with what I believe is your correct name and email address - could you please confirm or correct it?

Thanks!

jw14812:
@mcb30 - I was not able to view the updated commit contents.  My info is Joseph Wong;  joseph.wong@broadcom.com.  I will review my account info to make sure it is accurate.

Thanks,

PR #1130, opened 2024-01-23
fix: crash/reboot loop on Mac mini 2018

aarnaud:
On cold boot the wifi card is present on snp/nii and ipxe crash during device initialization. I discovered that grub didn't have this issue. I tried to understand why: on a cold boot grub has 2 interfaces efinet0 and efinet1, efinet0 failed but efinet1 is good it's the ethernet port. So I did debug deeper with iPXE to fix this issue.

iPXE is usable on Mac Mini 2018 with efi boot on usb key or nvme drive. I didn't found yet the correct DHCP setting to boot iPXE directly over network. 

But now it's working fine. 

Screenshot just before crash and reboot:
image


src/interface/efi/efi_veto.c
```diff
@@ -30,6 +30,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 #include <ipxe/efi/Protocol/LoadedImage.h>
 #include <ipxe/efi/Protocol/ComponentName.h>
 #include <ipxe/efi/efi_veto.h>
+#include <stdio.h>
```
stappersg, line 33: That `include` should probably be avoided.

aarnaud: my bad, it's removed now

PR #1136, opened 2024-01-30
[efi] Fix hang during ExitBootServices()

rosslagerwall:
When ExitBootServices() invokes efi_shutdown_hook(), there may be nothing to generate an interrupt since the timer is disabled in the first step of ExitBootServices(). Additionally, for VMs OVMF masks everything from the PIC (except the timer) by default. This means that calling cpu_nap() may hang indefinitely. This was seen in practice in netfront_reset() when running in a VM on XenServer.

Fix this by skipping the halt if an EFI shutdown is in progress.


mcb30:
Good catch!  I've generalised this to also cover the other supported CPU architectures, since it's likely to affect all of them.

PR #1145, opened 2024-02-08
[eepro100] Sort PCI_ROM() lines

stappersg:


stappersg:
> I'm happy with the kind of clean-up in this PR and in #1146, but I'd prefer it to be done as a single bulk commit that updates all `PCI_ROM()` lines in all drivers at the same time, rather than having one commit per driver.
> 
> I would suggest automating this process, rather than spending hours manually reordering lines.
> 
> For review purposes, it should be possible to use `git show --color-moved` to verify that the commit performs only movements of lines, with no modifications.

That became merge request #1158.

Issue #1159, opened 2024-02-21
LoongArch EFI images fail to build


asheplyakov:
```
make ARCH=loong64 CROSS_COMPILE=loongarch64-linux-gnu- bin-loong64-efi/ipxe.efi
[skipped lots of boring output]
Unrecognized relocation type 103
```

- GCC 13.2.0 + LoongArch SIMD patches from [loongson](https://github.com/loongson/gcc/tree/loongarch-gcc-vector)
- Binutils 2.41

PR #1160, opened 2024-02-21
elf2efi: ignore LoongArch PC-relative relocations and relaxations

asheplyakov:
With this patch I can (cross-) compile ipxe for LoongArch UEFI with GCC 13.2 and binutils 2.41+ (and netboot LoongArch machines).

Several new relocations types have been added in LoongArch ABI version 2.10. In particular
- R_LARCH_B16 (18-bit PC-relative jump)
- R_LARCH_B21 (23-bit PC-relative jump)
- R_LARCH_PCREL20_S2 (22-bit PC-relative offset)

Also relocation relaxations have been introduced. Recent GCC (13.2) and binutils 2.41+ use these types of relocations, which confuses elf2efi tool. As a result ipxe EFI images for LoongArch fail to build with the following error:
```
make ARCH=loong64 CROSS_COMPILE=loongarch64-linux-gnu- bin-loong64-efi/ipxe.efi
[skipped lots of boring lines]
Unrecognised relocation type 103
```

To avoid the problem
- Ignore R_LARCH_B{16,21} and R_LARCH_PCREL20_S2 (just like other PC-relative relocations)
- Ignore relaxations (R_LARCH_RELAX) for now. Relocation relaxations are basically optimizations, omitting them results in correct binary (although it might be suboptimal).

Fixes: #1159

### DIFF
--- a/src/util/elf2efi.c
+++ b/src/util/elf2efi.c
@@ -140,6 +140,12 @@
 #ifndef R_LARCH_64
 #define R_LARCH_64 2
 #endif
+#ifndef R_LARCH_B16
+#define R_LARCH_B16 64
+#endif
+#ifndef R_LARCH_B21
+#define R_LARCH_B21 65
+#endif
 #ifndef R_LARCH_B26
 #define R_LARCH_B26 66
 #endif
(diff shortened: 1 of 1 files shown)


mcb30:
@asheplyakov Thanks - have merged and also updated the matched copy of elf2efi.c in wimboot: https://github.com/ipxe/wimboot/commit/7f57dc3

Issue #1141, opened 2024-02-01
Feature request: support for /32 DHCP allocations


XioNoX:
Hi,

Thanks for iPXE, I'm slowly getting familiar with it.

On Linux it's possible to assign a /32 v4 IP to an interface (usually for VMs) with a specific route for outbound traffic. This allows to efficiently use IP space, without having to rely on tricks like proxy ARP.
For example on the VM side I can have `inet 10.192.24.3/32 scope global ens13`
As well as 
```
# ip route
default via 10.192.24.1 dev ens13 onlink 
10.192.24.1 dev ens13 scope link
```
Then on the hypervisor side `inet 10.192.24.1/32 scope link tap0`
and `10.192.24.3 dev tap0 proto static scope link`
(Note that the gw IP doesn't have to be from the same prefix as the VM IP)
This works very well.
It's also possible to have ISC-DHCP hand out DHCP offers compatible with that setup, see for example this [blogpost](https://blog.fhrnet.eu/2020/03/07/dhcp-server-on-a-32-subnet/).

The problem I'm facing is at the iPXE step. When DHCP offers such IP config to the client, iPXE accepts (see the /32 IP) it but then can't fetch the debian-installer files.
![Screenshot from 2024-01-31 18-37-53](https://github.com/ipxe/ipxe/assets/688817/6935c075-f246-4b6c-a0a7-e184c5e46177)

You can find some more of my troubleshooting steps (and DHCP exchange on this task [comment](https://phabricator.wikimedia.org/T300152#9503098).

Please let me know if I'm doing something wrong or if it's indeed a bug/missing feature. Happy to help, but I don't know C enough to send a PR :( 

Cheers,

NiKiZe:
This has been up for discussion before. (But can't find the post)
It is not a bug, limited ability to implement it in small code size.
Is there any RFC that defines this? I would agree that in a perfect world iPXE should have support for /32 on legacy IPv4, on the other hand, is it worth it? Why not use IPv6 with possibly NAT64 and DNS64 instead?

MalteJ:
+1

I can confirm the problem with IPv6: When the RA does not contain an ICMPv6 prefix information, iPXE should configure with the address from DHCPv6 with a /128 prefix length and add the router as advertised by the RA. Instead iPXE configures the DHCPv6 assigned IP with a /64 subnet and the default route is not configured.

CC: @mcb30

wfleurant:
[...]

PR #1163, opened 2024-02-29
[ipv6] Support/Use DHCPv6 assigned address if there is no Prefix option in RA

guvenc:
It is a valid case that RAs dont have any Prefix options but still set the router/gateway when they are received. In such a case, the ipv6 address received via DHCPv6 needs to use the RA sender address as gateway.

This is outlined in the linked RFC 4861 section 6.3.4
https://www.rfc-editor.org/rfc/rfc4861#section-6.3.4

The current implementation doesnt take into the consideration the received RA sender address as a gateway, as required by 
RFC 4861, when there is no Prefix option attached to the RA message.

This PR is an attempt to correct this behavior.
This is related to the issue here as well: #1141 and to #449

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -1791,6 +1791,9 @@ const struct setting_type setting_type_ipv6 __setting_type = {
 /** IPv6 settings scope */
 const struct settings_scope dhcpv6_scope;
 
+/** NDP settings scope */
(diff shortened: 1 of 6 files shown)


mcb30:
Rewritten to use a different approach and pushed via #1245.  Please let me know if this doesn't fix your use case.

PR #1165, opened 2024-03-01
Loong64 cflags

yetist:
Prevent use of lsx and lasx registers and disable relaxations.


src/arch/loong64/Makefile
```diff
@@ -8,11 +8,20 @@ ASM_TCHAR_OPS	:= @
...
+
 # Check if -mno-explicit-relocs is valid
 ifeq ($(CCTYPE),gcc)
 MNER_TEST = $(CC) -mno-explicit-relocs -x c -c /dev/null -o /dev/null >/dev/null 2>&1
 MNER_FLAGS := $(shell $(MNER_TEST) && $(ECHO) '-mno-explicit-relocs')
 WORKAROUND_CFLAGS += $(MNER_FLAGS)
+
+# Check if -mno-relax is valid
+MNRX_TEST = $(CC) -mno-relax -x c -c /dev/null -o /dev/null >/dev/null 2>&1
+MNRX_FLAGS := $(shell $(MNRX_TEST) && $(ECHO) '-mno-relax' || $(ECHO) '-Wa,-mno-relax' )
```
mcb30, line 23: This looks wrong: we check for support of `-mno-relax`, but then apply `-Wa,-mno-relax`.  Why is this different?

mcb30 (changes requested):

yetist:
Updated, please review.

src/arch/loong64/Makefile
```diff
@@ -8,6 +8,18 @@ ASM_TCHAR_OPS	:= @
...
 
+## Check if the follow options is valid for current toolchains
+# -mno-lsx: Disable 128-bit SIMD instructions, added after gcc-14
+# -mno-lasx: Disable 256-bit SIMD instructions, added after gcc-14
+# -mno-relax: Disable linker relaxation, added after gcc-14
+# -Wa,-mno-relax: Disable linker relaxation, added after binutils-2.41
+CHECK_CC_FLAGS := -mno-lsx -mno-lasx -mno-relax -Wa,-mno-relax
+
+define check_cc_flag
+  $(shell $(CC) $(1) -x c -c /dev/null -o /dev/null > /dev/null 2>&1 && echo $(1))
```
xtexChooser, line 19: `-mno-lsx -mno-lasx -mno-relax` are added in the same version of GCC. Maybe they can share one check instead of checking for three times?
btw @mcb30 this looks good to me and does fixed issues on my machine. Could you review this?

Issue #138, opened 2020-09-03
Access to Current Working URI (cwuri)


omniproc:
**Description**

As a user, I would like the option of accessing the URI the current iPXE was loaded from (cwuri) from an iPXE script.

**Acceptance criteria**

- [ ] Have a IPXE script variable to access the cwuri when booting iPXE over EFI HTTP boot
- [ ] Split the URI into it's components (protocol, hostname, port, path, query) which can be accessed individually

omniproc:
Rational behind this: when we have the option to know where the initial iPXE image came from at runtime of the iPXE script we can load additional ressources from that location without the need to hardcode the server / location in the iPXE script. This provides great flexibility for any dynamic chainloading process. A single image can then be used to have the same install procedure in different geographical locations / on-/off-prem and so on (distributed) where latency and bandwidth would make it  hard to serve dynamic data from a single source.

NiKiZe:
If you do imgfetch script.ipxe, that should already be relative, no need to get or mangle the URL?

omniproc:
Well, not quite. This works only for TFP or mounted filesystems but not if the image was loaded from an URL e.g. using EFI HTTP(S) boot.

NiKiZe:
In that case, how was the initial script to iPXE provided?

omniproc:
UEFI HTTP Boot. The URL to the initial image is provided to UEFI as a boot option. The initial image can contain a generic (generic to the extend that it doesn't need to include any static URLs / data but can load anything dynamicly depending on where it was initially loaded from) embedded ipxe script that can then chainload more specific scripts from a path relative to where the image came from.

stappersg:
For what it is worth:  I just stumbled upon http://boot.ipxe.org/tinycore.ipxe which has this content:
```text
#!ipxe

set base http://tinycorelinux.net/11.x/x86/release/distribution_files

cpuid --ext 29 && set arch 64 || set arch

kernel ${base}/vmlinuz${arch} initrd=rootfs.gz initrd=modules${arch}.gz
initrd ${base}/rootfs.gz
initrd ${base}/modules${arch}.gz
boot
```
in `${base}` do I see `${cwuri}`.

( Now _time will tell_ what the effect is of touching this stalled issue ... )

omniproc:
@stappersg 

no worries. I'm still watching this issue and hoping for it to be implemented.
[...]

PR #1168, opened 2024-03-04
New iPXE setting for efi HTTPS boot uri source.

keithga:
uEFI provides a method for retrieving the source path of the binary, even when an uEFI binary was booted over HTTPS.

The source is stored as a Device Path, and @mcp30 already has a solution in efi_int() to extract out the source Device Path from uEFI, it's just a matter of parsing and saving the result so any iPXE script can leverage the path.

This is of importance on machines like the Lenovo m80q, which will ONLY HTTPS boot from Secure Boot signed binaries. It becomes economically difficult to hard code a server address into the singed binary for each customer/site. It would be preferred to find the source uri using this method with a single signed file.

Tested and verified on my Dell Optiplex 7010. 

**Implementation Notes**: _The basic structure of this PR appears to be correct from my limited iPXE experience, although I'm not sure if `efivars_init()` is the best place to start. If there is feedback, I can work to fix, however I would not be offended if we dump my code, and choose to rewrite._ 

cc: @Hammarskjold 
Issue: #138

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -2443,6 +2443,13 @@ const struct setting root_path_setting __setting ( SETTING_SANBOOT, root-path)={
 	.type = &setting_type_string,
 };
 
+/** Root path setting */
+const struct setting uri_path_setting __setting ( SETTING_BOOT, uri-path)={
+	.name = "uri-path",
+	.description = "HTTPS uri source from uEFI",
+	.type = &setting_type_string,
+};
+
 /** SAN filename setting */
 const struct setting san_filename_setting __setting ( SETTING_SANBOOT,
 						      san-filename ) = {
--- a/src/include/ipxe/efi/efi_uripath.h
+++ b/src/include/ipxe/efi/efi_uripath.h
@@ -1,0 +1,17 @@
+#ifndef _IPXE_EFI_URIPATH_H
+#define _IPXE_EFI_URIPATH_H
+
+/** @file
+ *
+ * EFI uri-path
+ *
+ */
+
+FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
+
+#include <ipxe/efi/efi.h>
+
+extern int efi_set_uri_path ( EFI_HANDLE device,
+		EFI_DEVICE_PATH_PROTOCOL *path );
+
+#endif /* _IPXE_EFI_URIPATH_H */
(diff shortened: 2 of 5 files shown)


mcb30:
It is a good idea to expose the URI from the loaded image device path within iPXE.  This approach is not the way to do it, however.

We already have a concept of a current working URI in iPXE.  I have just pushed commit https://github.com/ipxe/ipxe/commit/88c2a01e1 to expose this via [`${cwuri}`](https://ipxe.org/cfg/cwuri) and [`${cwduri}`](https://ipxe.org/cfg/cwduri).

I will work on a further commit to set the current working URI based on the loaded image URI device path, if present.  The loaded image device path will then be accessible via the above settings, and relative URIs will be interpreted as being relative to iPXE's own URI.  This is a more general solution than your approach.

This overlaps slightly with the work to obtain `autoexec.ipxe` from a UEFI HTTP boot, since that will also require access to the URI from the loaded image device path.

mcb30:
> It is a good idea to expose the URI from the loaded image device path within iPXE. This approach is not the way to do it, however.
> 
> We already have a concept of a current working URI in iPXE. I have just pushed commit [88c2a01e1](https://github.com/ipxe/ipxe/commit/88c2a01e1) to expose this via [`${cwuri}`](https://ipxe.org/cfg/cwuri) and [`${cwduri}`](https://ipxe.org/cfg/cwduri).
> 
> I will work on a further commit to set the current working URI based on the loaded image URI device path, if present. The loaded image device path will then be accessible via the above settings, and relative URIs will be interpreted as being relative to iPXE's own URI. This is a more general solution than your approach.

Implemented in commit https://github.com/ipxe/ipxe/commit/390bce951

Please open a new issue if the upstream approach does not address your use case.

NiKiZe:
@mcb30 any chance we could get the rawefidevicepath or maybe better named efibootpath ? Since that have som other potential values about boot.

mcb30:
> @mcb30 any chance we could get the rawefidevicepath or maybe better named efibootpath ? Since that have som other potential values about boot.

Unfortunately not: there is no guarantee that the EFI_DEVICE_PATH_TO_TEXT_PROTOCOL will be installed, and we've definitely seen some (Apple?) systems that don't include it.  This is why iPXE uses `efi_devpath_text()` only for debug messages.

What kind of information are you thinking it would be useful to obtain?

PR #1174, opened 2024-03-17
[golan.c] Fix inline header size

rabiamn:
Fixing the issue https://github.com/ipxe/ipxe/issues/1091
Fix Eth inline header size to 14 bytes instead of 18 bytes. because the eapol packet is 18 bytes and all the packet was inserted to the inline header, so it appears as empty packet and the driver cant handle it.

Rabia manaa
rabiam@nvidia.com

### DIFF
--- a/src/drivers/infiniband/golan.c
+++ b/src/drivers/infiniband/golan.c
@@ -2502,7 +2502,7 @@ static mlx_status shomron_fill_eth_send_wqe ( struct ib_device *ibdev,
 	}
 
 #define SHOMRON_GENERATE_CQE 0x3
-#define SHOMRON_INLINE_HEADERS_SIZE 18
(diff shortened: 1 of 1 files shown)


mcb30:
That's great, thank you.  Could you please update the commit to have your proper name and email address?  (Alternatively, since it's such a small change, just let me know what name and email you want to appear and I'll update it)

rabiamn:
> That's great, thank you. Could you please update the commit to have your proper name and email address? (Alternatively, since it's such a small change, just let me know what name and email you want to appear and I'll update it)

Rabia Manaa
rabiam@nvidia.com

mcb30:
Merged; thank you.  I rewrote the commit message and adjusted the change to use `ETH_HLEN` so that it's clear where the magic number 14 came from.

I've been looking at this issue recently: it's good to have a vendor bugfix submitted!

PR #1047, opened 2023-10-03
[netdevice] Add ll-protocol setting

Porsh33:
Add a new setting that allows using "netdev->ll-protocol" property in scripts. This can be useful in order to skip configuring interfaces using their link layer protocol or, conversely, configure only selected interface type (Ethernet, IPoIB, etc.)

Example script:
```
set idx:int32 0
:loop
isset ${net${idx}/mac} || exit 0
iseq ${net${idx}/ll-protocol} IPoIB && goto try_next ||
autoboot net${idx} ||
:try_next
inc idx && goto loop
```

PR #1180, opened 2024-03-25
[netdevice] Add ll-protocol setting

Porsh33:
This PR continues the previous, accidentally closed PR https://github.com/ipxe/ipxe/pull/1047

Added a new setting that allows using this property in scripts. This can be useful in order to skip configuring
interfaces using their link layer protocol or, conversely, configure only selected interface type (Ethernet, IPoIB, etc.). In systems which have 8+ IB cards and one Eth card, this script can save 80+ seconds to boot by skipping await of IB's to be configured

Example script:
```
set idx:int32 0
:loop
isset ${net${idx}/mac} || exit 0
iseq ${net${idx}/ll-protocol} IPoIB && goto try_next ||
autoboot net${idx} ||
:try_next
inc idx && goto loop
```

### DIFF
--- a/src/include/ipxe/settings.h
+++ b/src/include/ipxe/settings.h
@@ -471,6 +471,8 @@ mac_setting __setting ( SETTING_NETDEV, mac );
 extern const struct setting
 busid_setting __setting ( SETTING_NETDEV, busid );
 extern const struct setting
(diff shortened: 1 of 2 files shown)


mcb30:
I've modified the commit to use `linktype` as the setting name instead of `ll-protocol`, largely for consistency with the other net device setting names such as `bustype`.

PR #1184, opened 2024-04-04
Create a build pipeline that builds, publishes and generates checksums

### LINKED ISSUE #1183: Provide checksums for downloads on boot.ipxe.org
I'd like to verify the the items I'm downloading from boot.ipxe.org are what they say there are and have been built by ipxe.org. Are you able to start providing checksums or a signature file?

TomTucka:
Fixes #1183. 

@mcb30 @NiKiZe I made a start on this yesterday before I saw your latest comments. Let me know what you think of the general approach I'm taking, I probably won't have time to do more on this until the weekend.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,9 +3,33 @@ name: Build and Release
 on: push
 
 jobs:
+
+  cache:
+    name: Cache
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Cache permissions
+        run: |
+          sudo chown $(id -un) /var/cache/apt/archives
+      - name: Cache packages
+        uses: actions/cache@v4
+        with:
+          path: /var/cache/apt/archives/*.deb
+          key: apt-cache-${{ github.run_id }}-${{ github.run_attempt }}
+          restore-keys: |
+            apt-cache-
+      - name: Download packages
+        run: |
+          sudo apt update
+          sudo apt install -y -d -o Acquire::Retries=50 \
+                           mtools syslinux isolinux \
+                           libc6-dev-i386 valgrind \
+                           gcc-arm-none-eabi gcc-aarch64-linux-gnu
+
   build:
     name: Build on ${{ matrix.arch }}
     runs-on: ubuntu-22.04
+    needs: cache
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


NiKiZe:
We are currently not making releases, if we want to have version numbers etc, that is a whole separate set of changes.

.github/workflows/build.yml
```diff
@@ -26,96 +26,92 @@ jobs:
...
-      - name: Build (BIOS)
-        run: |
-          make -j 4 -C src
-      - name: Build (Everything)
-        run: |
-          make -j 4 -C src everything
-      - name: Test
-        run: |
-          valgrind ./src/bin-i386-linux/tests.linux
-          valgrind ./src/bin-x86_64-linux/tests.linux
```
mcb30, line 63: The change to use a matrix seems to have deleted the test step

.github/workflows/build.yml
```diff
@@ -26,96 +26,92 @@ jobs:
...
+          ${{ matrix.build_commands }}
 
-  arm64:
-    name: ARM64
+      - name: Upload artifacts to workflow
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.arch }}-artifacts
+          path: |
+            src/bin-*
```
mcb30, line 80: This would upload all object files, dependency fragments, etc., which seems unlikely to be what is intended.  (It would also skip the `bin/` directory used for BIOS builds)

.github/workflows/build.yml
```diff
@@ -26,96 +26,92 @@ jobs:
...
           fetch-depth: 0
-      - name: Cache permissions
-        run: |
-          sudo chown $(id -un) /var/cache/apt/archives
-      - name: Cache packages
-        uses: actions/cache/restore@v4
+
+      - name: Bump version
+        id: bump_version
+        uses: anothrNick/github-tag-action@1.67.0
```
mcb30, line 94: I avoid using random third-party actions in order to not fall into the https://xkcd.com/2347/ trap

.github/workflows/build.yml
```diff
@@ -26,96 +26,92 @@ jobs:
...
+          INITIAL_VERSION: 1.21.1
+          DEFAULT_BUMP: minor
+          PRERELEASE: true
+          PRERELEASE_SUFFIX: ${{ env.BRANCH_NAME }}
+          RELEASE_BRANCHES: master
+          WITH_V: true
+
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
```
mcb30, line 106: A GitHub release per commit seems like overkill

mcb30 (changes requested):
I'm happy to use the matrix approach to condense the build file, provided that doing so doesn't actually increase the overall complexity.

I'm not convinced that a GitHub release per commit is the way to go - this seems to violate the way that releases are generally expected to be used.

NiKiZe:
I think it would be nice to see one commit using matrix where possible, and another using assets, and one with doing boot.ipxe.org upload based on wimboot qa.yml ?

NiKiZe:
Closing this on the same basis as https://github.com/ipxe/ipxe/pull/1602#issuecomment-3792875397

PR #1193, opened 2024-04-16
Veto for Insyde Serial Terminal Driver.

hetii:
This pull request is to solve uart issue described in the issue https://github.com/ipxe/ipxe/issues/1189

### DIFF
--- a/src/interface/efi/efi_veto.c
+++ b/src/interface/efi/efi_veto.c
@@ -520,6 +520,37 @@ static int efi_veto_dhcp6 ( EFI_DRIVER_BINDING_PROTOCOL *binding __unused,
 	return 1;
 }
 
+/**
+ * Veto Insyde Serial Terminal Driver
+ *
+ * @v binding           Driver binding protocol
+ * @v loaded            Loaded image protocol
+ * @v wtf               Component name protocol, if present
+ * @v manufacturer      Manufacturer name, if present
+ * @v name              Driver name, if present
+ * @ret vetoed          Driver is to be vetoed
+ */
+static int
+efi_veto_insyde_serial_terminal ( EFI_DRIVER_BINDING_PROTOCOL *binding __unused,
+			    EFI_LOADED_IMAGE_PROTOCOL *loaded __unused,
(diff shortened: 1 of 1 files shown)


mcb30:
I think we need to have the veto take effect only if `CONSOLE_SERIAL` is enabled, in the same way that `efi_veto_hp_xhci()` will veto the HP xHCI driver only if we are expecting to drive the xHCI hardware ourselves.

This needs to be done in a way that doesn't involve putting the relevant code inside `#ifdef CONSOLE_SERIAL` because that's a pattern I strongly avoid in iPXE (see e.g. https://dox.ipxe.org/ifdef_harmful.html).  It also can't use a symbol reference that would cause the serial console code to be dragged in unconditionally.  Possible solutions are to use a sideband lookup mechanism (as for xHCI, though I don't think there's anything directly equivalent we could use for serial), to use a weak symbol (see e.g. `vlan_tci()` in `netdevice.c`), or to use a limited `#ifdef` block that merely assigns a constant value (see e.g. `shim_cmd.c`).

Some thought needs to be given as to whether the veto should depend upon the presence of `serial.o` or `uart.o`: the former indicates that we are expecting to control the serial console directly, the latter indicates that we are expecting to control a serial port directly for either console or some other purpose (e.g. a gdb connection).

With the above solved, I wonder if it would be sensible to drop the restriction on the vendor being Insyde, and have the veto take effect for anything that we can identify as a serial port driver if iPXE is expecting to drive the serial port directly itself?  In the default configuration (no `CONSOLE_SERIAL`) we would not veto anything.  Enabling `CONSOLE_SERIAL` in an EFI build would then mean "take ownership of and use the serial port" rather than the current "use the serial port and hope that nothing else is already using it".

stappersg:
> hetii force-pushed the veto-insyde-serial-terminal-driver

NiKiZe:
Is this something that is likely to have changed or could be revisited with several of the Serial changes that has been done since this PR was opened?

NiKiZe:
Another ping, is the issue still relevant with current master, and does this fix the issue?

PR #1206, opened 2024-04-30
tls: answer empty client certificate if we don't have any

LadyNamedLaura:
Some webservers always request a client certificate, but don't require it to be provided.

The rfc tells us that in that case we should send an empty list of certificates, and not send the verify packet.

This PR changes the behavior of `tls_new_certificate_request()` to not error out when no certificate is found, but instead just leaves `tls->certs` as an empty "chain".
In `tls_validator_done()` we then set `TLS_TX_CERTIFICATE_VERIFY` only if there is at least one certificate in `tls->certs`

### DIFF
--- a/src/net/tls.c
+++ b/src/net/tls.c
@@ -2466,35 +2466,33 @@ static int tls_new_certificate_request ( struct tls_connection *tls,
 	x509_chain_put ( tls->certs );
 	tls->certs = NULL;
 
-	/* Determine client certificate to be sent */
-	cert = certstore_find_key ( tls->key );
-	if ( ! cert ) {
-		DBGC ( tls, "TLS %p could not find certificate corresponding "
-		       "to private key\n", tls );
-		rc = -EPERM_CLIENT_CERT;
-		goto err_find;
-	}
-	x509_get ( cert );
-	DBGC ( tls, "TLS %p selected client certificate %s\n",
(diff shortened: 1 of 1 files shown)


stappersg:
image
The commit has no commit message, only a commit summary.
Use `git log` for inspiration how commit messages look and make a commit message by `git commit --amend`. (And then do a `git push` to update this merge request.)

LadyNamedLaura:
@stappersg oops, my bad. I got too used to working on internal projects, I hope the updated commit message is more suitable.

stappersg:
On Mon, May 06, 2024 at 04:31:53AM -0700, Laura Peeters wrote:
> oops, I hope the updated commit message is more suitable.

Yes, it does.  Initialy I thought that moving of lines was not covered
in the commit message. Upon review I did see that the "and only send a
certificate verifymessage if we did find a certificate" covers the
relocation of lines.


Groeten
Geert Stappers
-- 
Silence is hard to parse

stappersg:
_This merge request branch is conflicting with main branch, please resolve._

image

mcb30:
A corrected version of this PR was implemented in commit https://github.com/ipxe/ipxe/commit/f115cfcf9

My apologies for failing to include an `Originally-implemented-by:` credit corresponding to this PR.

PR #1294, opened 2024-09-03
[ena] Set iPXE as the OS type for iPXE driver

osama-2298:
The Nitro card supports different OSs, each OS has a unique identifier.
Updating the iPXE OS identifier to the correct one.
Newer versions of Nitro properly identify iPXE OS.

### DIFF
--- a/src/drivers/net/ena.c
+++ b/src/drivers/net/ena.c
@@ -1163,7 +1163,7 @@ static int ena_probe ( struct pci_device *pci ) {
 	}
 	ena->info = info;
 	memset ( info, 0, PAGE_SIZE );
-	info->type = cpu_to_le32 ( ENA_HOST_INFO_TYPE_LINUX );
+	info->type = cpu_to_le32 ( ENA_HOST_INFO_TYPE_IPXE );
 	snprintf ( info->dist_str, sizeof ( info->dist_str ), "%s",
 		   ( product_name[0] ? product_name : product_short_name ) );
 	snprintf ( info->kernel_str, sizeof ( info->kernel_str ), "%s",
--- a/src/drivers/net/ena.h
+++ b/src/drivers/net/ena.h
@@ -191,14 +191,7 @@ struct ena_host_info {
 	uint32_t features;
 } __attribute__ (( packed ));
 
-/** Linux operating system type
- *
- * There is a defined "iPXE" operating system type (with value 5).
- * However, some very broken versions of the ENA firmware will refuse
- * to allow a completion queue to be created if the "iPXE" type is
- * used.
- */
-#define ENA_HOST_INFO_TYPE_LINUX 1
+#define ENA_HOST_INFO_TYPE_IPXE 5
 
 /** Driver version
  *


NiKiZe:
I think it would be best if the commit detailed which versions or similar this works with, as well as if there still is any non working available.

> very broken versions of the ENA firmware will refuse
  to allow a completion queue to be created if the "iPXE" type is used.

From 3b81a4e2568cfca50893095638a03069785cd33e

Nit, subject still in commit should probably be updated following what is described at https://ipxe.org/admin and all already existing commits.

mcb30:
As per the code comment referenced by @NiKiZe, we need to lie and pretend that we are a Linux kernel to work around known bugs in the firmware as deployed by Amazon.

To merge this commit, we would need to have some kind of authoritative statement from Amazon that the firmware bug has been fixed and deployed to all AWS machines in all AWS regions.  Are you able to provide this?

ndagan:
The EC2 fleet is fully updated to support this iPXE change.

mcb30:
> The EC2 fleet is fully updated to support this iPXE change.

Is there any kind of public document which states this?  (A GitHub comment from a first-time GitHub user doesn't count, sorry.)

ndagan:
This is a long-standing gap that was recently fixed by ENA to properly detect iPXE drivers.  I lead ENA in AWS and we confirm the fix is available.

NiKiZe:
> This is a long-standing gap that was recently fixed by ENA to properly detect iPXE drivers. I lead ENA in AWS and we confirm the fix is available.

Not a question about availability, but rather deployed.
Or to take it even a step further, can we be sure the buggy firmware isnt available/deployed anyware? (And never again will be)

As mentioned, could this be verified thru a channel that isn't newly created GitHub accounts? 😉

ndagan:
The fix is fully deployed world-wide.  We do not have such public documentation.

mcb30:
> This is a long-standing gap that was recently fixed by ENA to properly detect iPXE drivers. I lead ENA in AWS and we confirm the fix is available.

Thank you for responding.  I do see your email address in the Linux ENA driver git logs.

> The fix is fully deployed world-wide. We do not have such public documentation.

I note that the Linux kernel driver still includes the [hack](92040c6daaa46027c2d11aab0a5366ab667c5358) that was required to work around the ENA firmware bug, and that this pull request does not change the equivalent magic value for the driver version number: https://github.com/ipxe/ipxe/blob/804f35cb5a6ac7d35912822372b74fd8856f850b/src/drivers/net/ena.h#L203-L222

Are we able to safely drop the use of `ENA_HOST_INFO_VERSION_WTF` completely?  Something like:
```patch
--- a/src/drivers/net/ena.c
+++ b/src/drivers/net/ena.c
@@ -1163,12 +1163,11 @@ static int ena_probe ( struct pci_device *pci ) {
 	}
 	ena->info = info;
 	memset ( info, 0, PAGE_SIZE );
-	info->type = cpu_to_le32 ( ENA_HOST_INFO_TYPE_LINUX );
+	info->type = cpu_to_le32 ( ENA_HOST_INFO_TYPE_IPXE );
 	snprintf ( info->dist_str, sizeof ( info->dist_str ), "%s",
 		   ( product_name[0] ? product_name : product_short_name ) );
 	snprintf ( info->kernel_str, sizeof ( info->kernel_str ), "%s",
 		   product_version );
-	info->version = cpu_to_le32 ( ENA_HOST_INFO_VERSION_WTF );
 	info->spec = cpu_to_le16 ( ENA_HOST_INFO_SPEC_2_0 );
 	info->busdevfn = cpu_to_le16 ( pci->busdevfn );
 	DBGC2 ( ena, "ENA %p host info:\n", ena );
diff --git a/src/drivers/net/ena.h b/src/drivers/net/ena.h
index 0f280c700..d3742543c 100644
--- a/src/drivers/net/ena.h
+++ b/src/drivers/net/ena.h
@@ -191,35 +191,8 @@ struct ena_host_info {
 	uint32_t features;
 } __attribute__ (( packed ));
 
-/** Linux operating system type
- *
- * There is a defined "iPXE" operating system type (with value 5).
- * However, some very broken versions of the ENA firmware will refuse
- * to allow a completion queue to be created if the "iPXE" type is
- * used.
- */
-#define ENA_HOST_INFO_TYPE_LINUX 1
-
-/** Driver version
- *
- * The driver version field is nominally used to report a version
- * number outside of the VM for consumption by humans (and potentially
- * by automated monitoring tools that could e.g. check for outdated
- * versions with known security flaws).
- *
- * However, at some point in the development of the ENA firmware, some
- * unknown person at AWS thought it would be sensible to apply a
- * machine interpretation to this field and adjust the behaviour of
- * the firmware based on its value, thereby creating a maintenance and
- * debugging nightmare for all existing and future drivers.
- *
- * Hint to engineers: if you ever find yourself writing code of the
- * form "if (version == SOME_MAGIC_NUMBER)" then something has gone
- * very, very wrong.  This *always* indicates that something is
- * broken, either in your own code or in the code with which you are
- * forced to interact.
- */
-#define ENA_HOST_INFO_VERSION_WTF 0x00000002UL
+/** iPXE operating system type */
+#define ENA_HOST_INFO_TYPE_LINUX 5
 
 /** ENA specification version */
 #define ENA_HOST_INFO_SPEC_2_0 0x0200
```

davidarinzon:
Hi @mcb30

In order for iPXE to work with newer nitro versions and to sustain overall legacy maintainability across OSs,
we still require this magic value for the version to be set in combination with the iPXE OS.

davidarinzon:
Hey @mcb30 

Looks OK to me.
Thank you.

mcb30:
Merged #1297, and updated the published iPXE image list at https://ipxe.org/howto/ec2

PR #1298, opened 2024-09-07
[ipv6] Change ULA route scope to global for interoperability

y5c4l3:
IPv6 miniroute filters out routes that are out of scope

https://github.com/ipxe/ipxe/blob/f88761ef491c16a33078e46ad9d49ebd8f36fd47/src/net/ipv6.c#L320-L338

It breaks the global scope reachability if there is some mechanism implemented at subsequent hops that ensures the global reachability, like NPTv6 which is common in multihoming networks.

Typical multihoming NPTv6

```
iPXE> route
net0: 10.1.16.171/255.255.255.0 gw 10.1.16.1
net0: fd00:dead:beef:10:5054:ff:feca:4f05/64 gw fe80::3a22:d6ff:fee2:3a76
net0: fe80::5054:ff:feca:4f05/64
iPXE> ping 2001:4860:4860::8888
IPv6 has no route to 2001:4860:4860::8888
0 bytes from <none>: seq=1: Connection timed out (https://ipxe.org/4c1b2092)
IPv6 has no route to 2001:4860:4860::8888
0 bytes from <none>: seq=2: Connection timed out (https://ipxe.org/4c1b2092)
IPv6 has no route to 2001:4860:4860::8888
0 bytes from <none>: seq=3: Connection timed out (https://ipxe.org/4c1b2092)
IPv6 has no route to 2001:4860:4860::8888
Finished: Operation canceled (https://ipxe.org/0b072095)
```

The workaround is changing the ULA route scope to global. **It might not be the best final solution.**

```
iPXE> route
net0: 10.1.16.171/255.255.255.0 gw 10.1.16.1
net0: fd00:dead:beef:10:5054:ff:feca:4f05/64 gw fe80::3a22:d6ff:fee2:3a76
net0: fe80::5054:ff:feca:4f05/64
iPXE> ping 2001:4860:4860::8888
64 bytes from 2001:4860:4860::8888: seq=1
64 bytes from 2001:4860:4860::8888: seq=2
64 bytes from 2001:4860:4860::8888: seq=3
64 bytes from 2001:4860:4860::8888: seq=4
Finished: Operation canceled (https://ipxe.org/0b072095)
```

### DIFF
--- a/src/net/ipv6.c
+++ b/src/net/ipv6.c
@@ -107,7 +107,7 @@ static unsigned int ipv6_scope ( const struct in6_addr *addr ) {
 	 * global.
 	 */
 	if ( IN6_IS_ADDR_ULA ( addr ) )
-		return IPV6_SCOPE_ORGANISATION_LOCAL;
+		return IPV6_SCOPE_GLOBAL;
 
 	/* All other addresses are assumed to be global */
 	return IPV6_SCOPE_GLOBAL;


stappersg:
As I see it, should this merge request be closed.

@y5c4l3 reproduce the `route` as you have in _iPXE_:
```text
net0: 10.1.16.171/255.255.255.0 gw 10.1.16.1
net0: fd00:dead:beef:10:5054:ff:feca:4f05/64 gw fe80::3a22:d6ff:fee2:3a76
net0: fe80::5054:ff:feca:4f05/64
```
on a full blown operation system and do the `ping 2001:4860:4860::8888` to find out that it neither does work **and please close this MR**.

mcb30:
> As I see it, should this merge request be closed.
> 
> @y5c4l3 reproduce the `route` as you have in _iPXE_:
> ```text
> net0: 10.1.16.171/255.255.255.0 gw 10.1.16.1
> net0: fd00:dead:beef:10:5054:ff:feca:4f05/64 gw fe80::3a22:d6ff:fee2:3a76
> net0: fe80::5054:ff:feca:4f05/64
> ```
> on a full blown operation system and do the `ping 2001:4860:4860::8888` to find out that it neither does work **and please close this MR**.

I don't think that's a sufficient reproduction to justify rejecting this change. You would need to not only reproduce the routing table but also the IPv6 NAT on the router. 

Although IPv6 NAT is an abomination, it is a valid use case, and so we should try to support it. That said, we need to make sure that we don't break other valid use cases, so we'll need to look at the commit history to see why we currently use this scope for ULAs.

NiKiZe:
You might have site-local routing, but not global, in that case IPv4 might be wanted instead.

Current state of RFCs also say that the routing priority should be:
Global IPv6
IPv4
Local IPv6

As such it have made sense to not use routing for ULA, and changing this might break for some with non working IPv6 configurations that currently is using IPv4 instead, but that is just hiding the error. (Might not be a bad thing to have these break)

y5c4l3:
> As I see it, should this merge request be closed.
> 
> @y5c4l3 reproduce the `route` as you have in _iPXE_:
> 
> ```
> net0: 10.1.16.171/255.255.255.0 gw 10.1.16.1
> net0: fd00:dead:beef:10:5054:ff:feca:4f05/64 gw fe80::3a22:d6ff:fee2:3a76
> net0: fe80::5054:ff:feca:4f05/64
> ```
> 
> on a full blown operation system and do the `ping 2001:4860:4860::8888` to find out that it neither does work **and please close this MR**.

@stappersg Sorry that I didn't include a minimal reproducible example. Given that a setup of NPTv6 playground is not so trivial, I should add a clue making the reachability more convincing:
ULA-only setup under Linux

y5c4l3:
> You might have site-local routing, but not global, in that case IPv4 might be wanted instead.
> 
> Current state of RFCs also say that the routing priority should be: Global IPv6 IPv4 Local IPv6
> 
> As such it have made sense to not use routing for ULA, and changing this might break for some with non working IPv6 configurations that currently is using IPv4 instead, but that is just hiding the error. (Might not be a bad thing to have these break)

Yeah I have to follow RFCs and that is why I was saying this is not optimal. This PR is more likely served as an issue with a naive workaround.

I can for sure use IPv4, but only IPv4 literals instead of hostnames, because `dns_resolv` in iPXE still prioritizes IPv6 DNS requests unless I unset the DNS6 server or force IPv4-only DHCP `dhcp -c dhcp` before any step.

```
iPXE> ping dns.google
NAMED 0x7d355c88 opening "dns.google"
RESOLV 0x7d355d08 attempting to resolve "dns.google"
RESOLV 0x7d355d08 trying method NUMERIC
IPv6 invalid separator 'n' in "dns.google"
RESOLV 0x7d355d08 trying method DNS
DNS 0x7d358808 sending fd00:dead:beef:2:200:5eff:fe00:102 query ID 0x49a4 for dns.google type AAAA
DNS 0x7d358808 sending 10.1.2.2 query ID 0x49a4 for dns.google type AAAA
DNS 0x7d358808 received response ID 0x49a4
DNS 0x7d358808 found address 2001:4860:4860::8888
INTF 0x7d358808+10->0x7d355c88+30 resolv_done
NAMED 0x7d355c88 redirected successfully
IPv6 has no route to 2001:4860:4860::8888
0 bytes from <none>: seq=1: Connection timed out (https://ipxe.org/4c1b2092)
IPv6 has no route to 2001:4860:4860::8888
0 bytes from <none>: seq=2: Connection timed out (https://ipxe.org/4c1b2092)
```

https://github.com/ipxe/ipxe/blob/a1830ff43c5f8ee2957eb537f24670265ebd233b/src/net/udp/dns.c#L1009-L1011

The name resolution behavior is however not conform to the destination address selection criteria defined in [RFC 3484](https://www.rfc-editor.org/rfc/rfc3484.html#page-5), which matches source / destination scope pairs. Here is a quick check upon `getaddrinfo` in Linux, which can be seen as a reference of the expected behavior:

Linux getaddrinfo behavior

If we got only ULAs on the interface, the DNS response carrying `AAAA` GUA addresses will not match the scope of a ULA, so it's sorted to the end.

stappersg:
image

Those we can close this issue, please do.

y5c4l3:
> Those we can close this issue, please do.

The `route` command in iPXE never shows anything about "default", otherwise there should be an IPv4 default route presented there. After checking the code, I can confirm that IPv4 miniroute fallbacks the route to `netX.gateway` while IPv6 does not. That's the right key to the problem here.

```
iPXE> show net0/gateway
net0.dhcp/gateway:ipv4 = 10.1.16.1
iPXE> show net0/gateway6
net0.ndp.0/gateway6:ipv6 = fe80::3a22:d6ff:fee2:3a76
iPXE> NDP net0 found default router fe80::3a22:d6ff:fee2:3a76 for on-link autonomous prefix fd00:dead:beef:10::/64
```

y5c4l3:
TLDR on issues found so far:
* DNS resolution default behavior does not follow RFC 3484
* Make `gateway6` a fallback gateway when appropriate

I'm trying to make a fix on these parts instead. Thanks for your patience. <3

stappersg:
> Thanks for your patience.

Thanks for closing this merge request.

:heart:

NiKiZe:
> TLDR on issues found so far:
> 
> * DNS resolution default behavior does not follow RFC 3484
> * Make `gateway6` a fallback gateway when appropriate
> 
> I'm trying to make a fix on these parts instead. Thanks for your patience. <3

You have a real issue, I personally would be fine with keeping this open, but it is up to you.

PR #1304, opened 2024-09-13
[ipv6] Use IPv6 fallback route when scope matching fails

y5c4l3:
When routing to an IPv6 destination, the lookup process ignores routes that are out of scope. This means any global unicast address is unreachable if the link only advertises a ULA prefix, which could be unwanted when subsequent hops implements address translations, like NPTv6.

This patch implements fallback routing with best efforts, preferring the largest IPv6 route scope which is iterated in the first place.


clonejo:
This patch would be quite useful to me, as i would like to use netboot.xyz in a ULA-only network with NPt. If a maintainer could have a look at this, that would be much appreciated =)

2psChristianN:
If I don't have a defined route to the outside, I would expect the stack to fallback to try and use legacy IPv4 connection instead.
If that also fails, then a global, non declared route, such as this one, could be tried, but only if defined IPv4 route fails.
Isn't this even covered in RFCs for preferences of different network prefixes?

PR #1330, opened 2024-10-30
[core] Add a new `isin` command

redat00:
Hi everyone ! 

We recently started using iPXE, but we lacked a proper way to write script like we wanted to. Especially because for some reason the SMBIOS data where not correctly formatted (whitespaces and other misc...). 

One easy solution for us was to be able to check that a certain substring was present in another string. But the iPXE shell was missing a command for this. That's the problem I'm trying to address with this PR.

For this, I wrote a new `isin` command. It is quite similar to the `iseq` command, but instead of comparing equality, it will compare the presence of the second string (the substring) in the first string.

```
iPXE> isin --help
Usage:

  isin <string> <substring>

See https://ipxe.org/cmd/isin for further information
iPXE> isin ${product} Capri2 && echo is_in || echo is_not_in
is_in
iPXE> isin ${product} test && echo is_in || echo is_not_in
is_not_in
iPXE> echo ${product}
Capri2
```

This is my first time ever writing C code, but it was fairly easy to do since the `iseq` command is quite similar. However if I'm missing something, feel free to say it to me and I'll correct my code.

Also if I'm missing something like adding documentation : Please, let me know, and I'll add whatever is required for this PR to be accepted!

Thanks in advance for your time,
Renaud.

### DIFF
--- a/src/core/exec.c
+++ b/src/core/exec.c
@@ -549,6 +549,42 @@ struct command iseq_command __command = {
 	.exec = iseq_exec,
 };
 
+/** "isin" options */
+struct isin_options {};
+
+/** "isin" option list */
+static struct option_descriptor isin_opts[] = {};
+
+/** "isin" command descriptor */
+static struct command_descriptor isin_cmd =
+	COMMAND_DESC ( struct isin_options, isin_opts, 2, 2,
+					 "<string> <substring>");
+
+/**
+ * "isin" command
+ *
+ * @v argc    Argument count
+ * @v argv    Argument list
+ * @ret rc    Return status code
+ */
+static int isin_exec (int argc, char **argv ) {
+	struct isin_options opts;
+	int rc;
+
+	/* Parse options */
+	if ( ( rc = parse_options ( argc, argv, &isin_cmd, &opts ) ) != 0 )
+		return rc;
+
+	/* Return success if first string contains second string */
+	return ( ( strstr ( argv[optind], argv[ optind +1 ] ) != NULL ) ?
+		0: -ERANGE );
+}
+
+struct command isin_command __command = {
+	.name = "isin",
+	.exec = isin_exec,
+};
+
 /** "sleep" options */
 struct sleep_options {};
 


redat00:
Hi ! 

I see the idea and can understand the fact that iPXE want to be as small as possible. Especially when it's also supposed to be burnt on devices.
However it seems like a big stretch to expect people to have something server side to do this kind of stuff.
But again, I can see the idea behind it.

(In my case since I already have an application running on the server, I might be able to make it work. But I would have preferred to not rely too much on an external server. It's one less thing to maintain. Or in my case less code to think about.)

Concerning the trim/strip thing : I didn't see any way to trim/strip a string within iPXE, but I might have missed that! Do you have any resources/documentation on how to use it ? 

In this specific case, yes it will help and solve my issue. But I still think that checking if a string contains a substring is something very basic that would be welcomed by all.

Thanks,
Renaud.

redat00:
Also, just wondering.

In some case we might accept that iPXE is a little bit bigger (for example when fetching iPXE overt TFTP, it can be a little heavier than when being written on say a network card). Would it then be possible somehow to have the ability to enable/disable some options ? 

I can live with the idea of having to build iPXE myself (which I already have to since I want to embed a script). So I'd be more than happy to see this kind of thing merged and have a bulkier version of iPXE, even it it means that I need to enable it with a flag when building it ! 

But this conversation is far more advanced than the idea behind this MR. 

Thanks,
Renaud.

mcb30:
> In this specific case, yes it will help and solve my issue. But I still think that checking if a string contains a substring is something very basic that would be welcomed by all.

I agree.  The problem is that there are many small bits of functionality that could be described as "would be welcomed by all", and they add up.

> In some case we might accept that iPXE is a little bit bigger (for example when fetching iPXE overt TFTP, it can be a little heavier than when being written on say a network card). Would it then be possible somehow to have the ability to enable/disable some options ?

Yes, you can already do this by modifying the various `config/local/*.h` files: see https://ipxe.org/buildcfg

stappersg:
> Will you be willing to merge this if I added a new option to do it ? So by default it's not present, but when needed, it can.

Just make such merge request and see what happens.

Karma bonus points for closing this MR (that already got _"wouldn't happen"_), the closing with will prevent further drain of human energy on this dangling MR.

NiKiZe:
> > Will you be willing to merge this if I added a new option to do it ? So by default it's not present, but when needed, it can.
> 
> Just make such merge request and see what happens.
> 
> Karma bonus points for closing this MR (that already got _"wouldn't happen"_), the closing with will prevent further drain of human energy on this dangling MR.

Don't follow this...
Unlikely that any such PR would be merged.
But if you do decide to try, you are better of rebasing and fixing this PR.

redat00:
Hi! 

Seeing how little interest this get, I'll just keep on having a forked version of iPXE. 

Since I have to build it anyway in order to embed our script, I'll just continue with applying Git patches to the code before compiling it.

Thanks for anyone who's interacted with this MR. Although the outcome is not what I hoped for, I'm still grateful for anyone taking the time to look at it!

PR #1386, opened 2025-01-23
[bnxt] Allocate TX rings with firmware input

jw14812:
Use queue_id value, retrieved from firmware, unconditionally when allocating TX rings.

Signed-off by: Joseph Wong <joseph.wong@broadcom.com>

### DIFF
--- a/src/drivers/net/bnxt/bnxt.c
+++ b/src/drivers/net/bnxt/bnxt.c
@@ -1840,7 +1840,7 @@ static int bnxt_hwrm_ring_alloc ( struct bnxt *bp, u8 type )
 		req->page_size   = LM_PAGE_BITS ( 8 );
 		req->int_mode    = RING_ALLOC_REQ_INT_MODE_POLL;
 		req->length 	 = ( u32 )bp->tx.ring_cnt;
-		req->queue_id    = TX_RING_QID;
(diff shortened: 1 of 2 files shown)


stappersg:
Please make it possible that the two separate commits remain two separate commits.

image

Yes, my fear is that _github_ mangles the two commits in one commit. That would be horrable for `git disect`.

NiKiZe:
> Please make it possible that the two separate commits remain two separate commits.
> 
> image
> 
> Yes, my fear is that _github_ mangles the two commits in one commit. That would be horrable for `git disect`.

Why would these commits not be able to stay separate? Maintainers do not use any of the defunct "merge" functions of GitHub, so there should not be any need to worry.

jw14812:
Agreed.  I created the initial pull request with only the first commit (Allocate TX rings with firmware input).  However when I pushed the second commit(a few days later) it got included into the pull request which was not my intention( to club the two together). I have done a force-push back to the first/original commit to eliminate the second.  Once this has been merged, I will re-submit the other with it's own pull request.  Thanks

jw14812:
@mcb30 - Is there anything further needed for this request.

mcb30:
> @mcb30 - Is there anything further needed for this request.

All merged, thank you for the fix!

Michael

PR #1403, opened 2025-02-10
[bnxt] Use updated DMA APIs

jw14812:
Replace malloc_phys with dma_alloc, free_phys with dma_free, alloc_iob with alloc_rx_iob, free_iob with free_rx_iob, virt_to_bus with dma or iob_dma.

### DIFF
--- a/src/drivers/net/bnxt/bnxt.c
+++ b/src/drivers/net/bnxt/bnxt.c
@@ -9,6 +9,7 @@ FILE_LICENCE ( GPL2_ONLY );
 #include <byteswap.h>
 #include <ipxe/pci.h>
 #include <ipxe/iobuf.h>
+#include <ipxe/dma.h>
 #include <ipxe/timer.h>
 #include <ipxe/malloc.h>
 #include <ipxe/if_ether.h>
@@ -340,7 +341,7 @@ static inline u32 bnxt_tx_avail ( struct bnxt *bp )
 	return ( avail-use );
 }
 
-void bnxt_set_txq ( struct bnxt *bp, int entry, dma_addr_t mapping, int len )
+void bnxt_set_txq ( struct bnxt *bp, int entry, physaddr_t mapping, int len )
 {
 	struct tx_bd_short *prod_bd;
 
@@ -355,7 +356,7 @@ void bnxt_set_txq ( struct bnxt *bp, int entry, dma_addr_t mapping, int len )
 	else
 		prod_bd->flags_type = TX_BD_SHORT_FLAGS_LHINT_GTE2K;
 	prod_bd->flags_type |= TX_BD_FLAGS;
-	prod_bd->dma.addr = mapping;
+	prod_bd->dma      = mapping;
 	prod_bd->len	  = len;
 	prod_bd->opaque   = ( u32 )entry;
 }
@@ -383,7 +384,7 @@ int bnxt_free_rx_iob ( struct bnxt *bp )
 
 	for ( i = 0; i < bp->rx.buf_cnt; i++ ) {
 		if ( bp->rx.iob[i] ) {
-			free_iob ( bp->rx.iob[i] );
+			free_rx_iob ( bp->rx.iob[i] );
 			bp->rx.iob[i] = NULL;
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/drivers/net/bnxt/bnxt.c
```diff
@@ -1823,13 +1841,13 @@ static int bnxt_hwrm_ring_alloc ( struct bnxt *bp, u8 type )
 		req->int_mode   = BNXT_CQ_INTR_MODE ( ( (FLAG_TEST ( bp->flags, BNXT_FLAG_IS_CHIP_P7) ) || bp->vf ) );
 		req->length 	= ( u32 )bp->nq.ring_cnt;
 		req->logical_id = 0xFFFF; // Required value for Thor FW?
-		req->page_tbl_addr = virt_to_bus ( bp->nq.bd_virt );
+		req->page_tbl_addr = ( unsigned long ) ( bp->nq.bd_virt );
```
mcb30, line 1844: This looks implausible to me.  The old code would have filled `page_tbl_addr` with a physical (bus) address, but the replacement code looks as though it fills `page_tbl_addr` with a virtual address.

src/drivers/net/bnxt/bnxt.c
```diff
@@ -2116,8 +2134,8 @@ int bnxt_hwrm_run ( hwrm_func_t cmds[], struct bnxt *bp )
 	int ret;
 
 	for ( ptr = cmds; *ptr; ++ptr ) {
-		memset ( bp->hwrm_addr_req,  0, REQ_BUFFER_SIZE );
-		memset ( bp->hwrm_addr_resp, 0, RESP_BUFFER_SIZE );
+		memset ( ( void * )bp->req_addr_mapping,  0, REQ_BUFFER_SIZE );
```
mcb30, line 2137: Doesn't necessarily need to be changed, but an explicit cast to `void *` is not necessary.

src/drivers/net/bnxt/bnxt.c
```diff
@@ -2152,9 +2170,6 @@ static void bnxt_tx_adjust_pkt ( struct bnxt *bp, struct io_buffer *iob )
 	dbg_tx_vlan ( bp, ( char * )iob->data, prev_len, iob_len ( iob ) );
 	if ( iob_len ( iob ) != prev_len )
 		prev_len = iob_len ( iob );
-
-	iob_pad ( iob, ETH_ZLEN );
```
mcb30, line 2156: The removal of TX padding seems to be unrelated to using the DMA APIs.  Is this change intentional?

src/drivers/net/bnxt/bnxt.h
```diff
@@ -933,6 +933,19 @@ struct bnxt {
...
 	void                      *hwrm_addr_dma;
+	struct dma_device         *dma;
+	struct dma_mapping        req_mapping;
+	struct dma_mapping        resp_mapping;
+	struct dma_mapping        dma_mapped;
+	struct dma_mapping        tx_mapping;
+	struct dma_mapping        rx_mapping;
+	struct dma_mapping        cq_mapping;
+	struct dma_mapping        nq_mapping;
+	dma_addr_t                tx_addr_mapping;
```
mcb30, line 944: There's no real need for these `dma_addr_t` fields, or for having `dma_addr_t` defined as a type specific to the Broadcom driver.  Since you're cleaning up the DMA code anyway, I would suggest deleting the `dma_addr_t` typedef in `bnxt.h`, removing all of these fields within `struct bnxt`, and instead using `dma()` at the point of use for each of these values.

mcb30 (changes requested):

src/drivers/net/bnxt/bnxt.c
jw14812: Restored this change, and will remove the unnecessary padding in a future commit

src/drivers/net/bnxt/bnxt.c
jw14812: Ack, and fixed

src/drivers/net/bnxt/bnxt.h
jw14812: Ack, removed this and created macros to do the dma() calls for better readability

jw14812:
@mcb30 - Hi Michael, I have pushed some changes to address your comments.  When convenient can you check if they have addressed your concerns.  Thanks

mcb30:
Looks plausible to me, thank you for updating.  Will QA and merge shortly.

mcb30:
All merged, thank you!

PR #1417, opened 2025-02-26
Reduce the size of built EFI driver when EFI_DOWNGRADE_UX is requested

shankerwangmiao:
When EFI_DOWNGRADE_UX is requested, the efiroms built from iPXE should be a simple driver, without the iPXE stack. However, the objects containing the iPXE stack are still dragged by `obj_config`, which makes the size larger. This patch series addresses this issue by 1. linking only necessary objects; 2. preventing requesting the main `ipxe ()` symbol from `efi_snp_load_file ()`


mcb30 (changes requested):
Sorry, this cannot be merged in the current form.  You'll need to find an alternative way of handling it that doesn't introduce `#ifdef`s in the middle of C code (see any of the various "#ifdef considered harmful" papers), and that doesn't introduce a separate special-case replacement for `config.c`.

Since `EFI_DOWNGRADE_UX` is a very specialist corner case that isn't even covered by my regular compile testing, let alone functional testing, the complexity overhead that I'm prepared to tolerate for maintaining it is basically zero.

You can rely on the compiler to perform dead code elimination within `efi_snp_load_file()`.  If you arrange for the start of this function to include code along the lines of
```c
    if ( constant_expression )
        return EFI_UNSUPPORTED;
```
where `constant_expression` evaluates to true if `EFI_DOWNGRADE_UX` is defined, then the compiler will eliminate the rest of the function along with any symbol references it might have otherwise generated.

shankerwangmiao:
> Sorry, this cannot be merged in the current form. You'll need to find an alternative way of handling it that doesn't introduce `#ifdef`s in the middle of C code (see any of the various "#ifdef considered harmful" papers), and that doesn't introduce a separate special-case replacement for `config.c`.
> 
> Since `EFI_DOWNGRADE_UX` is a very specialist corner case that isn't even covered by my regular compile testing, let alone functional testing, the complexity overhead that I'm prepared to tolerate for maintaining it is basically zero.

Do you have any alternative idea to avoid the replacement for `config.c`? Will it be acceptable if a wrapping `#ifdef` is added to `config.c`, so that `config.c` will become something like this:

```c
#ifndef EFI_DOWNGRADE_UX

#ifdef FOO
REQUIRE_OBJECT (foo);
#endif

#ifdef BAR
REQUIRE_OBJECT (bar);
#endif

#endif
```

> 
> You can rely on the compiler to perform dead code elimination within `efi_snp_load_file()`. If you arrange for the start of this function to include code along the lines of
> 
> ```c
>     if ( constant_expression )
>         return EFI_UNSUPPORTED;
> ```
> 
> where `constant_expression` evaluates to true if `EFI_DOWNGRADE_UX` is defined, then the compiler will eliminate the rest of the function along with any symbol references it might have otherwise generated.

I have moved the `#ifdef` to the front of `efi_snp.c`, and defined `is_efi_downgrade_ux` to reflect whether `EFI_DOWNGRADE_UX` is defined, which is then used as a constant_expression in `efi_snp_load_file ()` and thus `#ifdef` in the middle of C code is avoided.

mcb30:
> Do you have any alternative idea to avoid the replacement for `config.c`? Will it be acceptable if a wrapping `#ifdef` is added to `config.c`, so that `config.c` will become something like this:
> 
> ```c
> #ifndef EFI_DOWNGRADE_UX
> 
> #ifdef FOO
> REQUIRE_OBJECT (foo);
> #endif
> 
> #ifdef BAR
> REQUIRE_OBJECT (bar);
> #endif
> 
> #endif
> ```

No, that wouldn't work for me since it would still restart the exponential complexity problem that (ab)using `#ifdef`s tends to cause.

We do already have a related mechanism that may be useful to handle this kind of situation of "object X is required only if both feature X is enabled and object Y is already included".  See for example `config_usb.c`.  The `REQUIRE_OBJECT()`s for e.g. the various USB host controllers and non-NIC peripherals are located in `config_usb.c` rather than in `config.c`, and `config_usb.o` is in turn required by an explicit `REQUIRE_OBJECT()` in `drivers/bus/usb.c`.  The net result is that features such as `USB_KEYBOARD` may be left enabled by default, but the corresponding `usbkbd.o` object will be included only if something else already causes `usb.o` to be linked in to the build.

So, for example, the block of various `REQUIRE_OBJECT ( xxxx_cmd );` lines could be split out of `config.c` into a new `config_exec.c` with a `REQUIRE_OBJECT ( config_exec );` added to `core/exec.c`, on the basis that there's no point having the command implementations linked in if the code to execute commands is missing from the build anyway.

> I have moved the `#ifdef` to the front of `efi_snp.c`, and defined `is_efi_downgrade_ux` to reflect whether `EFI_DOWNGRADE_UX` is defined, which is then used as a constant_expression in `efi_snp_load_file ()` and thus `#ifdef` in the middle of C code is avoided.

Yes, that kind of approach is acceptable since it doesn't make the C code itself harder to read, and because all branches of the code (whether eliminated or not) will at least be syntax-checked in every build.

shankerwangmiao:
Hi, thanks for your explanation. I have moved most of `config.c` into a new `config_exec.c`, and let `autoboot.c` (actually the `ipxe()` function in it) require it. In this way, as long as the image needs `ipxe()` for the main ipxe boot prompt, the `config_exec` will get pulled in. 

- For `efidrv` with `EFI_DOWNGRADE_UX`: `efidrvprefix` will pull in `efi_snp`, where the call to `ipxe()` in `efi_snp_load_file()` is optimized out;
- For `efidrv` without `EFI_DOWNGRADE_UX `: `efidrvprefix` will pull in `efi_snp`, which needs `ipxe()` in `efi_snp_load_file()` and pulls in `autoboot`, which pulls in `config_exec`;
- For other images, the `*prefx`s need `main()` and pull in `main.o`, where `ipxe()` is needed and `autoboot` is pulled in, and `config_exec` is then pulled in.

`config_efi.c` is also moved to `config_efi_exec.c`, which is pulled in a `#ifdef EFI_DOWNGRADE_UX ` block in `config_efi.c`.

shankerwangmiao:
> e commands anyway. Similarly, the justification for `config_usb.c` is that the USB host controller drivers are relevant onl

Thanks for your clarification.  However, all the objects except `device` in `config.c` is not used by a `efidrv` with `EFI_DOWNGRADE_UX`, which means we should classify almost all the `REQUIRE` statements and put them into separate `config_foo.c` files. It will be a huge work and requires in depth understanding of iPXE. If not done properly, it may cause severe problems when some of the objects are missing in the final linked file. Due to the fact that iPXE is heavily relies on the table mechanism, miss some of the objects will not cause compilation errors, but only when the image begins to running should the problems shows. 

I propose that to begin with, can we move all the `REQUIRE` statements except `device` in `config.c` to something like `config_main.c` (or other more appropriate name) and let `autoboot.c` (actually the `ipxe()` function in it) requires it? Later we can classify the `REQUIRE` statements in a step-by-step way and finally reach the state that all the `REQUIRE` statements can find there way to a certain `config_foo.c`, which is pulled in by a certain depending function.

Regarding `config_efi.c`, we can also move all the contents to `config_efi_console.c`, and leave an empty `config_efi.c`, and let `efi_snp.c` and `efi_pxe.c` requires `config_efi_console` when `EFI_DOWNGRADE_UX` is not defined.

mcb30:
> Thanks for your clarification. However, all the objects except `device` in `config.c` is not used by a `efidrv` with `EFI_DOWNGRADE_UX`, which means we should classify almost all the `REQUIRE` statements and put them into separate `config_foo.c` files. It will be a huge work and requires in depth understanding of iPXE.

Yes.  That is kind of my point.  It needs to be done properly, not just hacked together to work around the single problem that is currently affecting you.

> I propose that to begin with, can we move all the `REQUIRE` statements except `device` in `config.c` to something like `config_main.c` (or other more appropriate name) and let `autoboot.c` (actually the `ipxe()` function in it) requires it? Later we can classify the `REQUIRE` statements in a step-by-step way and finally reach the state that all the `REQUIRE` statements can find there way to a certain `config_foo.c`, which is pulled in by a certain depending function.

Who do you propose is the person who will eventually do this work, or alternatively provide resources for it to be done?

As a very quick example of what your design would break: any attempt to build with `EFI_DOWNGRADE_UX` and any debug messages would not work as expected, because you would have ripped out all of the console drivers.

shankerwangmiao:
> > Thanks for your clarification. However, all the objects except `device` in `config.c` is not used by a `efidrv` with `EFI_DOWNGRADE_UX`, which means we should classify almost all the `REQUIRE` statements and put them into separate `config_foo.c` files. It will be a huge work and requires in depth understanding of iPXE.
> 
> Yes. That is kind of my point. It needs to be done properly, not just hacked together to work around the single problem that is currently affecting you.
> 
> > I propose that to begin with, can we move all the `REQUIRE` statements except `device` in `config.c` to something like `config_main.c` (or other more appropriate name) and let `autoboot.c` (actually the `ipxe()` function in it) requires it? Later we can classify the `REQUIRE` statements in a step-by-step way and finally reach the state that all the `REQUIRE` statements can find there way to a certain `config_foo.c`, which is pulled in by a certain depending function.
> 
> Who do you propose is the person who will eventually do this work, or alternatively provide resources for it to be done?

I should admit that I am not confident enough to do this job. 

> As a very quick example of what your design would break: any attempt to build with `EFI_DOWNGRADE_UX` and any debug messages would not work as expected, because you would have ripped out all of the console drivers.

Yes, you are correct on this. Console drivers are indeed necessary for normal iPXE builds. However, if debugging features are not enabled, and `EFI_DOWNGRADE_UX` is also requested, console drivers can also be omitted. 

Overall, it seems that `EFI_DOWNGRADE_UX` has huge differences to normal iPXE builds, and as a result, supporting for building an size optimized ROM seems to add complexity into the current build system.

PR #1418, opened 2025-02-26
[lkrnprefix] Support for longer version string

shankerwangmiao:
The lkrnprefix.S file contains a hardcoded 2-byte jump instruction to the setup function. When the version string is too long, the relative offset becomes larger than 127 bytes and the jump instruction is no longer valid. This patch fixes the issue by inserting a label _setup before the version string and the hardcoded jump instruction is changed to jump to _setup. We then jump to the setup function from _setup.


mcb30:
Fixed in https://github.com/ipxe/ipxe/commit/be3a78eaf by moving the version string to `.prefix.data`, which makes the problem go away without needing to add a second jump instruction.

Issue #1419, opened 2025-03-02
build errors with gcc 15


crobinso:
Fedora rawhide with gcc-15.0.1-0.8.fc43.x86_64
Full output of `make -j8 V=1 NO_WERROR=1 -k`: [build log](https://github.com/user-attachments/files/19044533/output.log)

I think the main issue is the first section in gcc 15 porting guide: function declaration without parameters. Example:

```
In file included from drivers/net/prism2_pci.c:30:
drivers/net/prism2_pci.c: In function ‘prism2_pci_driverprism2_pci_disable’:
drivers/net/prism2_pci.c:58:28: error: initialization of ‘void (*)(void)’ from incompatible pointer type ‘void (*)(struct nic *)’ [-Wincompatible-pointer-types]
   58 |          prism2_pci_probe, prism2_pci_disable );
      |                            ^~~~~~~~~~~~~~~~~~
include/nic.h:220:49: note: in definition of macro ‘DRIVER’
  220 |                 void ( * _unsafe_disable ) () = _disable;                 \
      |                                                 ^~~~~~~~
drivers/net/prism2_pci.c:47:13: note: ‘prism2_pci_disable’ declared here
   47 | static void prism2_pci_disable ( struct nic *nic ) {
      |             ^~~~~~~~~~~~~~~~~~
include/nic.h:221:17: error: too many arguments to function ‘_unsafe_disable’; expected 0, have 2
  221 |                 _unsafe_disable ( nic, hwdev );                           \
      |                 ^~~~~~~~~~~~~~~   ~~~
drivers/net/prism2_pci.c:57:1: note: in expansion of macro ‘DRIVER’
   57 | DRIVER ( "Prism2/PCI", nic_driver, pci_driver, prism2_pci_driver,
      | ^~~~~~
In file included from drivers/net/prism2_plx.c:30:
drivers/net/prism2_plx.c: In function ‘prism2_plx_driverprism2_plx_disable’:
drivers/net/prism2_plx.c:124:28: error: initialization of ‘void (*)(void)’ from incompatible pointer type ‘void (*)(struct nic *)’ [-Wincompatible-pointer-types]
  124 |          prism2_plx_probe, prism2_plx_disable );
      |                            ^~~~~~~~~~~~~~~~~~
include/nic.h:220:49: note: in definition of macro ‘DRIVER’
  220 |                 void ( * _unsafe_disable ) () = _disable;                 \
      |                                                 ^~~~~~~~
[...]

PR #1455, opened 2025-04-26
Fix C23 build, error "unterminated-string-initialization"

h908714124:
In C23, a string literal must fit in its array including its NUL terminator.

In version.c, the comment says there should not be a NUL terminator in SBAT data. But it seems there was one, at least in sbat.h. So either this is an old bug, or the comment should be changed.

Fix typo in xenver.h.

Fixes: #1419

### DIFF
--- a/src/arch/x86/image/ucode.c
+++ b/src/arch/x86/image/ucode.c
@@ -67,7 +67,7 @@ union ucode_vendor_id {
 	/** CPUID registers */
 	uint32_t dword[3];
 	/** Human-readable string */
-	uint8_t string[12];
+	uint8_t string[13];
 };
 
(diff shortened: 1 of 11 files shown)


NiKiZe:
Each commit would need to change one thing, see recent git history for expected separation, and also expecting on commit message contents. The xenver.h fix could do in a separate commit and maybe even PR.

The other changes are questionable, it wouldn't be the first time GCC has bugs that iPXE would have to work around. (Checks that are correct in some cases, but are based on assumptions that aren't correct for iPXE)

h908714124:
The xenver change is now here:
https://github.com/ipxe/ipxe/pull/1456
xenver change was also removed from this PR.

PR #1456, opened 2025-04-26
[xen] Fix typo in include guard

h908714124:


mcb30:
Fixed as part of #1457

PR #1464, opened 2025-05-12
release

mly-johndpope:
Screenshot 2025-05-12 at 8 17 48 pm
https://github.com/johndpope/ipxe/actions/runs/14969680942

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,12 @@
-name: Build
+name: Build and Release
 
-on: push
+on:
+  push:
+    tags:
+      - 'v*'  # Trigger on version tags
+  workflow_dispatch:  # Allow manual triggering
 
 jobs:
-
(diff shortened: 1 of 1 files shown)


mcb30:
Not sure if this is intended for upstream iPXE or not.  If it is, then it's too convoluted for something that can only be tested inside GitHub Actions, sorry.  I deliberately minimise the complexity of the CI `.yml` files, simply because they are such a PITA to test.

mly-johndpope:
just throw code into claude and tell it what you want. probably will work on free tier.
i came to the site because i needed to download the file as im building a dev machine 
- you have free compute in gitactions that can spit out builds - here - people could just come and get the latest release 
- i went to the trouble to cater for every architecture. 
idk.....

please consider.

mly-johndpope:
you will also find it works for all your code too.

PR #1473, opened 2025-05-23
[bnxt] Add a delay when Tx buffers are not available

jw14812:
Add a wait  when buffers are not available for TX. Execute a PCI read to flush pending transactions when the condition is detected. Increase buffer and descriptor counts.

### DIFF
--- a/src/drivers/net/bnxt/bnxt.h
+++ b/src/drivers/net/bnxt/bnxt.h
@@ -145,8 +145,8 @@
 #define DEFAULT_NUMBER_OF_STAT_CTXS             0x01
 #define NUM_RX_BUFFERS                          8
 #define MAX_RX_DESC_CNT                         16
-#define MAX_TX_DESC_CNT                         16
-#define MAX_CQ_DESC_CNT                         64
+#define MAX_TX_DESC_CNT                         64
+#define MAX_CQ_DESC_CNT                         128
 #define TX_RING_BUFFER_SIZE (MAX_TX_DESC_CNT * sizeof(struct tx_bd_short))
 #define RX_RING_BUFFER_SIZE \
 	(MAX_RX_DESC_CNT * sizeof(struct rx_prod_pkt_bd))


src/drivers/net/bnxt/bnxt.c
```diff
@@ -2172,6 +2172,7 @@ static int bnxt_tx ( struct net_device *dev, struct io_buffer *iob )
 
 	if ( bnxt_tx_avail ( bp ) < 1 ) {
 		DBGP ( "- %s (  ): Failed no bd's available\n", __func__ );
+		udelay ( 10000 );
```
mcb30, line 2175: Why is a delay necessary here?  When there are no available transmit descriptors, the routine should either drop the packet, or use `netdev_tx_defer()` to enqueue it for automatic retransmission as soon as the next transmit completion is reported by the driver.

The only devices that need to use `netdev_tx_defer()` are those with very small transmit rings (e.g. RTL8139, where the legacy mode of operation allows for only 4 transmit descriptors).  Most drivers just drop packets, which is typically the correct thing to do: the only time we end up with high transmit volumes is when sending TCP ACKs, and it's generally better to harmlessly drop ACKs than to build up a large transmit queue.

src/drivers/net/bnxt/bnxt.c
```diff
@@ -2265,7 +2271,13 @@ static void bnxt_service_cq ( struct net_device *dev )
...
 			break;
 		default:
-			done = NO_MORE_CQ_BD_TO_SERVICE;
+			udelay ( 1000 );
+			bp->pci_flush_req = 1;
+			if (cq_retry > 5 ) {
+				done = NO_MORE_CQ_BD_TO_SERVICE;
+				break;
+			}
+			cq_retry++;
```
mcb30, line 2280: Again, I'm not sure what the purpose is for this delay.  The usual structure for driver poll routines is to look at the current consumer-indexed entry in the ring, and stop processing if the entry is not yet complete.  The `intel` driver has a clean example of this structure:

```c
void intel_poll_tx ( struct net_device *netdev ) {
	struct intel_nic *intel = netdev->priv;
	struct intel_descriptor *tx;
	unsigned int tx_idx;

	/* Check for completed packets */
	while ( intel->tx.cons != intel->tx.prod ) {

		/* Get next transmit descriptor */
		tx_idx = ( intel->tx.cons % INTEL_NUM_TX_DESC );
		tx = &intel->tx.desc[tx_idx];

		/* Stop if descriptor is still in use */
		if ( ! ( tx->status & cpu_to_le32 ( INTEL_DESC_STATUS_DD ) ) )
			return;

		DBGC2 ( intel, "INTEL %p TX %d complete\n", intel, tx_idx );

		/* Complete TX descriptor */
		netdev_tx_complete_next ( netdev );
		intel->tx.cons++;
	}
}
```

There is no need for any delay or retry mechanism, because the next poll will recheck the entry anyway.

If you are finding that you need some kind of delay and retry mechanism in `bnxt.c`, then this almost certainly means that the polling logic is incorrect.

From a quick look at `bnxt_rx_complete()`, it looks as though it is expecting completions to span two descriptors, and that your delay and retry logic is probably trying to hack around the race condition that arises when the hardware has written only one of the two descriptors, by trying to wait until the second descriptor has also been written.  Don't do this.

Instead, restructure the code so that it can cleanly complete a poll without needing both descriptors to have been written.  I don't know what information the hardware places in each descriptor, but it's likely that you can arrange matters so that either:
* the hardware generates only a single descriptor per packet (e.g. turn off header splitting, or whatever mechanism is causing an extra descriptor to be written), or
* one of the two descriptors contains no useful information and can just be ignored
If there really is no way to avoid needing information from both descriptors, then you still don't need to implement any delay and retry mechanism.  Just exit the poll without advancing the consumer index when you find that the second descriptor has not yet been written, and leave it to the next poll to find both descriptors.

mcb30 (changes requested):
Review comments inline

src/drivers/net/bnxt/bnxt.c
jw14812: Agreed, I have removed these unnecessary delays.

src/drivers/net/bnxt/bnxt.c
jw14812: I have removed these delays and updated the commit to only change the number of descriptors

mcb30:
Thank you for responding to the review comments.  I have rebased, updated the commit date, and merged.

PR #1484, opened 2025-06-26
[bnxt] Remove VLAN stripping logic

jw14812:
Remove logic that programs the hardware to strip out VLAN from RX packets. Do not drop packets due to VLAN mismatch and allow the upper layer to decide whether to discard the packets.


mcb30:
This looks like a fantastic removal, thank you!  Are you planning to also strip out the equivalent redundant functionality on the transmit path?

jw14812:
I will look into addressing the transmit VLAN logic in a separate commit. I need to do some further investigation on it's impacts on VF functionality.  Can this pull request be accepted without changes to the transmit path for now?

mcb30:
OK, makes sense.  I'll queue this up for merging now.

PR #1489, opened 2025-07-01
[bnxt] Update device name and descriptions

jw14812:
Use human readable strings for dev_name and dev_description in PCI_ROM array. Add support for additional device IDs and use an array to identify VF device IDs.


NiKiZe:
Instead of using `bnxt_vf_nics` have you considered using flags see the i219 special handling in intel.c as example.

Maybe split the changes into separate commits?

Edit: I notice that is what we already have, what is your rationale for moving away from flags?

NiKiZe:
As mentioned in my edit, we already have the flags I was requesting, and I don't see any reason to remove that logic and making the device table and logic more complex again.

See https://github.com/ipxe/ipxe/commit/059c4dc688cea6d9398d63ca10f5c00e94ae2159

NiKiZe:
Going one step further, could there be scripts in the wild having logic to match on the driver name, which would break with this change? (I actually have such server logic, but mostly used with devices with Intel cards)
I think this is less likely on the description, and really think that the description should change, but is there considerations here that hasn't been thought of?

mcb30:
I'm happy to accept the descriptions being changed, since those don't (AFAICR) show up in the final binary anyway.

This would potentially be a breaking change for automated builds: target names such as `make bin/14e4-1801.pxe` would no longer be valid.  I agree that the current names are not great (e.g. the mixed case in `14e4-D802`), but we need a strong justification for breaking backwards compatibility.

Like @NiKiZe, I prefer the existing flags to having a separate table of VF device IDs.  What's the motivation for having a separate table?

An even better approach for VF detection would be if there is some way independent of the device ID (e.g. an MMIO feature register that lets us know if we are dealing with a VF, or a PCI capability that is guaranteed to be present/absent only on VFs/PFs.)

jw14812:
Ack.  Thank you for your comments/suggestions.  They are sound and reasonable.  I have reduced the changes to only the device description updates(as well as adding a couple more IDs) and removed the VF detection changes.  If you would still request to split the changes into 2 separate commits(one for the description and one for the additional IDs), I can do that too.

NiKiZe:
For readability of change, I think it makes sense to have  separate commits for adding new lines and to modify the existing ones.

For the new lines I think you should use the better bcm driver names, but I would suggest to use them in lowercase, as seen in most other drivers.

NiKiZe:
_Sorry for the close, unintentional slipp of fingers_

jw14812:
Updated with only the device descriptions modified.

PR #1491, opened 2025-07-02
[bnxt] Update supported devices array

jw14812:
Add support for new device IDs. Remove  support for deprecated device IDs.


mcb30:
What does "deprecated" mean in this context?  Presumably these devices can still be found in existence, so the driver should continue to support them.

jw14812:
Sorry, this was a poor choice of words.  The IDs being removed were never productized.  I will update the description

PR #1492, opened 2025-07-03
[bnxt] Add support for Advanced NPAR mode

jw14812:
Add support to detect if the adapter is operating in Advanced NPAR mode.  Set the relevant flag when registering with the firmware. Reclaim FLAGS bit that was never used.


mcb30:
What is Advanced NPAR mode, and why does iPXE need to know about its existence?

jw14812:
It is a specific multifunction mode supported by some of our adapters.  The driver needs to be aware so that it can properly indicate the necessary support when it registers with the firmware.

mcb30:
I've taken a quick look through the code, and it seems as though the only change in behaviour in NPAR mode is that we bypass the call to `bnxt_get_link_speed()`.

iPXE generally does not need to know the link speed.  We provide no facility for forcing a non-autonegotiated link speed, and we generally rely on the MAC being able to automatically establish its GMII (or equivalent) link configuration with the PHY.

iPXE needs only to know whether the link is up or down (and even that is optional: some drivers just assume the link is always up, and rely on the fact that DHCP discovery will fail if the link is down).  It looks from a quick scan of the code that link up/down detection in the bnxt driver is not dependent on `bnxt_get_link_speed()`.

Can we remove all of the code relating to link speed detection, and consequently remove all knowledge of NPAR from iPXE?  If so, then this would substantially simplify the driver, which is very desirable.

jw14812:
Upon further investigation, these changes are not required.

PR #1501, opened 2025-07-18
[json-params] Add special handling for JSON parameters in http_form_params

refrigerator18:
Previously, all parameters were encoded as form data (key=value&key2=value2
    format with URI encoding). This doesn't work for JSON payloads which need
    to be sent as raw JSON strings.
    
    Added check for PARAMETER_JSON flag:
    - Form parameters: encoded as "key1=value1&key2=value2" with URI encoding
    - JSON parameters: return just the key as-is (key contains the full JSON)
    
    This is a temporary solution - JSON handling should be moved to a separate
    function in the future.

### DIFF
--- a/src/arch/riscv/core/riscv_dma.c
+++ b/src/arch/riscv/core/riscv_dma.c
@@ -34,6 +34,13 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
  *
  */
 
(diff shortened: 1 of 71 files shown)


NiKiZe:
IMO iPXE should not know JSON.
Any JSON backend ought to be reached thru a proxy doing conversion to/from JSON to the data formats iPXE already know.

PR #1507, opened 2025-07-24
[bnxt] Remove unnecessary macros

jw14812:
Remove unnecessary driver specific macros.  Use standard pci_read_config_xxxx, pci_write_config_xxx, writel/q calls.


mcb30:
Thank you for doing this!  I've also pushed https://github.com/ipxe/ipxe/commit/01b1028d4 to remove `test_if`, since it was right next to the code I reviewed here.

PR #1508, opened 2025-08-01
[bnxt] Add error recovery support

jw14812:
Add support to advertise adapter error recovery support to the firmware. Implement error recovery operations if adapter fault is detected.

### DIFF
--- a/src/drivers/net/bnxt/bnxt.c
+++ b/src/drivers/net/bnxt/bnxt.c
@@ -25,6 +25,7 @@ static void bnxt_adv_cq_index ( struct bnxt *bp, u16 cnt );
 static int bnxt_rx_complete ( struct net_device *dev, struct rx_pkt_cmpl *rx );
 void bnxt_link_evt ( struct bnxt *bp, struct hwrm_async_event_cmpl *evt );
 
+
 static struct pci_device_id bnxt_nics[] = {
 	PCI_ROM( 0x14e4, 0x1604, "14e4-1604", "Broadcom BCM957454", 0 ),
 	PCI_ROM( 0x14e4, 0x1605, "14e4-1605", "Broadcom BCM957454 RDMA", 0 ),
@@ -512,12 +513,20 @@ static int bnxt_rx_complete ( struct net_device *dev,
 		return NO_MORE_CQ_BD_TO_SERVICE;
 }
 
-void bnxt_mm_init ( struct bnxt *bp, const char *func )
+void bnxt_mm_init_hwrm ( struct bnxt *bp, const char *func )
 {
 	DBGP ( "%s\n", __func__ );
 	memset ( bp->hwrm_addr_req,  0, REQ_BUFFER_SIZE );
 	memset ( bp->hwrm_addr_resp, 0, RESP_BUFFER_SIZE );
 	memset ( bp->hwrm_addr_dma,  0, DMA_BUFFER_SIZE );
+	bp->hwrm_max_req_len  	= HWRM_MAX_REQ_LEN;
+	bp->hwrm_cmd_timeout 	= HWRM_CMD_DEFAULT_TIMEOUT;
+	dbg_mem ( bp, func );
+}
+
+void bnxt_mm_init_rings ( struct bnxt *bp, const char *func )
+{
+	DBGP ( "%s\n", __func__ );
 	memset ( bp->tx.bd_virt,  0, TX_RING_BUFFER_SIZE );
 	memset ( bp->rx.bd_virt,  0, RX_RING_BUFFER_SIZE );
 	memset ( bp->cq.bd_virt,  0, CQ_RING_BUFFER_SIZE );
@@ -526,7 +535,6 @@ void bnxt_mm_init ( struct bnxt *bp, const char *func )
 	bp->link_status = STATUS_LINK_DOWN;
 	bp->wait_link_timeout = LINK_DEFAULT_TIMEOUT;
 	bp->mtu = MAX_ETHERNET_PACKET_BUFFER_SIZE;
-	bp->hwrm_max_req_len  = HWRM_MAX_REQ_LEN;
 	bp->nq.ring_cnt 	  = MAX_NQ_DESC_CNT;
 	bp->cq.ring_cnt 	  = MAX_CQ_DESC_CNT;
 	bp->tx.ring_cnt 	  = MAX_TX_DESC_CNT;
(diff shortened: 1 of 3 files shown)


src/drivers/net/bnxt/bnxt.c
```diff
@@ -2123,17 +2245,257 @@ static void bnxt_adv_nq_index ( struct bnxt *bp, u16 cnt )
...
+void bnxt_er_task ( struct bnxt* bp, u8 hb_task )
+{
+	u32 present_hb_cnt;
+	unsigned short pci_command, new_command;
+	u8 i;
+
+	DBGP ( "%s\n", __func__ );
+	if ( bp->er.er_rst_on ) {
+		/* Reset already in progress */
+		udelay ( bp->er.normal_wait_period * 100000 );
```
mcb30, line 2365: This function seems to get called by `bnxt_poll()`, which is called in a tight loop by iPXE.  Why does it need to delay for some multiple of 100ms?

src/drivers/net/bnxt/bnxt.c
```diff
@@ -2123,17 +2245,257 @@ static void bnxt_adv_nq_index ( struct bnxt *bp, u16 cnt )
...
+	/* Reconfigure the PCI attributes */
+	pci_write_config_word ( bp->pdev, PCI_COMMAND, pci_command );
+
+	if ( hb_task ) {
+		if ( bp->er.master_pf ) {
+			/* wait for master func wait period */
+			udelay ( bp->er.master_wait_period * 100000 );
+
+			/* Reset register values */
+			bnxt_rst_er_registers ( bp );
```
mcb30, line 2405: Why does the recovery path not simply reset the NIC?  I'm not understanding why we need a separate "recovery reset" mechanism, given that we are presumably already resetting the NIC to a clean state during driver probe (in `bnxt_init_one()`).  Does the standard reset mechanism not work if the firmware has failed?

If the standard reset mechanism does not recover from a firmware failure, then we should also allow for the possibility that `bnxt_init_one()` will need to perform recovery.  Either way: there should be a single code path for "get the NIC into a clean and guaranteed working state", and this single code path should be used by both `bnxt_init_one()` and `bnxt_er_task()`.

mcb30 (changes requested):
Thank you for the modifications.  This looks closer to being mergeable: the duplication from #1505 seems to have been removed, and as far as I can tell the DMA memory is no longer freed and reallocated as part of the firmware error recovery mechanism.

There are some inline comments that need to be addressed: most importantly for `bnxt_init_get_mac()` and `bnxt_er_task()`.

Also: how has this recovery mechanism been tested?  It would be reassuring to see `inject_fault()` used to trigger a firmware failure (or to pretend that the firmware has failed), so that there is some way to test these code paths without waiting for a (presumably rare) firmware failure event.

src/drivers/net/bnxt/bnxt.c
```diff
@@ -2257,6 +2676,37 @@ static struct net_device_operations bnxt_netdev_ops = {
...
 };
 
+int bnxt_init_get_mac ( struct bnxt *bp )
+{
+	int rc = 0;
+
+        bp->hwrm_addr_req  = dma_alloc ( bp->dma, &bp->req_mapping,
+                                         REQ_BUFFER_SIZE, REQ_BUFFER_SIZE );
+        bp->hwrm_addr_resp = dma_alloc ( bp->dma, &bp->resp_mapping,
+                                         RESP_BUFFER_SIZE, RESP_BUFFER_SIZE );
```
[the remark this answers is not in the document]
jw14812: Ack.  I have refactored the memory allocation operations to better align with probe and open functionality. The memory operations have been separated into ring  and communication channel   functions.

src/drivers/net/bnxt/bnxt.c
jw14812: Ack

src/drivers/net/bnxt/bnxt.c
jw14812: Error recovery cannot be handled solely by the driver, as it is not capable of independently resetting the NIC.  If the firmware has failed, part of the error recovery process is to restart the firmware so that it is in a healthy state.  The adapter error recovery feature is intended to work without needing to re-probe/re-open the driver.  This is the main reason why there is a separate recovery path.  The standard probe/open path assumes the firmware is healthy and ready, while the recovery path is meant to restore the NIC to a usable state once the firmware has completed its recovery.

jw14812:
The error recovery feature has been tested using internal test tools to trigger a firmware fault.  I will look into adding a "bogus" fault injection to trigger this code path in a future commit, as it will require some discussions internally on how the fault injection should also mimic the firmware failure.

src/drivers/net/bnxt/bnxt.c
```diff
@@ -2226,27 +2655,44 @@ static void bnxt_service_nq ( struct net_device *dev )
...
 
 static void bnxt_close ( struct net_device *dev )
 {
 	struct bnxt *bp = dev->priv;
 
 	DBGP ( "%s\n", __func__ );
-	bnxt_down_nic (bp);
+	if ( bp->er.driver_initiated_recovery ) {
+		stop_timer ( &bp->timer );
+	}
```
mcb30, line 2691: Minor point, but this can (and probably should) be an unconditional `stop_timer()`, just so that nobody has to worry about whether or not it's possible for the timer to be left running inadvertently.  (It is safe to call `stop_timer()` on an already-stopped timer.)

mcb30:
> The error recovery feature has been tested using internal test tools to trigger a firmware fault. I will look into adding a "bogus" fault injection to trigger this code path in a future commit, as it will require some discussions internally on how the fault injection should also mimic the firmware failure.

Thank you, much appreciated.  I'd much rather have it possible for this driver to be maintained without requiring access to Broadcom internal test tools.

src/drivers/net/bnxt/bnxt.c
jw14812: If the firmware fails prior to iPXE execution, and it has not recovered prior to the driver initialization, the driver would fail to load.  As part of the normal driver initialization, it will query the firmware on error recovery capabilities as well as retrieve error recovery related parameters needed during the "reset" path, if applicable.  The driver must be able to go through a "normal" probe/open at least once in order for it be "error recovery aware".  There are no hardcoded error recovery registers on the part of the driver, as this information is retrieved from the firmware

mcb30: OK, I understand.  This seems like a design flaw in the recovery mechanism: if there's no way for a freshly loaded iPXE or Linux kernel driver (with no saved state) to perform firmware recovery, then that creates a situation in which the only way to proceed is a hard reboot.  But if this hardware flaw does exist then there's nothing we can do about it, so I accept the rationale of keeping recovery separate from normal reset.

src/drivers/net/bnxt/bnxt.c
jw14812: I have added a timer at the start of the error recovery code, and will use this as a means to keep track of the necessary wait time instead of using the long delay time

jw14812: @mcb30 - I have changed the hard delay to use a timer mechanism instead

mcb30:
Rebased (to include your existing change merged in #1509) and pushed, thank you.

It would still be good to have a way to test the recovery mechanism via `config/fault.h` without requiring the use of Broadcom internal tools.  At the moment, there's no way for anyone else (including me) to test it and so there's a good chance it will end up accidentally broken in future.

PR #1511, opened 2025-08-06
Enhancements to digest_cmd.c

keithga:
In a recent [post](https://techcommunity.microsoft.com/blog/hardwaredevcenter/ipxe-security-assurance-review/1062943) on iPXE Secure Boot signing, Microsoft stated that the `imgverify` command must be removed from ipxe SecureBoot submissions due to "Security Implications".

Although I disagree with Microsoft's findings on the `imgverify` command, there may still be the need to verify some level of integrity for files like boot.wim WinPE images.

This pull request makes two changes:

- Adds the command **sha256sum** to digest_cmd.c to allow for sha256 digests in addition to MD5 and sha1
- Add a single argument, allowing the computed digest to be compared with the argument

Example:
```
set myhash 014bafad85f31983aabd19ff000bf2336eb75995b0e7fa40d2d0ae4df80e6503
initrd -n boot.wim https://ftp.hp.com/pub/pcbios/CPR/sources/boot.wim || goto hpbootfail
echo expected hash: ${myhash}
sha256sum boot.wim ||
sha256sum -s ${myhash} boot.wim || goto bad_hash
```

### DIFF
--- a/src/hci/commands/digest_cmd.c
+++ b/src/hci/commands/digest_cmd.c
@@ -23,12 +23,15 @@ FILE_LICENCE ( GPL2_OR_LATER );
 #include <string.h>
 #include <unistd.h>
 #include <getopt.h>
+#include <errno.h>
 #include <ipxe/command.h>
 #include <ipxe/parseopt.h>
 #include <ipxe/image.h>
 #include <ipxe/crypto.h>
 #include <ipxe/md5.h>
 #include <ipxe/sha1.h>
+#include <ipxe/sha256.h>
+#include <ipxe/base16.h>
 #include <usr/imgmgmt.h>
 
 /** @file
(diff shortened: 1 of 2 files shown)


mcb30:
I can accept this use case, with the caveat that I will never accept a patch that uses a checksum result to mark the image as trusted.  Marking images as trusted will remain possible only by using `imgverify`.

There are some other issues with the patch: the digest should be decoded via `hex_decode()` and compared as a binary value, rather than compared as a string, and I'm not keen on `--sum` as the option name since it's not obvious that it implies verification/comparison.  I'll rework it.

mcb30:
I've pushed alternative functionality:
* https://github.com/ipxe/ipxe/commit/6fa901530
* https://github.com/ipxe/ipxe/commit/f45782f9f

with which you can now do e.g.:
```
set expected:hexraw 014bafad85f31983aabd19ff000bf2336eb75995b0e7fa40d2d0ae4df80e6503
initrd -n boot.wim https://ftp.hp.com/pub/pcbios/CPR/sources/boot.wim
sha256sum --set actual boot.wim
echo expected hash: ${expected}
echo actual hash: ${actual}
iseq ${actual} ${expected} || goto bad_hash
```

mcb30:
There is also now documentation at https://ipxe.org/cmd/sha256sum (et al).

keithga:
Yea, having sha256sum set a variable with the digest is better. Thanks!

Issue #1515, opened 2025-08-13
build with gcc15 `-ffat-lto-objects` fails for `arch/x86/interface/pcbios/int13.c`


TheJJ:
hey! this may well be a compiler bug, but maybe it's still something to address in ipxe :)

version 5b3ebf8b24ae40a6f9f9f78491702d508f843e56

gcc version `15.1.0` or `15.2.0-1ubuntu1`

build fails with ` -flto=auto -ffat-lto-objects`
```
x86_64-pc-linux-gnu-gcc -DARCH=i386 -DPLATFORM=pcbios -DPLATFORM_pcbios -DSECUREBOOT=0 -march=i386 -fomit-frame-pointer -fstrength-reduce -falign-jumps=1 -falign-loops=1 -falign-functions=1 -mpreferred-stack-boundary=2 -mregparm=3 -mrtd -freg-struct-return -m32 -fshort-wchar -Ui386 -DNVALGRIND -fno-stack-protector -Iarch/i386/include -Iarch/x86/include -Iinclude -I. -Os -g -ffreestanding -fcommon -Wall -W -Wformat-nonliteral -Wno-array-bounds -Wno-dangling-pointer -fno-dwarf2-cfi-asm -fno-exceptions  -fno-unwind-tables -fno-asynchronous-unwind-tables -Wno-address -Wno-stringop-truncation -Wno-address-of-packed-member -Wno-unterminated-string-initialization -fno-PIE -no-pie  -fcf-protection=none  -ffunction-sections  -include include/compiler.h -DASM_TCHAR='@' -DASM_TCHAR_OPS='@' -Ulinux   -DOBJECT=int13 -c arch/x86/interface/pcbios/int13.c -o bin-i386-pcbios/int13.o -flto=auto -ffat-lto-objects
In file included from arch/x86/include/realmode.h:7,
                 from arch/x86/interface/pcbios/int13.c:40:
arch/x86/include/librm.h:103:9: error: ‘*num_drives’ causes a section type conflict with ‘*rm_ds’ in section ‘.text16.data’
  103 |         _text16_ ## variable __asm__ ( #variable )
      |         ^~~~~~~~
arch/x86/include/librm.h:103:9: note: ‘*rm_ds’ was declared here
  103 |         _text16_ ## variable __asm__ ( #variable )
      |         ^~~~~~~~
```
when the lto flags are removed, the file compiles.

[...]

PR #1521, opened 2025-08-18
[build] allow make variables to be overridden

TheJJ:
CFLAGS, HOST_CFLAGS, ... are often overwritten by packaging tools, and using ?= assignment allows easy re-definition from the outside environment.
https://www.gnu.org/software/make/manual/html_node/Environment.html

By merging this patch upstream, Debian/Ubuntu could drop their patch adjusting the flags.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -3,30 +3,30 @@
 # Initialise various variables
 #
 
-CLEANUP		:=
-CFLAGS		:=
-ASFLAGS		:=
-LDFLAGS		:=
-HOST_CFLAGS	:=
+CLEANUP		?=
+CFLAGS		?=
+ASFLAGS		?=
+LDFLAGS		?=
+HOST_CFLAGS	?=
 MAKEDEPS	:= Makefile
 CROSS_COMPILE	?= $(CROSS)
-SYMBOL_PREFIX	:=
+SYMBOL_PREFIX	?=
 
 ###############################################################################
 #
 # Locations of tools
 #
-HOST_CC		:= gcc
-RM		:= rm -f
-TOUCH		:= touch
-MKDIR		:= mkdir
-CP		:= cp
-ECHO		:= echo
-PRINTF		:= printf
-PERL		:= perl
-PYTHON		:= python
-TRUE		:= true
-TRUNCATE	:= truncate
+HOST_CC		?= gcc
+RM		?= rm -f
+TOUCH		?= touch
+MKDIR		?= mkdir
+CP		?= cp
+ECHO		?= echo
+PRINTF		?= printf
+PERL		?= perl
+PYTHON		?= python
+TRUE		?= true
+TRUNCATE	?= truncate
 CC		= $(CROSS_COMPILE)gcc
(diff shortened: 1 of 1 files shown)


NiKiZe:
https://github.com/ipxe/ipxe/issues/1515#issuecomment-3197597775

mcb30:
The comment at https://github.com/ipxe/ipxe/issues/1515#issuecomment-3197597775 is exactly correct.  iPXE should not be built using distro-specific flags intended for Linux applications, since these will be completely wrong for the bare-metal environment in which iPXE runs.

TheJJ:
Thanks for the feedback, I wasn't aware of that.
Debian/Ubuntu have carried a patch to change the CFLAGS for years... I can try that they get rid of this, but I assume resistance as these flags seen are "just hardening, improvements and defaults".
So far I'm not aware of past problems, and LTO is just enabled on Ubuntu as far as i know.

mcb30:
> Thanks for the feedback, I wasn't aware of that. Debian/Ubuntu have carried a patch to change the CFLAGS for years... I can try that they get rid of this, but I assume resistance as these flags seen are "just hardening, improvements and defaults". So far I'm not aware of past problems, and LTO is just enabled on Ubuntu as far as i know.

Oh dear.  Do you have a link to the patch?

NiKiZe:
> > Thanks for the feedback, I wasn't aware of that. Debian/Ubuntu have carried a patch to change the CFLAGS for years... I can try that they get rid of this, but I assume resistance as these flags seen are "just hardening, improvements and defaults". So far I'm not aware of past problems, and LTO is just enabled on Ubuntu as far as i know.
> 
> Oh dear. Do you have a link to the patch?

This ?
https://git.launchpad.net/ubuntu/+source/ipxe/tree/debian/patches/respect-CPPFLAGS-CFLAGS-LDFLAGS-presets.patch?h=debian/sid

mcb30:
> > Oh dear. Do you have a link to the patch?
> 
> This ? https://git.launchpad.net/ubuntu/+source/ipxe/tree/debian/patches/respect-CPPFLAGS-CFLAGS-LDFLAGS-presets.patch?h=debian/sid

Thanks, that looks as though it must be it.  Yes, that patch looks as though it's likely to break actual functionality for the sake of inhibiting a false positive warning from a blunt automated tool.

@TheJJ please feel free to report this back to Ubuntu.  If there are any issues with their binaries, our first response is going to be "try reproducing in a standard build without the Ubuntu CFLAGS patch".

TheJJ:
Reported in https://bugs.launchpad.net/ubuntu/+source/ipxe/+bug/2121439 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1112220

TheJJ:
We're removing the flag customizations in Ubuntu.
Our security team recommended you to have a look at https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++ if some of the options for hardening are suitable for you :slightly_smiling_face:

PR #1535, opened 2025-09-18
[bnxt] Handle link related async events

jw14812:
Handle async events related to link speed change, link speed config change, and port phy config changes.


mcb30:
All looks good to me, sorry for missing this PR.  Rebased and merged now.

PR #1563, opened 2025-11-17
[bnxt] Use separate DMA address for short commands

jw14812:
Allocate and use a separate DMA buffer to send HWRM short commands.  Update condition where short command should be used.


mcb30:
Why does this require a separate DMA buffer allocation?  As far as I can tell, the buffer is passed only to `hwrm_write_req()`, which then reads the buffer contents and uses `writel()` to write the values to the hardware.  It doesn't look as though `short_cmd_req` ever gets used for DMA?

jw14812:
Agreed with feedback. Closing PR as unnecessary.

PR #1565, opened 2025-11-19
[intelxl] Added support for I225 and I226 chipsets

Bert-Ezendam:
I added the support for the 2.5Gb Ethernet Chips of Intel:
- I225 series
- I226 series
also known as the igc driver set of intel.

By just adding the PCI-codes, I was able to use ipxe.efi to bootstrap my computer which has a I226-it device.
Hopefully you are willing to merge this into the master.

### DIFF
--- a/src/drivers/net/intel.c
+++ b/src/drivers/net/intel.c
@@ -1043,6 +1043,7 @@ static struct pci_device_id intel_nics[] = {
 	PCI_ROM ( 0x8086, 0x0d4f, "i219v-10", "I219-V (10)", INTEL_I219 ),
 	PCI_ROM ( 0x8086, 0x0d53, "i219lm-12", "I219-LM (12)", INTEL_I219 ),
 	PCI_ROM ( 0x8086, 0x0d55, "i219v-12", "I219-V (12)", INTEL_I219 ),
+	PCI_ROM ( 0x8086, 0x0d9f, "i225it", "I225-IT", INTEL_PBSIZE_RST ),       
 	PCI_ROM ( 0x8086, 0x0dc5, "i219lm-23", "I219-LM (23)", INTEL_I219 ),
 	PCI_ROM ( 0x8086, 0x1000, "82542-f", "82542 (Fiber)", 0 ),
 	PCI_ROM ( 0x8086, 0x1001, "82543gc-f", "82543GC (Fiber)", 0 ),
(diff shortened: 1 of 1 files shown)


NiKiZe:
You have tested some of the I225, but not 226?

Please also have a read at https://ipxe.org/admin

Bert-Ezendam:
> Thanks, this will need some changes.
> Keep entries sorted on PCI IDs, and see other entries for naming (the string contents)
> 
> Another good thing is to check existing git commit history for the file you are modifying.
> 
> When updating, to keep a clean history you should modify your existing commit and force push.

What do you mean by keep entries short on PCI IDs? I copied this from the latest linux kernel.
I looked at the commit history. Thats why i started already with [intelxl]. This must be [intelx] which i will correct. I hope that this is what is wrong.
Thanks for the hint to use force push on my commit.

> You have tested some of the I225, but not 226?

I have tested the i226-it. That's the only device i have, So i am not able to test the other ones. Since ipxe uses a bare minimum of the device, it is pritty assumable that the other will work aswell, althow there is no evidence for that. Sould i then just add this one device, and leave the other values as comment, or what? Please advice me on this. I am new to the open source maintenance.

> Please also have a read at https://ipxe.org/admin

I have read these and wil try to correct the commit message for this.

NiKiZe:
Don't use values from the kernel, instead use a pciid database to get friendly name and NIC siblings.

We sort the list of pciids on the pciid, not the names.

See intel with all variants of i219, they are in the pciid order.

I do believe 226 will work as well, but since it isn't tested it might be a good idea to leave it out, because what if it doesn't work and causes issues for a number of people. There is many users of iPXE, so we need to be a bit carefull.

Bert-Ezendam:
I corrected the commit and added only those id's that were mentioned in the pci.ids database. I corrected the commit message.
I may have accidentally hit a wrong button that now tell merge master into master. Sorry for that.

NiKiZe:
You need to rebase and force push only one commit.
`git rebase -i` is jsefull

Bert-Ezendam:
See this when i use git rebase --i: I am lost:
al15591@L-AW24YR:~/ipxe-fork$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   .github/FUNDING.yml
        modified:   .github/workflows/build.yml
        modified:   .github/workflows/coverity.yml

no changes added to commit (use "git add" and/or "git commit -a")

Bert-Ezendam:
I think I solved the problem and rebase the thing correctly.

NiKiZe:
Yes, well done!
I do have reservations about using the Linux kernel as source reference, and I think maybe the "tested on" in the commit message is wrong?

But otherwise I think this is a good candidate to be merged into master. (I do not make the final decision)

Bert-Ezendam:
Modified the commit message. Do you agree?

mcb30:
I've fixed up the whitespace errors, adjusted the commit message, and pushed as commit https://github.com/ipxe/ipxe/commit/d73981aec

Thank you!

PR #1566, opened 2025-11-26
add command to get all available netdevs

ramox0:


mcb30:
The [`ifstat`](https://ipxe.org/cmd/ifstat) command already provides this capability.

PR #1573, opened 2025-12-15
[intel] Add PCI ID for I219-V and -LM 24

NiKiZe:


NiKiZe:
8086:57a1 Tested on LENOVO 21QC00A8MX
Seeing similar as https://github.com/ipxe/ipxe/pull/770#issuecomment-1273648208
firmware NII has 2-3 shorter download of boot.wim than native driver (this patch)

NiKiZe:
Just noticed an odd behavior.
The e1d.inf driver, compatible with win10, does not work when the nic is initialized thru ipxe.efi.
It does work fine if using snponly.efi
e1dn.inf is not tested, since it is win11 only, and current winpe setup is win10 based - will try to revisit.

NiKiZe:
e1n.inf worked in windows 11, no working driver found at the time for windows 10.

PR #1578, opened 2025-12-25
[spcr] Add "spcr" builtin to dump redirection parameters

shenghaoyang:
On x86 the Linux kernel doesn't automatically use the SPCR console
(though it is an option for `earlycon`).

Expose the SPCR-provided console address in a new `spcr` builtin -
pass this to the kernel with:

```
kernel ... console=${spcr}
```

Unfortunately this is a bit ugly because we have to consider x86
I/O ports.

Thanks,

Shenghao

### DIFF
--- a/src/arch/x86/core/x86_io.c
+++ b/src/arch/x86/core/x86_io.c
@@ -32,33 +32,6 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
  *
  */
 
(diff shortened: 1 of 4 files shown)


shenghaoyang:
The kernel gains `acpi=spcr` in 7.0 for x86 systems to use the SPCR provided console - https://patch.msgid.link/20260117072827.355360-1-me@shenghaoyang.info

PR #1581, opened 2025-12-29
[ipv6] MTU setting from NDP

### LINKED ISSUE #1580: IPv6 MTU issue, set from RA AdvLinkMTU
Microsoft unfortunately have some issues with pMTUd
This means that to have working IPv6 to Azure over a HE tunnel the local MTU needs to be lowered.
So local fix is to update radvd.conf with `AdvLinkMTU 1440;`

Not sure if this should be set on only IPv6 packets, or If we could be fine with setting it on link?

NiKiZe:
RA contains MTU setting, this is especially needed in some networks which don't have a a full 1500 MTU link to IPv6 internet. Mostly due to some providers (such as Microsoft Azure) not having a working pMTUd setup.

Fixes #1580

### DIFF
--- a/src/include/ipxe/ndp.h
+++ b/src/include/ipxe/ndp.h
@@ -68,6 +68,19 @@ struct ndp_prefix_information_option {
 /** NDP autonomous address configuration flag */
 #define NDP_PREFIX_AUTONOMOUS 0x40
 
+/** NDP MTU option */
+#define NDP_OPT_MTU 5
+
+/** NDP MTU */
+struct ndp_mtu_option {
+	/** NDP option header */
+	struct ndp_option_header header;
+	/** Reserved */
+	uint16_t reserved;
+	/** MTU */
+	uint32_t mtu;
+} __attribute__ (( packed ));
+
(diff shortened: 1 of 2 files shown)


mcb30:
Updated to move the named setting definition from `SETTING_IP6` to `SETTING_NETDEV` (since that's where the original `mtu` is grouped), plus a couple of minor cleanups.

I've tested with a local `radvd` and it seems to work as expected.  @NiKiZe please confirm that this does what you expect in Azure, and I'll merge it today if so.

NiKiZe:
Sorry for late (in the day) response, purely private issue so had no time during work.
Confirming that the updated version resolves the issue.

Tested from network that is using a HE tunnel and can't do 1500MTU packets that fetches a minimal response behind an Azure Frontdoor instance, without this fix the request results in a timeout (I have not verified number of retransmissions), but with this fix that lowers MTU to 1440, it works fine and responds almost immediately.

This is in practice only an issue if ICMPv6 traffic is blocked anywhere (such as in Azure). If that traffic wasn't blocked, this wouldn't have been needed. /rant

PR #1582, opened 2025-12-30
[tls] Remove current time from client random bytes

Fullmetal5:
As it turn out, the attempt at putting the timestamp in the client hello random bytes has been half broken since it's initial commit in 5da712385e07f0965a3f7548933c2bd3c4f254f6 over 13 years ago due to forgetting to run `htonl` on the value before encoding it. Resulting in the timestamp from little endian machines being unreadable by automated tools like Wireshark.
Before that the timestamp was simply set to 0.

Seeing as there have been 0 bug reports about this in the 13 years since the feature was introduced (that I can find) and current "general consensus" seems to be to no longer send the timestamp, go ahead and remove the broken feature.

### DIFF
--- a/src/include/ipxe/tls.h
+++ b/src/include/ipxe/tls.h
@@ -309,10 +309,8 @@ struct tls_signature_hash_algorithm {
 
 /** TLS client random data */
 struct tls_client_random {
-	/** GMT Unix time */
-	uint32_t gmt_unix_time;
 	/** Random data */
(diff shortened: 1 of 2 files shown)


Fullmetal5:
Friendly reminder/attention bump @mcb30

mcb30:
> Friendly reminder/attention bump @mcb30

Thanks, merged with an updated commit message to fill in some details for the record.

PR #1588, opened 2026-01-12
[prefix] Add i386 support to unlzma.S

jcapik:
With -DI386_SUPPORT added to CFLAGS it is now possible to DEBUG on 386-class computers with image compression enabled. Previously the unlzma.S required 486-class computers due to the bswap instruction in use. The -DI386_SUPPORT replaces the bswap instruction with xchgb and roll.

### DIFF
--- a/src/arch/x86/prefix/unlzma.S
+++ b/src/arch/x86/prefix/unlzma.S
@@ -45,7 +45,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 
 	.section ".note.GNU-stack", "", @progbits
 	.code32
-	.arch i486
+	.arch i386
 	.section ".prefix.lib", "ax", @progbits
 
 #ifdef CODE16
@@ -962,7 +962,9 @@ decompress:
 	ADDR32 lodsb	/* discard initial byte */
 	print_hex_byte %al
 	ADDR32 lodsl
-	bswapl	%eax
+	xchgb	%al, %ah
+	roll	$16, %eax
+	xchgb	%al, %ah
 	print_hex_dword %eax
 	print_character $('\n')
 	movl	%eax, rc_code(%ebp)


mcb30:
Given that using `xchgb/roll/xchgb` adds only 5 bytes to the binary, that this is the only `.S` file that uses `.arch i486`, and that this is not anywhere near a performance-critical path, I'd rather avoid the `#ifdef` completely and just use the i386 instructions unconditionally.  (See https://dox.ipxe.org/ifdef_harmful.html for background on why iPXE tries to avoid `#ifdef` as much as possible.)

Could you please update the PR to remove the `#ifdef`?

Note that you'll almost certainly encounter other functionality within iPXE that breaks on a real i386, since that hasn't been a tested platform for decades.

jcapik:
> Given that using `xchgb/roll/xchgb` adds only 5 bytes to the binary, that this is the only `.S` file that uses `.arch i486`, and that this is not anywhere near a performance-critical path, I'd rather avoid the `#ifdef` completely and just use the i386 instructions unconditionally. (See https://dox.ipxe.org/ifdef_harmful.html for background on why iPXE tries to avoid `#ifdef` as much as possible.)
> 
> Could you please update the PR to remove the `#ifdef`?

done

> Note that you'll almost certainly encounter other functionality within iPXE that breaks on a real i386, since that hasn't been a tested platform for decades.

yeah ... and not only i386 is broken ... i486 with no PCI bus is broken too and a proper PCI detection would be more than just 5 extra bytes. I wanted a minimal interference so that nobody could blame me it increases the size just because of old PCs nobody except a group of retro fans is interested in.

src/arch/x86/prefix/unlzma.S
```diff
@@ -45,7 +45,7 @@ FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
 
 	.section ".note.GNU-stack", "", @progbits
 	.code32
-	.arch i486
+	.arch i386
```
stappersg: For what it is worth:  I can't tell how that change  effects the whole `unlzma.S`, I can't even tell what the impact is on whole `iPXE` executable.

Consider this _"patch got human attention message"_ as a slight hint for putting
the `s/.arch i486/.arch i386/` in a separate commit.  For `git bisect` and `git revert` reasons.

mcb30:
> > Could you please update the PR to remove the `#ifdef`?
> 
> done

Thank you.  The PR currently has two commits: one that introduces the unwanted `#ifdef` and a second one that removes it.  The first commit is not going to be merged.  Could you please rebase onto current master and squash the two commits down to a single commit that just makes the change in the accepted way?

jcapik:
> > > Could you please update the PR to remove the `#ifdef`?
> > 
> > 
> > done
> 
> Thank you. The PR currently has two commits: one that introduces the unwanted `#ifdef` and a second one that removes it. The first commit is not going to be merged. Could you please rebase onto current master and squash the two commits down to a single commit that just makes the change in the accepted way?

Done. Unfortunately I wasn't clever enough and opened the PR with the master branch as the PR source and it required a bit of git magic that resulted in the PR being automatically closed. Fortunately it could be re-opened.

mcb30:
@jcapik I've reworded the commit message and also stripped your copyright line, on the basis that this is a trivial change that does not merit a new assertion of copyright.  Are you happy for this to be merged in this form?

jcapik:
> @jcapik I've reworded the commit message and also stripped your copyright line, on the basis that this is a trivial change that does not merit a new assertion of copyright. Are you happy for this to be merged in this form?

Yes. It's ok. That's why I checked the "allow edits" checkbox. The only reason why I added the copyright line is because I thought it's missing and deleting is easier than adding :] I went through different copyright and licensing trainings and some legal departments are quite strict about that and require the copyright to be extended each time the commit contains anything else than whitespace or indent changes. Our current pipeline doesn't even let us merge to the devel branch if we forget to extend the copyright.

PR #1601, opened 2026-01-19
Work around 386 errata in "popal" (was: [bios] Introduce PCIBIOS_GUARD)

jcapik:
With -DPCIBIOS_GUARD added to CFLAGS an alternative PCI bus detection is performed before the PCI BIOS is called. That allows operation on pre-PCI era computers where the PCI BIOS installation check can cause the system to freeze. However, in rare cases, the alternative PCI bus detection methods may fail to detect the PCI bus on exotic hardware even though the PCI BIOS installation check is supported and works correctly. There is no easy solution for that, as iPXE has no room for various hardware-specific quirks providing dirty workarounds.

### DIFF
--- a/src/arch/x86/transitions/librm.S
+++ b/src/arch/x86/transitions/librm.S
@@ -1103,9 +1103,25 @@ vc_rmode:
 	popal
 	/* popal skips %esp.  We therefore want to do "movl -20(%sp),
 	 * %esp", but -20(%sp) is not a valid 80386 expression.
-	 * Fortunately, prot_to_real() zeroes the high word of %esp, so
-	 * we can just use -20(%esp) instead.
+	 *
+	 * In theory, the high word of %esp is already zero at this
+	 * point (since prot_to_real() should set it to zero), and the
+	 * popal instruction does not load %esp from the saved
+	 * register dump.  This would allow us to just use -20(%esp)
+	 * instead.
+	 *
+	 * However, some 386 chips are observed to have an
+	 * undocumented errata that causes the popal instruction to
+	 * load the high 16 bits of %esp.  We therefore explicitly
+	 * zero-extend %sp to %esp to work around this errata.
+	 *
+	 * Inserting this instruction also happens to work around
+	 * another (known and documented) errata in the 386, in which
+	 * the CPU may malfunction if popal is followed immediately by
+	 * an instruction that uses a base address register to form an
+	 * effective address.
 	 */
+	movzwl	%sp, %esp
 	addr32 movl -20(%esp), %esp
 	popfl
 	popw	%ss /* padding */


NiKiZe:
As already mentioned in https://github.com/ipxe/ipxe/pull/1588#issuecomment-3754306923 the `#ifdef` is a no go.
Some other things can be read at https://ipxe.org/admin

Implementing this as a separate pcibus implementation that you can switch to is probably the better approach or as a separate pciguard check with a default implementation that returns ok.

jcapik:
> As already mentioned in [#1588 (comment)](https://github.com/ipxe/ipxe/pull/1588#issuecomment-3754306923) the `#ifdef` is a no go. Some other things can be read at https://ipxe.org/admin
> 
> Implementing this as a separate pcibus implementation that you can switch to is probably the better approach or as a separate pciguard check with a default implementation that returns ok.

I've been looking at that and managed to create two modules providing the same function (one dummy always returning True and the other really doing the detection) and it builds, however ... now I don't know how to enforce usage of the right one. It would be cool if the linker tables had something like application notes or be more documented. I was thinking about creating two ipxe.dsk images ... one with the dummy linked in and the other with the detection. Any hints where to start looking?

jcapik:
I was thinking again and maybe #ifdef around the PROVIDE_PCIAPI is acceptable since the code still gets compiled, just not linked? I'll try that and modify the patch accordingly. Hopefully that's what you meant.

mcb30:
> I was thinking again and maybe #ifdef around the PROVIDE_PCIAPI is acceptable since the code still gets compiled, just not linked?

100% not acceptable, sorry.  Please assume by default that any introduction of new `#ifdef`s is going to result in an automatic rejection of any PR for any new feature, or even for a bugfix.  I spent literally years cleaning all of the random `#ifdef` mess out of what was then the Etherboot codebase, and I am not ever going to accept letting one back in.

I would suggest finding out why the PCI BIOS installation check is causing your system to freeze, and fixing the check so that it no longer causes a problem.  If you can't find a way to fix the check, then you can instead select `PCIAPI_DIRECT` in your `config/local/ioapi.h` to avoid using the PCI BIOS code.

jcapik:
> > I was thinking again and maybe #ifdef around the PROVIDE_PCIAPI is acceptable since the code still gets compiled, just not linked?
> 
> 100% not acceptable, sorry. Please assume by default that any introduction of new `#ifdef`s is going to result in an automatic rejection of any PR for any new feature, or even for a bugfix. I spent literally years cleaning all of the random `#ifdef` mess out of what was then the Etherboot codebase, and I am not ever going to accept letting one back in.

I understand that. Hopefully it doesn't look like I'm pushy or something. I just feel my knowledge of the project is not good enough yet and any hints/guidance would be cool.

> I would suggest finding out why the PCI BIOS installation check is causing your system to freeze, and fixing the check so that it no longer causes a problem. 

I did a bit of research and know the reason ... the problem is not in the installation check implementation on the ipxe side. It seems the pre-PCI era BIOSes just freeze when you call INT 1Ah with unsupported function. We tested that on three different computers and all of them got frozen.

> If you can't find a way to fix the check, then you can instead select `PCIAPI_DIRECT` in your `config/local/ioapi.h` to avoid using the PCI BIOS code.

Well, the pcidirect implementation seems to be less featured than PCIBIOS as it only supports one PCI bus and therefore the combination of the guard with the pcibios looks like a bit more robust solution to me since it can handle PCs with PCI-to-PCI bridges which are fairly common.
So ... please, what is the best option to tell the linker what variant of the pcibios_safe function to pick depending on the configuration?

mcb30:
> > I would suggest finding out why the PCI BIOS installation check is causing your system to freeze, and fixing the check so that it no longer causes a problem.
> 
> I did a bit of research and know the reason ... the problem is not in the installation check implementation on the ipxe side. It seems the pre-PCI era BIOSes just freeze when you call INT 1Ah with unsupported function. We tested that on three different computers and all of them got frozen.

Interesting!  Can you get a dump of the BIOS code (e.g. from `dd if=/dev/mem of=bios.bin bs=1M count=1` to get the low megabyte of physical memory)?  The BIOS really should be just returning with CF set for an unknown INT 1A call, and it would be interesting to trace the disassembly to see where it goes wrong, and if there's a way to work around it.

> > If you can't find a way to fix the check, then you can instead select `PCIAPI_DIRECT` in your `config/local/ioapi.h` to avoid using the PCI BIOS code.
> 
> Well, the pcidirect implementation seems to be less featured than PCIBIOS as it only supports one PCI bus and therefore the combination of the guard with the pcibios looks like a bit more robust solution to me since it can handle PCs with PCI-to-PCI bridges which are fairly common. So ... please, what is the best option to tell the linker what variant of the pcibios_safe function to pick depending on the configuration?

`PCIAPI_DIRECT` does cope with bridges.  There is no way to read the maximum number of PCI buses, so we start by assuming a single bus and then expand the range each time a bridge is encountered: https://github.com/ipxe/ipxe/blob/d0ea2b1bb8f78b219f74424d435b92ff8aa0ea8d/src/drivers/bus/pci.c#L348-L363

This does assume that something else has already enumerated the bus once: we don't have any logic to go through assigning the bus resources ourselves.  I will be looking at adding this at some point to cope with RISC-V SBI systems, but I don't have any timeline for that because adding RISC-V support is something I'm doing solely for fun.

jcapik:
> Interesting! Can you get a dump of the BIOS code (e.g. from `dd if=/dev/mem of=bios.bin bs=1M count=1` to get the low megabyte of physical memory)? The BIOS really should be just returning with CF set for an unknown INT 1A call, and it would be interesting to trace the disassembly to see where it goes wrong, and if there's a way to work around it.

Well. It seem it won't be necessary, because I was completely wrong with my assumption the freeze happens in the INT 1A code ... For some reason I considered the flatten_real_mode call innocent and that was a mistake. After playing with debug prints and splitting the __asm__ snippet to two parts I discovered the freeze happens in the flatten_real_mode call. If I remove it, the PCIBIOS installation check works like expected and returns zero.

Please, accept my apology for not being pedantic enough.

So, now we need to find out what's wrong with the flatten_real_mode call and of course the whole PR is a piece of garbage now.

> `PCIAPI_DIRECT` does cope with bridges. There is no way to read the maximum number of PCI buses, so we start by assuming a single bus and then expand the range each time a bridge is encountered:

And that was my second mistake :) For some reason I thought the direct variant works the same way as PCIBIOS and is finished when we leave the discover function. Of course I missed the comment preceeding the range settings, where the bridge detection is mentioned.

mcb30:
> Well. It seem it won't be necessary, because I was completely wrong with my assumption the freeze happens in the INT 1A code ... For some reason I considered the flatten_real_mode call innocent and that was a mistake. After playing with debug prints and splitting the **asm** snippet to two parts I discovered the freeze happens in the flatten_real_mode call. If I remove it, the PCIBIOS installation check works like expected and returns zero.

Great find!  The PCI BIOS probe is potentially the first point that we attempt to use flat real mode.  (We typically use genuine real mode: the use of flat real mode here is to work around a bug in some HP BIOSes as described in commit https://github.com/ipxe/ipxe/commit/e46154fbb.)

The `flatten_real_mode` code is fairly trivial: https://github.com/ipxe/ipxe/blob/641ea020f1b10a4317f758a0b73e3c8fc00f8914/src/arch/x86/transitions/librm.S#L1346-L1364

You could try tweaking it to find out what your CPU dislikes by e.g. changing the `0x8f` values to `0x00` (which is the value they will have already, so this should make the whole call basically a no-op).

mcb30:
In case it's not clear how `flatten_real_mode` is supposed to work: it gets entered in real mode (via the call to `prot_to_real` generated by the `REAL_CODE()` macro).  It overwrites the limit values for the real-mode segments in the GDT, then performs a no-op call to the protected-mode function `flatten_dummy`.  The `virtcall` macro expands to call the `virt_call` function, which will call `real_to_prot` to switch up to protected mode, call the specified payload (`flatten_dummy`, which just returns immediately in this case) and then `prot_to_real` to switch back down to real mode.  The switch back down to real mode loads the real-mode segments from the GDT.  Since we have previously overwritten the limit values for these segments, we end up with 4GB segment limits.  Finally, we restore the limits in the GDT so that the **next** transition to real mode will end up with the standard 64kB segment limits.

jcapik:
> You could try tweaking it to find out what your CPU dislikes by e.g. changing the `0x8f` values to `0x00` (which is the value they will have already, so this should make the whole call basically a no-op).

Even with zeros it gets frozen. So, there must be a problem with the virtcall.

I'm just curious ... does the code always expect FPU present? I see there some comments related to FPU preservation and since 386 / 486SX CPUs have no intergrated FPU, it could be the source of troubles.

mcb30:
> Even with zeros it gets frozen. So, there must be a problem with the virtcall.
> 
> I'm just curious ... does the code always expect FPU present? I see there some comments related to FPU preservation and since 386 / 486SX CPUs have no intergrated FPU, it could be the source of troubles.

We do test for whether or not FXSAVE is supported.  It's plausible that the test is not complete and relies on some assumptions that don't hold on a real 386SX.

However, we know that `virtcall` must already have succeeded at least once, since it gets used to make the initial call into `main()`.  If the whole `virtcall` mechanism were broken then you'd never get as far as attempting a PCI BIOS installation check in the first place.

mcb30:
> However, we know that `virtcall` must already have succeeded at least once, since it gets used to make the initial call into `main()`. If the whole `virtcall` mechanism were broken then you'd never get as far as attempting a PCI BIOS installation check in the first place.

Thinking in more detail about this: we know that the **first half** of ` virtcall` has already succeeded, since we managed to reach `main()` successfully.  It is possible that the PCI BIOS installation check is the first point where we exercise the **return half** of `virtcall`, i.e. the transition back down to real mode.  I'd suggest digging around there.

mcb30:
There's an old errata for the `popal` instruction (aka `popad`) described in the "Intel386™ DX PROCESSOR
SPECIFICATION UPDATE" (copy at https://kib.kiev.ua/x86docs/Intel/486/272874-002.pdf).  It suggests inserting a `nop` immediately after the `popal` instruction (which is used on the return path from `virtcall`): https://github.com/ipxe/ipxe/blob/b7e7f62b87dcab7f4feed3b77352d057723b134c/src/arch/x86/transitions/librm.S#L1097-L1111

It's worth a try to see if it has any effect on your 386.

jcapik:
> There's an old errata for the `popal` instruction (aka `popad`) described in the "Intel386™ DX PROCESSOR SPECIFICATION UPDATE" (copy at https://kib.kiev.ua/x86docs/Intel/486/272874-002.pdf). It suggests inserting a `nop` immediately after the `popal` instruction (which is used on the return path from `virtcall`):
> 
> https://github.com/ipxe/ipxe/blob/b7e7f62b87dcab7f4feed3b77352d057723b134c/src/arch/x86/transitions/librm.S#L1097-L1111
> 
> It's worth a try to see if it has any effect on your 386.

Tested the following and still freezing ...
```
        popw    %es
        popw    %fs
        popw    %gs
        popal
        nop
        /* popal skips %esp.  We therefore want to do "movl -20(%sp),
         * %esp", but -20(%sp) is not a valid 80386 expression.
         * Fortunately, prot_to_real() zeroes the high word of %esp, so
         * we can just use -20(%esp) instead.
         */
        addr32 movl -20(%esp), %esp
        popfl
        popw    %ss /* padding */

```

mcb30:
Thanks for testing.  I've realised that your system must already be surviving the return half of `virt_call`, since we use this during the initial installation of iPXE into high memory (within `init_librm`), which happens before the call to `main()`.

I don't have access to your hardware, so you'll have to do rest of the digging yourself.  Assuming that it's old enough to have a real physical serial port, then you could try using a three-instruction sequence to write a single byte to the serial port, e.g.:
```
movb $42, %al
movw $0x3f8, %dx
outb %al, %dx
```
You can drop that instruction sequence in basically anywhere.  It works whether the CPU is in real mode, protected mode, or even halfway in between, when almost everything else is completely unusable.  If the instruction sequence gets reached, then you should see an asterisk appear on the serial port.  Don't expect execution to continue properly after the asterisk is printed (since %ax and %dx will have been corrupted), and be careful of attempting to preserve them on the stack and restore afterwards (since at some points during the mode transition there is no functional stack pointer).

Before trying this approach, you should enable `CONSOLE_SERIAL` (to initialise the serial port), and patch `ns16550_transmit()` in `drivers/uart/ns16550.c` to duplicate the wait loop after the call to `ns16550_write()`.  That way you don't risk the asterisk being lost due to the transmit register being full with a previously written character.  Check that the iPXE startup banner appears as expected on the serial port after making these changes, then start trying the above three-instruction sequence at various places, subject to the listed caveats.

jcapik:
> Thanks for testing. I've realised that your system must already be surviving the return half of `virt_call`, since we use this during the initial installation of iPXE into high memory (within `init_librm`), which happens before the call to `main()`.
> 
> I don't have access to your hardware,

I can grant you an access via KVM-over-IP, that supports VNC protocol.

mcb30:
> > Thanks for testing. I've realised that your system must already be surviving the return half of `virt_call`, since we use this during the initial installation of iPXE into high memory (within `init_librm`), which happens before the call to `main()`.
> > I don't have access to your hardware,
> 
> I can grant you an access via KVM-over-IP, that supports VNC protocol.

Sorry, that probably wouldn't be sufficient (I'd most likely need remote media and a real (non-virtual) serial console, not just an IP-KVM), plus I can't really justify spending much time on this.  But it's interesting so I'm happy to point you in sensible directions, as in https://github.com/ipxe/ipxe/pull/1601#issuecomment-3813375181

jcapik:
> remote media and a real (non-virtual) serial console

There's a linux distro installed on the box for floppy writes and a serial console from different linux machine (including remote reset) is also doable. However, iterating with reboots is slow and I understand it's your time. Of course I could try the above too, but then we'd have to iterate through messages here. Not sure what's better.

mcb30:
> There's a linux distro installed on the box for floppy writes and a serial console from different linux machine (including remote reset) is also doable. However, iterating with reboots is slow and I understand it's your time. Of course I could try the above too, but then we'd have to iterate through messages here. Not sure what's better.

I think you've got enough in https://github.com/ipxe/ipxe/pull/1601#issuecomment-3813375181 to track down which instruction sequence is causing the problems, and I'm very happy to respond here with further ideas once you've got that far.  Good luck!

jcapik:
Ok, I finally found some time for troubleshooting and it seems the malicious behavior appears in the following snippet ... any chance you know why?

```
/* popal skips %esp.  We therefore want to do "movl -20(%sp),
 * %esp", but -20(%sp) is not a valid 80386 expression.
 * Fortunately, prot_to_real() zeroes the high word of %esp, so
 * we can just use -20(%esp) instead.
 */
addr32 movl -20(%esp), %esp
popfl
popw    %ss /* padding */

```

mcb30:
> Ok, I finally found some time for troubleshooting and it seems the malicious behavior appears in the following snippet ... any chance you know why?
> 
> ```
> /* popal skips %esp.  We therefore want to do "movl -20(%sp),
>  * %esp", but -20(%sp) is not a valid 80386 expression.
>  * Fortunately, prot_to_real() zeroes the high word of %esp, so
>  * we can just use -20(%esp) instead.
>  */
> addr32 movl -20(%esp), %esp
> popfl
> popw    %ss /* padding */
> ```

Great work narrowing it down!  Some things that you could try to narrow it down further:

* Comment out the `addr32 movl -20(%esp), %esp` instruction completely.  This is done only to restore the high 16 bits of `%esp` for the sake of completeness.  For all calls to `virt_call` up to and including the point of PCI enumeration, the high 16 bits of `%esp` should be zero anyway, so this instruction is redundant for your purposes.

* Change the `popw %ss /* padding */` instruction **and** its matching `pushw %ss /* padding */` instruction (at the start of `virt_call`) to something that doesn't affect the `%ss` register.  The constraint is that any instruction that replaces `popw %ss` must add two bytes to the stack pointer while affecting no registers and no flags.  This is a difficult constraint to meet, which is why the `pushw %ss / popw %ss` pair was chosen.  (`%cs` and `%ss` are the only two registers that cannot meaningfully have been modified by whatever protected-mode function was called, and `popw %cs` is not a valid instruction.)  For the sake of experimentation, you could change the two instructions to `pushw %gs / popw %gs`: I'm fairly sure we don't have any protected-mode calls that deliberately modify the value in `%gs`.

jcapik:
Ok ... commenting the `addr32 movl -20(%esp), %esp` out helped and the execution continued till the next hang in the keyboard read ... so, before I start troubleshooting that (hopefully) last blocker, we probably need answers for that %esp magic removal ... 1.) why does it help? 2.) Why does it work without problems on younger hardware? 3.) Is that a standard construct? 4.) Can we just remove it without affecting the execution on other hardware?

mcb30:
> Ok ... commenting the `addr32 movl -20(%esp), %esp` out helped and the execution continued till the next hang in the keyboard read ... so, before I start troubleshooting that (hopefully) last blocker, we probably need answers for that %esp magic removal ...
> 1.) why does it help?
> 2.) Why does it work without problems on younger hardware?
> 3.) Is that a standard construct?
> 4.) Can we just remove it without affecting the execution on other hardware?

The purpose of that code is to restore the high 16 bits of `%esp`.  The call to `virt_call` is made with a 16-bit stack segment so only the low 16 bits (`%sp`) are necessarily valid.  The high 16 bits could potentially hold any arbitrary value.  The design of `virt_call` is to preserve all registers and flags unless explicitly modified by the protected-mode function (by updating fields in the `struct i386_all_regs` passed as a parameter to the protected-mode function).

The transition back down to real mode (via `real_to_prot`) will always produce a stack pointer that is valid but that has had the high 16 bits cleared to zero.  We therefore need some way of restoring the original (and potentially arbitrary) high 16 bits, which is what this instruction does.

There were differences in **very** old chips (pre-286) in the behaviour of the standalone `pushw %sp` instruction: everything from the 286 onwards will push the value of the stack pointer as at *before* the instruction, but the older processors would push the value of the stack pointer as at *after* the instruction (i.e. already decremented by two).  On those very old processors, the sequence `pushw %sp ; popw %sp` would therefore **not** leave `%sp` effectively unchanged.

I am not aware of any equivalent behavioural differences with the `pushal` instruction.  However, it's worth investigating to see what values we actually get.  You can test this quite easily with the following hack patch:

```diff
--- a/src/arch/x86/transitions/librm.S
+++ b/src/arch/x86/transitions/librm.S
@@ -1351,7 +1351,8 @@ flatten_real_mode:
        movb    $0x8f, real_cs + 6
        movb    $0x8f, real_ds + 6
        /* Call dummy protected-mode function */
-       virtcall flatten_dummy
+       movl    %esp, %edx
+       virtcall _dump_regs // flatten_dummy
        /* Restore GDT */
        movb    $0x00, real_cs + 6
        movb    $0x00, real_ds + 6
```

We **should** see that the value of `%esp` has the high 16 bits clear and is precisely 12 (0xc) less than the value of `%edx`, as in this example:

```
iPXE initialising devices...
EAX=0000b101 EBX=0001f184 ECX=0001c440 EDX=00002f4c
ESI=0041f188 EDI=00000000 EBP=0001f184 ESP=00002f40
CS=9c20 SS=9cc8 DS=9cc8 ES=9cc8 FS=9cc8 GS=9cc8
```

jcapik:
The following is from libvirt/kvm ... 
image
... and the following from the real HW where the %esp magic fails ...  
image
and where the high word seems to be 004d and not zero ...

mcb30:
> ... and the following from the real HW where the %esp magic fails ...
> image
> and where the high word seems to be 004d and not zero ...

This is progress!  So, we have an interesting situation, because at that point there is no way that the upper 16 bits should be able to have any value other than zero.

Other than push/pop/call/ret, there are only a handful of instructions that could possibly be affecting the value we see in `%esp` at that point:

* The `movzwl %bp, %esp` at the end of `prot_to_real`, which is where the real-mode stack pointer is first loaded.  This is definitely loading the 16 bits correctly (since we see the same value `0x30a0` for both the libvirt/kvm and the 386 hardware cases).  It is plausible that there exists an unknown hardware errata on the 386 that causes `movzwl` to behave incorrectly when the destination register is `%esp`.

* The `popal` in the middle of `real_call`, just before the call to the real-mode function (i.e. to `flatten_real_mode`, in this case).  This instruction is documented to skip loading `%esp`.  The value of `%esp` present in that register dump is the protected-mode `%esp` (which we definitely do not want to restore, since it has no meaning in real mode).  We know that `popal` cannot be loading the low 16 bits of `%esp` (since we see the same value `0x30a0` for both the libvirt/kvm and the 386 hardware cases).  It is plausible that there exists an unknown hardware errata on the 386 that causes `popal` to load only the high 16 bits of `%esp`.

Both of these are relatively easy to test by adding further instructions to the temporary hack.  Here's a test to see if `movzwl %bp, %esp` is misbehaving:

```diff
--- a/src/arch/x86/transitions/librm.S
+++ b/src/arch/x86/transitions/librm.S
@@ -1351,7 +1351,14 @@ flatten_real_mode:
 	movb	$0x8f, real_cs + 6
 	movb	$0x8f, real_ds + 6
 	/* Call dummy protected-mode function */
-	virtcall flatten_dummy
+	pushl	%ebp
+	movl	$0x42420000, %ebp
+	movw	%sp, %bp
+	movl	%ebp, %ebx	// record test %ebp value in %ebx
+	movzwl	%bp, %esp	// test movzwl %bp, %esp
+	movl	%esp, %edx	// record resulting %esp value in %edx
+	popl	%ebp
+	virtcall _dump_regs
 	/* Restore GDT */
 	movb	$0x00, real_cs + 6
 	movb	$0x00, real_ds + 6
```

mcb30:
> Both of these are relatively easy to test by adding further instructions to the temporary hack. Here's a test to see if `movzwl %bp, %esp` is misbehaving:
> 
> ```diff
> --- a/src/arch/x86/transitions/librm.S
> +++ b/src/arch/x86/transitions/librm.S
> @@ -1351,7 +1351,14 @@ flatten_real_mode:
>  	movb	$0x8f, real_cs + 6
>  	movb	$0x8f, real_ds + 6
>  	/* Call dummy protected-mode function */
> -	virtcall flatten_dummy
> +	pushl	%ebp
> +	movl	$0x42420000, %ebp
> +	movw	%sp, %bp
> +	movl	%ebp, %ebx	// record test %ebp value in %ebx
> +	movzwl	%bp, %esp	// test movzwl %bp, %esp
> +	movl	%esp, %edx	// record resulting %esp value in %edx
> +	popl	%ebp
> +	virtcall _dump_regs
>  	/* Restore GDT */
>  	movb	$0x00, real_cs + 6
>  	movb	$0x00, real_ds + 6
> ```

And here's a test to see if `popal` is misbehaving:

```diff
--- a/src/arch/x86/transitions/librm.S
+++ b/src/arch/x86/transitions/librm.S
@@ -1351,7 +1351,14 @@ flatten_real_mode:
 	movb	$0x8f, real_cs + 6
 	movb	$0x8f, real_ds + 6
 	/* Call dummy protected-mode function */
-	virtcall flatten_dummy
+	movl	%esp, %ebx	// record original %esp in %ebx
+	pushal			// push with original %esp
+	movl	%esp, %edx	// explicitly clear high 16 bits of %esp
+	andl	$0xffff, %edx
+	movl	%edx, %esp
+	popal			// restore (which should skip restoring %esp)
+	movl	%esp, %edx	// record resulting %esp value in %edx
+	virtcall _dump_regs
 	/* Restore GDT */
 	movb	$0x00, real_cs + 6
 	movb	$0x00, real_ds + 6
```

jcapik:
The first test
===========
From libvirt/kvm:
image
and from real HW:
image

jcapik:
The second test
=============
From libvirt/kvm:
image
and from real HW:
image

jcapik:
And one old photo from my first tries to debug where hgw77a8 noticed a strange offset between .text16 and .data16
image

mcb30:
> # The first test
> and from real HW: image

This test demonstrates that `movzwl %bp, %esp` is working as expected on the 386.

> # The second test
> 
> and from real HW: image

We have a culprit!  This test demonstrates that `popal` is **not** working as expected.  We know that the high bits of `%esp` were explicitly cleared between the `pushal` and the `popal`, so the `popal` must have loaded these high bits of `%esp`, contrary to the documented behaviour for the instruction.

mcb30:
Could you please run one (hopefully final) experiment:

```diff
--- a/src/arch/x86/transitions/librm.S
+++ b/src/arch/x86/transitions/librm.S
@@ -1351,7 +1351,14 @@ flatten_real_mode:
 	movb	$0x8f, real_cs + 6
 	movb	$0x8f, real_ds + 6
 	/* Call dummy protected-mode function */
-	virtcall flatten_dummy
+	movl	%esp, %ebx
+	pushal
+	movw	%sp, %bp
+	movl	$0x12345678, 12(%bp)
+	popal
+	movl	%esp, %edx
+	movl	%ebx, %esp
+	virtcall _dump_regs
 	/* Restore GDT */
 	movb	$0x00, real_cs + 6
 	movb	$0x00, real_ds + 6
```

This code deliberately overwrites the stored value for `%esp` created by `pushal` with a garbage value `0x12345678` before executing `popal`.  The value that `popal` loads into `%esp` is then stored into `%edx`, and the correct value of `%esp` is restored (from `%ebx`) before continuing.

Given the behaviour so far, I'm expecting that we will see `%edx` ending up containing a value such as `0x123430ac` (i.e. high 16 bits erroneously restored by `popal`, low 16 bits left untouched).

mcb30:
> Could you please run one (hopefully final) experiment:
> 
> ```diff
> --- a/src/arch/x86/transitions/librm.S
> +++ b/src/arch/x86/transitions/librm.S
> @@ -1351,7 +1351,14 @@ flatten_real_mode:
>  	movb	$0x8f, real_cs + 6
>  	movb	$0x8f, real_ds + 6
>  	/* Call dummy protected-mode function */
> -	virtcall flatten_dummy
> +	movl	%esp, %ebx
> +	pushal
> +	movw	%sp, %bp
> +	movl	$0x12345678, 12(%bp)
> +	popal
> +	movl	%esp, %edx
> +	movl	%ebx, %esp
> +	virtcall _dump_regs
>  	/* Restore GDT */
>  	movb	$0x00, real_cs + 6
>  	movb	$0x00, real_ds + 6
> ```
> 
> This code deliberately overwrites the stored value for `%esp` created by `pushal` with a garbage value `0x12345678` before executing `popal`. The value that `popal` loads into `%esp` is then stored into `%edx`, and the correct value of `%esp` is restored (from `%ebx`) before continuing.
> 
> Given the behaviour so far, I'm expecting that we will see `%edx` ending up containing a value such as `0x123430ac` (i.e. high 16 bits erroneously restored by `popal`, low 16 bits left untouched).

And if that does what I expect, then one possible fix is to add an `addr32` prefix to the `popal` instruction (having first ensured that the full value of `%esp` forms a valid stack pointer):

```diff
--- a/src/arch/x86/transitions/librm.S
+++ b/src/arch/x86/transitions/librm.S
@@ -1351,7 +1351,15 @@ flatten_real_mode:
 	movb	$0x8f, real_cs + 6
 	movb	$0x8f, real_ds + 6
 	/* Call dummy protected-mode function */
-	virtcall flatten_dummy
+	movl	%esp, %ebx
+	pushal
+	movw	%sp, %bp
+	movl	$0x12345678, 12(%bp)
+	movzwl	%bp, %esp
+	addr32 popal
+	movl	%esp, %edx
+	movl	%ebx, %esp
+	virtcall _dump_regs
 	/* Restore GDT */
 	movb	$0x00, real_cs + 6
 	movb	$0x00, real_ds + 6
```

There's no need to provide results for either of these in KVM: I've run the tests in KVM already before handing over the patches.  Only the real hardware needed.

jcapik:
> There's no need to provide results for either of these in KVM: I've run the tests in KVM already before handing over the patches. Only the real hardware needed.

That was just to double-check I applied the patches correctly ... 

1st:
image

2nd:
image

The EDX values seem to be the same -> the addr32 had no effect.

mcb30:
> 1st: image

Excellent!  This gives us unambiguous confirmation that `popal` is loading the high 16 bits of `%esp`, contrary to the documented behaviour.

> 2nd: image
> 
> The EDX values seem to be the same -> the addr32 had no effect.

OK, this also makes sense.  The documentation for `popal` doesn't explicitly mention an effect of the address size prefix, but the documentation for the single-register `pop` does state that "the address size is used only when writing to a destination operand in memory" and that there is a separate concept of "stack-address size": "the B flag in the current stack-segment descriptor determines the size of the stack pointer (16 or 32 bits)".  So the fact that the `addr32` has no effect is at least consistent with this documentation.

I think at this point we can fully characterise the hardware errata as this: when the stack-address size is 16 bits and the operand size is 32 bits, the `popal` instruction will erroneously load the high 16 bits of `%esp` from the value stored on the stack.

mcb30:
> I think at this point we can fully characterise the hardware errata as this: when the stack-address size is 16 bits and the operand size is 32 bits, the `popal` instruction will erroneously load the high 16 bits of `%esp` from the value stored on the stack.

Out of interest: what precise chip is it that has this behaviour?  Is it a real Intel 386 or a clone?

mcb30:
In terms of a potential fix, could you please try:

```diff
--- a/src/arch/x86/transitions/librm.S
+++ b/src/arch/x86/transitions/librm.S
@@ -1106,6 +1106,7 @@ vc_rmode:
 	 * Fortunately, prot_to_real() zeroes the high word of %esp, so
 	 * we can just use -20(%esp) instead.
 	 */
+	movzwl	%sp, %esp
 	addr32 movl -20(%esp), %esp
 	popfl
 	popw	%ss /* padding */
```

jcapik:
> > I think at this point we can fully characterise the hardware errata as this: when the stack-address size is 16 bits and the operand size is 32 bits, the `popal` instruction will erroneously load the high 16 bits of `%esp` from the value stored on the stack.
> 
> Out of interest: what precise chip is it that has this behaviour? Is it a real Intel 386 or a clone?

These two ... and the third was from a remote tester
image

The latest tests were done on the Intel.

mcb30:
> > Out of interest: what precise chip is it that has this behaviour? Is it a real Intel 386 or a clone?
> 
> These two ... and the third was from a remote tester image

That's quite astonishing, if the same exact undocumented errata exists on both Intel and AMD chips!

mcb30:
> In terms of a potential fix, could you please try:
> 
> ```diff
> --- a/src/arch/x86/transitions/librm.S
> +++ b/src/arch/x86/transitions/librm.S
> @@ -1106,6 +1106,7 @@ vc_rmode:
>  	 * Fortunately, prot_to_real() zeroes the high word of %esp, so
>  	 * we can just use -20(%esp) instead.
>  	 */
> +	movzwl	%sp, %esp
>  	addr32 movl -20(%esp), %esp
>  	popfl
>  	popw	%ss /* padding */
> ```

I've updated this PR to include this potential fix (with lots of explanatory comments).  Could you try it out and see if it works?

jcapik:
> > In terms of a potential fix, could you please try:
> > ```diff
> > --- a/src/arch/x86/transitions/librm.S
> > +++ b/src/arch/x86/transitions/librm.S
> > @@ -1106,6 +1106,7 @@ vc_rmode:
> >  	 * Fortunately, prot_to_real() zeroes the high word of %esp, so
> >  	 * we can just use -20(%esp) instead.
> >  	 */
> > +	movzwl	%sp, %esp
> >  	addr32 movl -20(%esp), %esp
> >  	popfl
> >  	popw	%ss /* padding */
> > ```
> 
> I've updated this PR to include this potential fix (with lots of explanatory comments). Could you try it out and see if it works?

Yeah, this works!

mcb30:
> Yeah, this works!

Fantastic!  All merged: thank you very much for taking the time to get to the bottom of this.  Good luck with the next issue!

jcapik:
> That's quite astonishing, if the same exact undocumented errata exists on both Intel and AMD chips!

Well... even though the 386 market was defined by a shift away from licensed copies, the competition was full of reverse engineering and plagiarism, which led to legal battles, so they just went along with the bugs, probably because they thought it was the right thing to do and perhaps they were afraid of the problems that would arise from incompatibility. Hard to guess.

> Fantastic! All merged: thank you very much for taking the time to get to the bottom of this. Good luck with the next issue!

My pleasure! It was fun to dust off the long unused assembly language.

PR #1602, opened 2026-01-21
[ci] Use arch matrix in GitHub actions builds to reduce duplication

NiKiZe:
Use matrix to reduce some of the duplications between the steps.
More of the logic could be broken out to properties, but keep the changes minimal in this original refactoring.

Add snponly.efi for arm32 and arm64 is only intended logic change

Originally implemented by @TomTucka in #1184

Example of action running: https://github.com/NiKiZe/ipxe/actions/runs/21208638499

Possible addition with checksums and artifact uploads: https://github.com/NiKiZe/ipxe/actions/runs/21249367210

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,10 +26,25 @@ jobs:
                            libc6-dev-i386 liblzma-dev valgrind \
                            gcc-arm-none-eabi gcc-aarch64-linux-gnu
 
-  x86:
-    name: x86
+  buildmatrix:
+    name: Build ${{ matrix.arch }}
     runs-on: ubuntu-24.04
     needs: cache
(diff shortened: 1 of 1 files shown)


mcb30:
I'm in the middle of reworking the GitHub Actions completely, using the dedicated builder containers that are now created by https://github.com/ipxe/ipxe-builder.  Closing this PR since it will not form the basis for what gets merged.

mcb30:
> I'm in the middle of reworking the GitHub Actions completely, using the dedicated builder containers that are now created by https://github.com/ipxe/ipxe-builder.

For reference: this has been implemented as https://github.com/ipxe/ipxe/commit/d0ea2b1bb

PR #1610, opened 2026-01-29
Add support for distinct title color

3isenHeiM:
Adding the ability to modify the title color of the menu by adding a new `cpair` index 8.


mcb30:
I'm not opposed to this in principle.  But it would need several changes:

* Title colour should be applied to form titles as well as menu titles
* Default title colour should match the current colour used for titles, which is definitely not red/black
* Commit needs a proper author name, email address, and `Signed-off-by` tag

There is also a question that needs answering: how do you avoid breaking the visual appearance from an existing script where a user has defined a colour scheme that predates the addition of this feature?

PR #1617, opened 2026-02-03
[golan] update new Devices for Golan, and prevent setting SDR link speed for new devices.

ahmadm22a:
[Golan] update new Devices for Golan, and prevent setting SDR link speed for new devices.
[PCI] Prevent infinite loop in pcibus_probe.

### DIFF
--- a/src/drivers/bus/pci.c
+++ b/src/drivers/bus/pci.c
@@ -352,13 +352,20 @@ int pci_find_next ( struct pci_device *pci, uint32_t *busdevfn ) {
 		hdrtype &= PCI_HEADER_TYPE_MASK;
 		if ( hdrtype == PCI_HEADER_TYPE_BRIDGE ) {
 			pci_read_config_byte ( pci, PCI_SUBORDINATE, &sub );
-			end = PCI_BUSDEVFN ( PCI_SEG ( *busdevfn ),
-					     ( sub + 1 ), 0, 0 );
-			count = ( end - range.start );
-			if ( count > range.count ) {
-				DBGC ( pci, PCI_FMT " found subordinate bus "
-				       "%#02x\n", PCI_ARGS ( pci ), sub );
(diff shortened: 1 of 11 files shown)


src/drivers/bus/pci.c
```diff
@@ -463,7 +463,7 @@ static int pcibus_probe ( struct root_device *rootdev ) {
 
 		/* Skip automatic probing if prohibited */
 		if ( ! pci_can_probe ( pci ) )
-			continue;
+			break;
```
mcb30: This is completely wrong, and would cause the bus enumeration to simply terminate when the first non-permitted device is encountered.  The call to `pci_find_next()` ensures iteration to the next PCI device: there is no infinite loop here.

src/drivers/infiniband/golan.c
```diff
@@ -2239,8 +2239,12 @@ static inline int golan_bring_up(struct golan *golan)
 	if (( rc = golan_handle_pages(golan, GOLAN_INIT_PAGES, GOLAN_PAGES_GIVE) ))
 		goto pages;
 
-	if (( rc = golan_set_link_speed ( golan ) ))
-		goto pages_teardown;
+	if ( FLAG_TEST ( golan->pci->id->driver_data, SDR_SUPPORTED ) ) {
+		if (( rc = golan_set_link_speed ( golan ) ))
+			goto pages_teardown;
+	} else {
+		DBGC(golan, "%s: SDR not supported, skipping link speed set\n", __FUNCTION__);
```
mcb30: Is this logic actually correct?  Does link speed need to be set only for SDR?  Please justify this.

src/drivers/bus/pci.c
mcb30: OK.  You'll need to debug that further, sorry, because your suggested fix is definitely wrong.

Issue #1639, opened 2026-02-27
TLS stack ignores close_notify, causing HTTPS transfer stall


tuminoid:
# TLS stack ignores close_notify, causing HTTPS transfer stall

I was doing some comparison runs versus the iPXE TLS stack and my assumed happy case failed. It might be a test script fluke, so I'm including reproducer script so it can be blamed as well.

## Issue

[tls_new_alert()](https://github.com/ipxe/ipxe/blob/master/src/net/tls.c#L2043-L2074) in `src/net/tls.c` treats all [warning-level alerts](https://github.com/ipxe/ipxe/blob/master/src/net/tls.c#L2062-L2069) as informational no-ops, including close_notify (alert type 0). Per [RFC 5246 §7.2.1](https://www.rfc-editor.org/rfc/rfc5246#section-7.2.1), close_notify indicates the sender will not send any more data on the connection.

When a server sends close_notify after the HTTP response body (e.g. `openssl s_server -WWW`), iPXE ignores it and the transfer hangs until timeout rather than completing.

## Expected behavior

Transfer completes after server sends close_notify.

## Actual behavior

Transfer stalls until the 45-second fail-safe timeout in test script. Debug log shows the handshake and validation succeed, then:

```text
TLS 0x443948 received warning alert 0
```

## Reproducer

Save as `src/repro_close_notify.sh` and run from the repo root:

```bash
#!/bin/bash

set -euo pipefail

SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
WORK="/tmp/ipxe-close-notify-$$"
PORT=14433
GW="10.0.2.2"
SERVER_PID=""

cleanup() { kill "${SERVER_PID}" 2>/dev/null || true; }
trap cleanup EXIT
mkdir -p "${WORK}"

# Generate self-signed cert and test file
openssl req -x509 -newkey rsa:2048 -keyout "${WORK}/key.pem" -out "${WORK}/cert.pem" \
    -days 1 -nodes -subj "/CN=${GW}" -addext "subjectAltName=IP:${GW}" 2>/dev/null

echo "HELLO" > "${WORK}/hello.txt"
cat > "${WORK}/boot.ipxe" <<EOF
#!ipxe
dhcp net0
imgfetch https://${GW}:${PORT}/hello.txt && echo PASS || echo FAIL
exit
EOF

# Build iPXE
rm -rf "${SCRIPT_DIR}/bin-x86_64-linux"
[...]

PR #1640, opened 2026-02-27
Handle TLS closure alerts

mcb30:
Fixes: #1639


tuminoid:
Thanks for extremely quick fix!

PR #1644, opened 2026-03-06
CI: make build reproducible

shankerwangmiao:
This commit tries to make the built binaries reproducible by determining
the SOURCE_DATE_EPOCH according to the timestamp of the git commit and
mapping the path of the source files in the debug section to relative
ones.

xorriso is needed to be installed in the building environment to generate the iso images in a reproducible way.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,9 +10,27 @@ env:
 
 jobs:
 
+  sourcedate:
+    name: Source Date
+    runs-on: ubuntu-latest
+    outputs:
+      source_date_epoch: ${{ steps.sourcedate.outputs.source_date_epoch }}
+    steps:
+
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Source date
+        id: sourcedate
+        run: |
+          SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)
+          echo "source_date_epoch=${SOURCE_DATE_EPOCH}" | tee -a $GITHUB_OUTPUT
+
   bios:
     name: BIOS / ${{ matrix.arch }}
     runs-on: ubuntu-latest
+    needs:
+      - sourcedate
     strategy:
       fail-fast: false
       matrix:
@@ -24,6 +42,7 @@ jobs:
     env:
       bindir: >-
         ${{ matrix.arch == 'i386' && 'bin' || 'bin-x86_64-pcbios' }}
+      SOURCE_DATE_EPOCH: ${{ needs.sourcedate.outputs.source_date_epoch }}
     steps:
 
       - name: Check out code
@@ -69,6 +88,8 @@ jobs:
   sbi:
     name: SBI / ${{ matrix.arch }}
     runs-on: ubuntu-latest
+    needs:
+      - sourcedate
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -42,7 +61,8 @@ jobs:
                ${{ env.bindir }}/ipxe-legacy.lkrn \
                ${{ env.bindir }}/ipxe-legacy.pxe \
                ${{ env.bindir }}/undionly.kpxe \
-               ${{ env.bindir }}/errors
+               ${{ env.bindir }}/errors \
+               EXTRA_CFLAGS='-ffile-prefix-map=$(abspath $(CURDIR)/..)=.'
```
mcb30: Let's not use `EXTRA_CFLAGS` for this.  If it's worth doing at all, let's have it as a standalone commit that changes the Makefile to always specify this option (after checking that it is supported by the compiler).

shankerwangmiao:
> Before modifying anything in this PR, please get a PR merged to https://github.com/ipxe/ipxe-builder/ that replaces genisoimage with xorrisofs, so that this PR can actually pass CI checks.

I've opened ipxe/ipxe-builder#1 and temporarily point the docker image used to my fork. I have converted this PR a draft until ipxe/ipxe-builder#1 gets merged.

.github/workflows/build.yml
```diff
@@ -30,19 +30,70 @@ jobs:
...
+      ref-ipxe-builder-arm64: ${{ steps.imagehash.outputs.ref-ipxe-builder-arm64 }}
+      ref-ipxe-builder-loong64: ${{ steps.imagehash.outputs.ref-ipxe-builder-loong64 }}
+      ref-ipxe-builder-riscv32: ${{ steps.imagehash.outputs.ref-ipxe-builder-riscv32 }}
+      ref-ipxe-builder-riscv64: ${{ steps.imagehash.outputs.ref-ipxe-builder-riscv64 }}
+      ref-ipxe-builder-utils: ${{ steps.imagehash.outputs.ref-ipxe-builder-utils }}
+      ref-ipxe-signer: ${{ steps.imagehash.outputs.ref-ipxe-signer }}
+      ref-ipxe-signer-pcscd: ${{ steps.imagehash.outputs.ref-ipxe-signer-pcscd }}
+    steps:
+      - name: Setup Skopeo
+        uses: warjiang/setup-skopeo@main
```
mcb30: A hard no to this, sorry.  I'm not prepared to drag in random third-party actions.

.github/workflows/build.yml
```diff
@@ -79,6 +80,44 @@ jobs:
...
+                  hash: $image_hashes["ref-" + .image] | scan ("@([^@]*)$") | .[0],
+                }
+              ]),
+            }' | yq -p json | tee reproducible-env.yaml
+      - name: Upload
+        uses: actions/upload-artifact@v7
+        with:
+          if-no-files-found: error
+          path: reproducible-env.yaml
+          archive: false
```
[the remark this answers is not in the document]
mcb30: Understood.  I will not ever go down the route of "here is the SDK that you have to use to build this software", because I think that is fundamentally a very, very broken design.  A reproducible build is **necessarily** only reproducible **if you happen to be using the same toolchain**.  If you happen to be using a different toolchain then you should not expect the build to be reproducible.

This is a very fundamental limitation of the whole idea of reproducible builds.  The guarantee that it provides cannot ever sensibly be "you will be able to reproduce this binary".  The guarantee can only ever be "if you want to verify that this build pipeline is really doing what it says it is doing, then try rebuilding it yourself within a few minutes to a few hours of the original build, and see that you get the same binary".

PR #1652, opened 2026-03-14
added GCE support for TPU v7x gVNIC

omartin2010:
Per title - adding support for TPU V7x. Feel free to reach out if you have any questions.

### DIFF
--- a/src/drivers/net/gve.c
+++ b/src/drivers/net/gve.c
@@ -561,7 +561,7 @@ static int gve_describe ( struct gve_nic *gve ) {
 	/* Select preferred operating mode */
 	if ( gve->options & ( 1 << GVE_OPT_GQI_QPL ) ) {
 		/* GQI-QPL: in-order queues, queue page list addressing */
-		gve->mode = GVE_MODE_QPL;
+		gve->mode = ( GVE_MODE_DQO | GVE_MODE_QPL );
 	} else if ( gve->options & ( 1 << GVE_OPT_GQI_RDA ) ) {
 		/* GQI-RDA: in-order queues, raw DMA addressing */
 		gve->mode = 0;
@@ -573,7 +573,7 @@ static int gve_describe ( struct gve_nic *gve ) {
 		gve->mode = GVE_MODE_DQO;
 	} else {
 		/* No options matched: assume the original GQI-QPL mode */
-		gve->mode = GVE_MODE_QPL;
+		gve->mode = ( GVE_MODE_DQO | GVE_MODE_QPL );
 	}
 	DBGC ( gve, "GVE %p using %s mode\n",
 	       gve, gve_mode_name ( gve->mode ) );
--- a/src/drivers/net/gve.c
+++ b/src/drivers/net/gve.c
@@ -573,7 +573,7 @@ static int gve_describe ( struct gve_nic *gve ) {
 		gve->mode = 0;
 	} else {
 		/* No options matched: assume the original GQI-QPL mode */
-		gve->mode = ( GVE_MODE_DQO | GVE_MODE_QPL );
+		gve->mode = GVE_MODE_QPL;
 	}
 	DBGC ( gve, "GVE %p using %s mode\n",
 	       gve, gve_mode_name ( gve->mode ) );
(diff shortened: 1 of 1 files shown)


src/drivers/net/gve.c
```diff
@@ -561,7 +561,7 @@ static int gve_describe ( struct gve_nic *gve ) {
 	/* Select preferred operating mode */
 	if ( gve->options & ( 1 << GVE_OPT_GQI_QPL ) ) {
 		/* GQI-QPL: in-order queues, queue page list addressing */
-		gve->mode = GVE_MODE_QPL;
+		gve->mode = ( GVE_MODE_DQO | GVE_MODE_QPL );
```
mcb30: No.  The code branch here is for "does the driver advertise support for GQI-QPL?".  Selecting DQO rather than GQI on this branch is very obviously the wrong thing to do.

mcb30 (approved):
This PR breaks support for many other forms of gVNIC: see inline comments for detailed rejection reasons.

The code is already capable of operating devices that advertise support for the DQO-QPL mode.  You should not require any changes for the TPU v7x, unless that device is itself faulty and is advertising support for a mode that it cannot actually operate in.

Does the TPU v7x have a bug that causes it to advertise support for unsupported modes of operation?

mcb30 (changes requested):
Sorry, accidentally marked review as approved.  Should be rejected.

omartin2010:
Ok, thanks for the feedback - I will propose an updated change to this one, and if it doesn't make sense, will organize to connect offline. Thanks again!

mcb30 (changes requested):
NACK on this, sorry.  iPXE prefers to use GQI rather than DQO: see commit https://github.com/ipxe/ipxe/commit/ce30ba14f for reasons.

The existing code already supports the operating mode that you are trying to use.  If the hardware does not advertise support for GQI then iPXE will not use it.  Again: does the TPU v7x have a hardware bug that is causing it to advertise unsupported modes as supported?

omartin2010:
Per below (boot with `DEBUG=gve` in the build => 0x0f58) seems to indicate that both GQI_QPL and DQO_RDA are supported. However when we attempt booting with GQI_QPL as the selected mode, it's failing - and you are right, the device probably should not advertise this mode. I see what you have in the commit you referred to, above. Let me check internally.
```
GVE 0xbca8e3b0 is revision 0x00
GVE 0xbca8e3b0 AQ at [bcfaf000,bcfb0000) scratch [bcfae000,bcfae100)
GVE 0xbca8e3b0 skipping reset
GVE 0xbca8e3b0 using 1024 TX, 1024 RX, 32 events
GVE 0xbca8e3b0 MAC 42:01:0a:00:00:08 ("10.0.0.8") MTU 1894
GVE 0xbca8e3b0 supports options 0x00000f58
GVE 0xbca8e3b0 using DQO-RDA mode
GVE 0xbca8f3f0 is revision 0x00
GVE 0xbca8f3f0 AQ at [bc875000,bc876000) scratch [bc874000,bc874100)
GVE 0xbca8f3f0 skipping reset
GVE 0xbca8f3f0 using 1024 TX, 1024 RX, 32 events
GVE 0xbca8f3f0 MAC 42:01:0a:01:00:08 ("10.1.0.8") MTU 1894
GVE 0xbca8f3f0 supports options 0x00000f58
GVE 0xbca8f3f0 using DQO-RDA mode
```

omartin2010:
@mcb30 pls disregard the added commit for now.

mcb30:
> Per below (boot with `DEBUG=gve` in the build => 0x0f58) seems to indicate that both GQI_QPL and DQO_RDA are supported. However when we attempt booting with GQI_QPL as the selected mode, it's failing - and you are right, the device probably should not advertise this mode. I see what you have in the commit you referred to, above. Let me check internally.

Thank you.  One potentially valuable test would be to try patching the Linux driver's `gve_adminq_describe_device()` to prefer `GVE_GQI_QPL_FORMAT` and see if it is capable of driving the device in that mode.  If it is, then there is probably a bug in iPXE.  If not, then the bug is in the hardware's advertising the wrong list of operating modes.

mcb30:
Following up on this: Google has confirmed off-list that the TPU v7x does not support GQI and should therefore not be advertising it as a supported mode.  This change should by now have been rolled out within GCE, and so the TPU v7x instances should work as expected with an unmodified iPXE.

PR #1661, opened 2026-03-31
Security: Fix off-by-one heap overflow in nfs_uri_symlink()

warsang:
## Summary

The `nfs_uri_symlink()` function has an off-by-one heap buffer overflow. The length calculation omits +1 for the null terminator. When `malloc(len)` is called, it allocates exactly `len` bytes, but `strcpy()` writes the final NUL at offset `len`, one byte past the heap allocation.

## Issue Validation

Validated via taint analysis:

**Root Cause (lines 96-97 and 108):**

Absolute path case:
```c
len = strlen(uri->lookup_pos) + strlen(symlink) - strlen(uri->mountpoint);
new_path = malloc(len);  // Allocates len bytes, not len + 1
```

Relative path case:
```c
len = strlen(uri->lookup_pos) + strlen(symlink);
new_path = malloc(len);  // Allocates len bytes, not len + 1
```

**Vulnerable sink (lines 104-105 and 113-114):**
```c
strcpy(new_path, symlink + strlen(uri->mountpoint));
strcpy(new_path + strlen(new_path), uri->lookup_pos);
// Final NUL written at new_path[len] — one byte past allocated buffer
```

The calculation correctly computes the string content length but forgets the null terminator.

## Impact

A malicious NFS server can trigger heap corruption:
- iPXE client crash during boot (DoS)
- Heap memory corruption in iPXE's constrained environment

Note: Exploitation beyond DoS is impractical in iPXE's position-independent, minimal heap environment.

## Reproduction

For relative path case:
1. Create NFS symlink with target "" (empty string, 0 chars), where mountpoint is already set
2. When `nfs_uri_symlink()` processes it: `len = strlen("") + strlen(lookup_pos) = 0 + N`
3. `malloc(0+N)` allocates N bytes, but concatenated string needs N+1 bytes
4. Final `strcpy()` writes NUL at offset N — one byte past heap

## Fix

1. **Add +1 to both length calculations:**
   ```c
   len = strlen(uri->lookup_pos) + strlen(symlink) + 1;
   ```
2. Replace `strcpy()` chaining with `memcpy()` using explicit lengths
3. Explicit null termination

## Files Changed

- `src/net/oncrpc/nfs_uri.c`

### DIFF
--- a/src/net/oncrpc/nfs_uri.c
+++ b/src/net/oncrpc/nfs_uri.c
@@ -97,7 +97,7 @@ int nfs_uri_symlink ( struct nfs_uri *uri, const char *symlink ) {
 			return -EINVAL;
 
 		len = strlen ( uri->lookup_pos ) + strlen ( symlink ) - \
-		      strlen ( uri->mountpoint );
+		      strlen ( uri->mountpoint ) + 1;
 		if ( ! ( new_path = malloc ( len * sizeof ( char ) ) ) )
 			return -ENOMEM;
 
@@ -105,7 +105,7 @@ int nfs_uri_symlink ( struct nfs_uri *uri, const char *symlink ) {
 		strcpy ( new_path + strlen ( new_path ), uri->lookup_pos );
 
 	} else {
-		len = strlen ( uri->lookup_pos ) + strlen ( symlink );
+		len = strlen ( uri->lookup_pos ) + strlen ( symlink ) + 1;
 		if ( ! ( new_path = malloc ( len * sizeof ( char ) ) ) )
 			return -ENOMEM;
 


warsang:
Hi,

I wanted to follow up about this PR and on the overall disclosure process.

I should have reported this privately through a security advisory rather than opening a public PR. I'm new to the OSS disclosure process so this is a bit of a first. The fix and write-up are already visible, so the disclosure has effectively happened, but I'd like to do this properly going forward and make sure the fix gets formally tracked.

Would you be open to opening a GitHub Security Advisory for this so we can request a CVE and give it a proper paper trail? I'm happy to help draft it; most of the content from the PR description carries straight over.

Thanks for the understanding  and apologies for skipping the right channel on this one. I'll use the private advisory flow for any future findings.

mcb30:
No, please don't bother with any security advisory process. The NFS code is already known to not be production quality, which is why it is explicitly excluded from Secure Boot builds.

I expect that there will be many similar issues within the NFS parts of the code. Could you please locate any other places where the NFS  code fails to allocate space for a terminating NUL and update the PR to address them all in one go?

NiKiZe:
Could/should NFS be dropped? Anyone or any usecase where http can't be used instead?, or is there anything in NFS that provides any additional feature over http?
I know there have been asks about NFS when booting Linux, but that is from Linux, and not really required from iPXE, downside being that you have to run multiple services to serve the data,.

@Mareo still around and having any thoughts on the topic?

warsang:
> No, please don't bother with any security advisory process. The NFS code is already known to not be production quality, which is why it is explicitly excluded from Secure Boot builds.

Got it, no worries!

> I expect that there will be many similar issues within the NFS parts of the code. Could you please locate any other places where the NFS code fails to allocate space for a terminating NUL and update the PR to address them all in one go?

I took a quick look at this, the above bug I  initially found with some SAST tooling I'm working on.

For the above, request,

I manually looked again at every malloc, strcpy, strcat, strdup, and strndup call across all NFS-related files I could find and I ran the taint rule that found this in the first place again targeting the exact same bug pattern: malloc(N * sizeof(char)) without +1 for NUL then chained to strcpy(buf + strlen(buf), src)

I didn't find other occurrences with the taint rule.
Most of the NFS code seems to use strdup (which handles NUL internally), strndup with explicit length, or memcpy with explicit NUL termination.

So maybe I missed something but I think this PR addresses the full scope of the NUL-terminator bug  (specifically for the NFS code; I ran my tooling on the full ipxe project but I didn't manually review the whole codebase).

Mareo:
Hi!

The intent behind NFS support was to allow nfsroot users to avoid setting-up a HTTP server with the sole purpose of serving a kernel and initramfs. One could load iPXE via TFTP and then load its kernel and initramfs from the same NFS server used as a nfsroot target.

I do not have any way to know if this is a feature that is often used or not, after all the cost of spinning up an HTTP server is low nowadays.

I was a student and still fairly inexperienced when I developed NFS support, I am thus not surprised such a bug slipped by in spite of my trying to be careful with memory allocations.

NiKiZe:
Related https://github.com/ipxe/ipxe/pull/698#issuecomment-2031745651

mcb30:
> I didn't find other occurrences with the taint rule. Most of the NFS code seems to use strdup (which handles NUL internally), strndup with explicit length, or memcpy with explicit NUL termination.
> 
> So maybe I missed something but I think this PR addresses the full scope of the NUL-terminator bug (specifically for the NFS code; I ran my tooling on the full ipxe project but I didn't manually review the whole codebase).

Great, thank for you checking.  I will review the actual code in the PR now.

src/net/oncrpc/nfs_uri.c
```diff
@@ -96,22 +99,46 @@ int nfs_uri_symlink ( struct nfs_uri *uri, const char *symlink ) {
 			       strlen ( uri->mountpoint ) ) != 0 )
 			return -EINVAL;
 
-		len = strlen ( uri->lookup_pos ) + strlen ( symlink ) - \
-		      strlen ( uri->mountpoint );
-		if ( ! ( new_path = malloc ( len * sizeof ( char ) ) ) )
+		mountpoint_len = strlen ( uri->mountpoint );
+		symlink_len = strlen ( symlink );
+		lookup_pos_len = strlen ( uri->lookup_pos );
```
mcb30: This seems unnecessarily complicated.  Just add 1 to account for the NUL.  No need to introduce multiple extra variables.

src/net/oncrpc/nfs_uri.c
```diff
@@ -96,22 +99,46 @@ int nfs_uri_symlink ( struct nfs_uri *uri, const char *symlink ) {
...
+
+		/* Security: Calculate exact length needed and verify no overflow */
+		if ( symlink_len < mountpoint_len )
+			return -EINVAL;
+
+		len = lookup_pos_len + ( symlink_len - mountpoint_len );
+		if ( len >= symlink_len || len >= lookup_pos_len )
+			return -EINVAL;
```
mcb30: These overflow checks are meaningless and can be skipped.  iPXE's internal heap is tiny (~512kB) and there is no way that adding string lengths can ever overflow.  Please remove this noise.

src/net/oncrpc/nfs_uri.c
```diff
@@ -96,22 +99,46 @@ int nfs_uri_symlink ( struct nfs_uri *uri, const char *symlink ) {
...
+		if ( ! new_path )
 			return -ENOMEM;
 
-		strcpy ( new_path, symlink + strlen ( uri->mountpoint ) );
-		strcpy ( new_path + strlen ( new_path ), uri->lookup_pos );
+		/* Security: Use memcpy with explicit lengths instead of strcpy */
+		memcpy ( new_path, symlink + mountpoint_len, symlink_len - mountpoint_len );
+		new_path[symlink_len - mountpoint_len] = '\0';
+		memcpy ( new_path + ( symlink_len - mountpoint_len ), uri->lookup_pos, lookup_pos_len );
+		new_path[len] = '\0';
```
mcb30: No.  If you know that the length is correct, then just use `strcpy()`.  Open-coding `strcpy()` like this achieves nothing and just bloats the code.

src/net/oncrpc/nfs_uri.c
```diff
@@ -96,22 +99,46 @@ int nfs_uri_symlink ( struct nfs_uri *uri, const char *symlink ) {
...
-		len = strlen ( uri->lookup_pos ) + strlen ( symlink );
-		if ( ! ( new_path = malloc ( len * sizeof ( char ) ) ) )
-			return -ENOMEM;
+		symlink_len = strlen ( symlink );
+		lookup_pos_len = strlen ( uri->lookup_pos );
+
+		/* Security: Calculate exact length needed and verify no overflow */
+		len = lookup_pos_len + symlink_len;
+		if ( len < symlink_len || len < lookup_pos_len )
+			return -EINVAL;
```
mcb30: As above: checking for string length overflow is pointless since it is impossible within the memory model.

src/net/oncrpc/nfs_uri.c
```diff
@@ -96,22 +99,46 @@ int nfs_uri_symlink ( struct nfs_uri *uri, const char *symlink ) {
...
+		if ( ! new_path )
+			return -ENOMEM;
 
-		strcpy ( new_path, symlink );
-		strcpy ( new_path + strlen ( new_path ), uri->lookup_pos );
+		/* Security: Use memcpy with explicit lengths instead of strcpy */
+		memcpy ( new_path, symlink, symlink_len );
+		new_path[symlink_len] = '\0';
+		memcpy ( new_path + symlink_len, uri->lookup_pos, lookup_pos_len );
+		new_path[len] = '\0';
```
mcb30: As above: just use `strcpy()` if you already know that the lengths are correct.

mcb30 (changes requested):
General feedback: adding any code comment block starting with `/* Security: do thing ... */` is always wrong, unless there is a very obscure security point being addressed.  Ensuring correct NUL termination of strings is just everyday coding: the comments should not draw attention to this as though it is something special.

Other feedback is provided inline.  You should be able to get the patch down to just a couple of lines changed.

Thanks!

warsang (commented):
@mcb30 Applied requested changes; Much simpler approach. Thanks for the review, let me know if I need to change anything else.

mcb30:
Looks good now, please just edit the commit to use a real name and add a Signed-off-by declaration.

warsang:
@mcb30 Updated, let me know if you need me to make any other changes

mcb30 (approved):

mcb30:
Looks good, thank you.  Rebased, and merging now.

PR #1682, opened 2026-04-12
[serial] add dumb console support

igaw:
LAVA cannot handle escape sequences properly. Add a 'dumb' serial console option that filters out escape sequences to improve compatibility.

The LAVA project provides a 'hacked' version of iPXE (https://docs.lavasoftware.org/lava/ipxe.html) but I haven't really seen any upstreaming attempts of this version. Thus this is my attempt to get a 'dumb' serial console feature upstream. 

My first version use a ifdef inside serial.c but my understanding is that is not how the project handles this, but just in case you are wondering how it would look like: https://github.com/igaw/ipxe/commit/5882dbdf7c4c017ecaa8f4e0b7b59f28c823bbce

Anyway, with this feature I am able to use the current upstream version with LAVA.

### DIFF
--- a/src/core/serial_dumb.c
+++ b/src/core/serial_dumb.c
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2026 Daniel Wagner <wagi@monom.org>.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ *
+ * You can also choose to distribute this program under the terms of
+ * the Unmodified Binary Distribution Licence (as given in the file
+ * COPYING.UBDL), provided that you have satisfied its requirements.
+ */
+
+FILE_LICENCE ( GPL2_OR_LATER_OR_UBDL );
+
+/** @file
+ *
+ * Dumb serial console
+ *
+ * Strips ANSI escape sequences from serial console output.  Useful
+ * for environments (e.g. LAVA) that cannot handle escape sequences.
+ *
+ */
+
+#include <ipxe/uart.h>
+#include <ipxe/ansiesc.h>
+#include <ipxe/console.h>
+#include <ipxe/serial.h>
+#include <ipxe/init.h>
+
+/** Serial console driver */
+extern struct console_driver serial_console_driver;
+
+/** Dumb serial console ANSI escape sequence handlers */
+static struct ansiesc_handler serial_dumb_ansiesc_handlers[] = {
+	{ 0, NULL }
+};
+
+/** Dumb serial console ANSI escape sequence context */
+static struct ansiesc_context serial_dumb_ansiesc_ctx = {
+	.handlers = serial_dumb_ansiesc_handlers,
+};
+
+/**
+ * Print a character to dumb serial console
+ *
+ * @v character		Character to be printed
+ */
+static void serial_dumb_putchar ( int character ) {
+
+	/* Do nothing if we have no UART */
+	if ( ! serial_console )
+		return;
+
+	/* Intercept ANSI escape sequences */
+	character = ansiesc_process ( &serial_dumb_ansiesc_ctx, character );
+	if ( character < 0 )
+		return;
+
+	/* Transmit character */
+	uart_transmit ( serial_console, character );
+}
+
+/**
+ * Initialise dumb serial console
+ *
+ * Replaces the default serial console putchar with a version that
+ * strips ANSI escape sequences before transmitting.  This runs at
+ * INIT_EARLY priority, before any console output takes place.
+ */
+static void serial_dumb_init ( void ) {
+	serial_console_driver.putchar = serial_dumb_putchar;
+}
+
+/** Dumb serial console initialisation function */
+struct init_fn serial_dumb_init_fn __init_fn ( INIT_EARLY ) = {
+	.name = "serial_dumb",
+	.initialise = serial_dumb_init,
+};
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -24,6 +24,7 @@ FILE_SECBOOT ( PERMITTED );
 
 #include <config/general.h>
 #include <config/console.h>
+#include <config/serial.h>
 #include <config/sideband.h>
 #include <config/settings.h>
 
@@ -54,6 +55,9 @@ PROVIDE_REQUIRING_SYMBOL();
 
 #ifdef CONSOLE_SERIAL
 REQUIRE_OBJECT ( serial );
+#ifdef SERIAL_DUMB
+REQUIRE_OBJECT ( serial_dumb );
+#endif
 #endif
 #ifdef CONSOLE_DIRECT_VGA
 REQUIRE_OBJECT ( video_subr );
--- a/src/config/serial.h
+++ b/src/config/serial.h
@@ -30,6 +30,11 @@ FILE_LICENCE ( GPL2_OR_LATER );
 //#undef SERIAL_SPCR
 //#define SERIAL_FIXED
 
+/* Uncomment to suppress ANSI escape sequences on the serial console.
+ * Useful for environments (e.g. LAVA) that cannot handle escape sequences.
+ */
+//#define SERIAL_DUMB
+
 /* Early UART configuration (for bare metal prefix debugging only) */
 //#define EARLY_UART_MODEL	8250
 //#define EARLY_UART_REG_BASE	0x10000000


igaw:
Adding @broonie as he might be interested in this feature as well.

NiKiZe:
So, have to ask, why does LAVA not add support for handling ansii escape codes, could be useful for other things as well?

igaw:
I don't know. Lava is written in Python and as far I can tell from my own experience, getting encoding/decoding right seems to be dark magic.

@ivoire as you are very active on the Lava repo, do you happen to know the reason why Lava does not handle ANSII escape codes?

src/core/serial_dumb.c
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Initialise dumb serial console
+ *
+ * Replaces the default serial console putchar with a version that
+ * strips ANSI escape sequences before transmitting.  This runs at
+ * INIT_EARLY priority, before any console output takes place.
+ */
+static void serial_dumb_init ( void ) {
+	serial_console_driver.putchar = serial_dumb_putchar;
```
mcb30, line 82: I appreciate the effort to avoid using a plain `#ifdef`, but this kind of monkey patching would also not be acceptable, sorry.

mcb30 (changes requested):
First, there needs to be a very good justification for why LAVA can't be fixed to handle (or just strip out) ANSI escape sequences.

ANSI escape sequences (at least as used by iPXE) always match the format defined at https://en.wikipedia.org/wiki/ANSI_escape_code#CSIsection:
> For Control Sequence Introducer, or CSI, commands, the `ESC [` (written as `\e[`, `\x1b[` or `\033[` in several programming languages) is followed by any number (including none) of "parameter bytes" in the range 0x30–0x3F (ASCII `0–9:;<=>?)`, then by any number of "intermediate bytes" in the range 0x20–0x2F (ASCII space and `!"#$%&'()*+,-./`), then finally by a single "final byte" in the range 0x40–0x7E (ASCII ```@A–Z[\]^_`a–z{|}~```).

which is simple enough to strip with a one-line regex along the lines of:
```python
stripped_output = re.sub(r'\x1b\[.*?[\x40-\x7e]', '', original_output)
```

src/core/serial_dumb.c
igaw: Thanks for the feedback. I agree it is still a hack.

I just wanted to point out that there is also en effort to `NO_COLOR` env switch supported by command line tools: https://no-color.org/ Maybe I could convince you, that not everyone wants color output?

Anyway, I'll see if I can get the regex filter into Lava's line parser then. If not, it's not the end of the world, maintaining this patch isn't the that big of a deal.

NiKiZe: I kind of get a feeling we are missing context here, are there examples of which sequences would appear in a normal flow, maybe a raw serial capture could be done? And also then why the sequences are creating issues?
Maybe you should just escape the escape char instead?

iPXE does a few highlights on startup for readability, on screen, these are minimal. The only thing after this using colors by default is menues, which would be weird if you tried to use i LAVA(?)
Other than that it is user scripts, which iPXE does not control, it just sends bytes raw.

There is control chars to repaint % progress, which is not colour.

So I assume that the only real thing that a LAVA setup would encounter is the quite static iPXE header, if that somehow is parsed in LAVA, but then it might as well be parsed with the ANSI sequences. After that there should be no colors from iPXE itself in a LAVA context, unless the user has themselves set colors in iPXE scripts.

So there is not much colors to deal with here, and even so it does make sense in a terminal application to want to reduce this where piping is common, it does make less sense to avoid this where the main output is raw on screen output.

igaw: I can grab some logs later. In the meantime I can explain where LAVA sometimes falls on its nose:

image

Basically the second line is the offending one ('iPXE...' and 'http://...' is using colours) when it is waiting for the 'Press Ctrl-B for the iPXE command line...'.  Then it sends the 'Ctrl-b' and types in the commands which is no issue at all.

If there would be an option to disable this part, all would be fine for the use case I am interested in.

NiKiZe: Yes, that is essentially the only places with colors, But what does the ANSII escape codes actually cause for issues in python then?, it should just ignore this anyway. Maybe if this ends up in a log somewhere it might be weird, but iPXE can not be the only thing that outputs escape sequences, where colors is the least of your problems. It just sounds like there is a bug elsewhere that LAVA is trying to work around. LAVA should be just fine dealing with this as either raw bytes, or even ASCII.

broonie: In my usage it tends to work fine, just result in illegible logs and can create issues matching on the text that's got escape sequences in it since it's hard to type those into configuration.

igaw: > Yes, that is essentially the only places with colors, But what does the ANSII escape codes actually cause for issues in python then?

I don't know for sure. I had troubles with one of machine because LAVA didn't match on the expected string. I was not really able to figure out why this is the case but after removing the colours did the trick for me.

Okay, what about making 

```
	printf ( NORMAL "\n\n" PRODUCT_NAME "\n" BOLD PRODUCT_SHORT_NAME " %s"
		 NORMAL " -- " PRODUCT_TAG_LINE " -- "
		 CYAN PRODUCT_URI NORMAL "\nFeatures:", product_version );
```

configurable then? Just joking...

As I said I can carry this patch locally if none else needs it.

NiKiZe:
Can someone explain how the lava flow works here? I would assume python code runs menu with iPXE and grabs serial thru console output?

igaw:
There is a bit of abstractions in LAVA, so not that easy to follow what's happening. So this is just my understanding.  

Every command in LAVA waits for the prompt to appear and when it sees it does the next action. It reads from the serial console/uart and uses pexpect for finding the prompt (or in this case it looks for the Ctrl-B message:

image


image

Note the logging output happens after the match, thus it looks like in the wrong order...

NiKiZe:
It feels odd that you drive iPXE from the shell, rather than having an iPXE script. (it would make sense if iPXE was what was tested, but it AFAIK is not)
But my main question is how does the iPXE output reach the python code, the log suggests a connection over Legacy IPv4 localhost.

Is QemuSession involved - but no connection there?
Since in this case iPXE isn't running directly, I would assume qemu.
So iPXE outputs to a serial port on qemu, which then shows this somewhere, ssh/telnet, which then python code connects to, parses and sends to a log.
Which means there is a few places to verify which output you get, is it the program, potentially telnet, that mashes up the serial port output? Or is there somewhere in there where the encoding goes wrong, Someone that can run LAVA needs to do some debugging or getting some RAW bytes out of the pipe to see where it potentially goes wrong.

Changing what iPXE outputs is a hack that is clear is a bug in the implementation used in LAVA.
Again colors is just one part - a line that you essentially ignore anyways. except for logs, you want those sequences in there.
After Ctrl-B is displayed, it actually moves the cursor to erase it, so that is a different part where this happens.

igaw:
The script will not work for LAVA. LAVA creates a new job with a new unique id for every run, thus a static script doesn't do the trick. 

LAVA allows different ways how to communicate with a DUT. In my setup I use ser2net which allows a remote access to the port (thus the telnet log entry above). As I said, there are a lot of abstraction which allows to use it in various configurations. I don't know what LAVA uses for Qemu, but I would expect it is hooking up into the serial interface via the command line. 

I understand you don't want to support any plain serial console output. As Mark pointed out it might just work fine as it is and I am chasing a ghost. I do like that my logs are free of the escape sequences and I think it is worth having plain logs without special stuff in it. But that's just me obviously.

Anyway,I don't want to waste your time any longer. Thanks for your feedback.

NiKiZe:
You can for sure use dynamic scripts as well, they don't have to be static, there is several ways this can be done.
I can understand that this path was chosen, because it is done in most/all other cases. I must however still question if it is the best or right way to do it.

So ser2net, how is that configured? or maybe I was mistaken here and DUT is a physical machine with a serial port? - that actually makes my point with the dynamic script mentioned above even easier.
I'm trying to understand where the possible issues is, if the escape sequences show up correctly in the logs, then you might just want to apply the filter to the logs instead.
but if it is not properly formatted there, maybe there is something else going on.

I find LAVA to be an interesting concept now that I know about it, so I personally tried to take the opportunity to learn, so not in any way wasted time if we can come to a solution that maybe even if an improvement for all.

igaw:
When a new job is submitted to LAVA, it contains different actions, e.g.

```yaml
actions:
- deploy:
    kernel:
      type: zimage
      url: http://lava.lan:8080/artifacts/c2d/c2d-image-rt-v6.12.79-rt17
    nfsrootfs:
      url: http://lava.lan:8080/artifacts/rootfs/lava-image.x86_64-1.tar.xz
    os: debian
    timeout:
      minutes: 5
    to: tftp
- boot:
    auto_login:
      login_commands:
      - 'export PS1="lava: "'
      - export TERM=dumb
      - export NO_COLOR=1
      login_prompt: 'login:'
      username: root
    commands: nfs
    method: ipxe
    prompts:
    - 'lava:'
    timeout:
      minutes: 5
```

The deploy action instructs LAVA to setup a root filesystem accessible via nfs and the kernel via tftp. The boot actions tells LAVA which method is used, here ipxe (or other types such as u-boot etc). The paths to the nfs root and kernel have the job id in the path:

```
- {DTB}: None
- {INITRD}: None
- {KERNEL}: 15/tftp-deploy-ddkt56me/kernel/c2d-image-rt-v6.12.79-rt17
- {LAVA_MAC}: b8:ae:ed:74:f9:68
- {NFSROOTFS}: /var/lib/lava/dispatcher/tmp/15/extract-nfsrootfs-gybdcks9
- {NFS_SERVER_IP}: 192.168.60.5
- {PRESEED_CONFIG}: None
- {PRESEED_LOCAL}: None
- {RAMDISK}: None
- {ROOT_PART}: None
- {ROOT}: None
- {SERVER_IP}: 192.168.60.5
- {TEE}: None
```

As far I understand there is no support in LAVA which could create a boot script which iPXE could fetch. I assume this what you mean with dynamic scripts could work. 

I am using [lava-docker](https://github.com/BayLibre/lava-docker) and in my setup ser2net is:

```yaml
%YAML 1.1
---
connection: &con63001
  accepter: telnet(rfc2217),tcp,63001
  enable: on
  connector: serialdev,/dev/bbb,115200n81,local
  options:
    max-connections: 10
    trace-both-hexdump: true
    trace-both-timestamp: true
    trace-both: /tmp/bbb.trace
connection: &con63002
  accepter: telnet(rfc2217),tcp,63002
  enable: on
  connector: serialdev,/dev/c2d,115200n81,local
  options:
    max-connections: 10
    trace-both-hexdump: true
    trace-both-timestamp: true
    trace-both: /tmp/c2d.trace
connection: &con63003
  accepter: telnet(rfc2217),tcp,63003
  enable: on
  connector: serialdev,/dev/rpi3-32,115200n81,local
  options:
    max-connections: 10
    trace-both-hexdump: true
    trace-both-timestamp: true
    trace-both: /tmp/rpi3-32.trace
connection: &con63004
  accepter: telnet(rfc2217),tcp,63004
  enable: on
  connector: serialdev,/dev/rpi3-64,115200n81,local
  options:
    max-connections: 10
    trace-both-hexdump: true
    trace-both-timestamp: true
    trace-both: /tmp/rpi3-64.trace
```

The LAVA slave container has direct access to the DUTs, but it is recommended to use ser2net.

ivoire:
> I don't know. Lava is written in Python and as far I can tell from my own experience, getting encoding/decoding right seems to be dark magic.
> 
> @ivoire as you are very active on the Lava repo, do you happen to know the reason why Lava does not handle ANSII escape codes?

This should just work fine in LAVA without any modification in iPXE. So it's either a bug in LAVA or a misconfiguration on your side. I'm off this week, so I will ping you next Monday to try to reproduce your issue.

ivoire:
@igaw could you open an issue on LAVA directly? At https://gitlab.com/lava/lava/-/work_items

igaw:
@ivoire Sure will do. I'll first collect some logs etc. This will take a bit of time as I am busy with other things.

PR #1694, opened 2026-04-28
[efi] Fix operator precedence in autoexec network download

refrigerator18:
## Bug

In `efi_autoexec_network()`, the `!=` operator has higher precedence than `=`, so:

```c
rc = imgacquire ( ..., image ) != 0
```

is parsed as:

```c
rc = ( imgacquire ( ..., image ) != 0 )
```

This assigns the boolean result (0 or 1) to `rc` instead of the actual return code from `imgacquire()`. When both downloads fail, `strerror(rc)` reports an incorrect error since `rc` is 1 rather than the real error code (e.g. `-ETIMEDOUT`).

Both assignments in the `if` condition on lines 135-138 have this problem.

## Fix

Add parentheses around each assignment to ensure `rc` captures the actual return value:

```c
if ( ( ( rc = imgacquire ( ... ) ) != 0 ) &&
     ( ( rc = imgacquire ( ... ) ) != 0 ) ) {
```

This matches the pattern already used correctly in `efi_autoexec_filesystem()` on lines 88-94 of the same file.

### DIFF
--- a/src/interface/efi/efi_autoexec.c
+++ b/src/interface/efi/efi_autoexec.c
@@ -132,10 +132,10 @@ static int efi_autoexec_network ( EFI_HANDLE handle, struct image **image ) {
 	}
 
 	/* Attempt download from current working URI, then from root */
-	if ( ( rc = imgacquire ( EFI_AUTOEXEC_NAME, EFI_AUTOEXEC_TIMEOUT,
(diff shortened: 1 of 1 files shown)


mcb30:
Good catch, thank you.  Modified to adjust line length and to remove the invalid Signed-off-by line.

PR #1699, opened 2026-04-30
[efi] Register EFI IPv6 device path settings as netX.ndp

### LINKED ISSUE #1695: IPv6 gateway missing fe80:: prefix when using NII driver on Intel E610 (Dell PowerEdge)
## Summary

When iPXE uses the NII driver (UEFI firmware-provided), the IPv6 default gateway learned from Router Advertisements is missing its `fe80::` link-local prefix. The router's link-local address `fe80::21c:73ff:fe00:1` is stored as `::21c:73ff:fe00:1`, making the route unusable and breaking all IPv6 connectivity.

## Environment

- **iPXE version:** 2.0.0+ (g8d2eb)
- **NIC:** Intel Ethernet Controller E610 (PCI IDs 8086:57ae–57b2)
- **Server:** Dell PowerEdge (UEFI HTTP Boot → iPXE)
- **Driver in use:** NII (no native iPXE driver for E610 (?))
- **DHCPv6 server:** Kea/Stork
- **Router Advertisements:** radvd (standard fe80:: link-local source)

## Symptoms

After `ifconf -c ipv6 net2`, the routing table shows:

     net2: <IPv6>/64 gw ::21c:73ff:fe00:1

The gateway should be `fe80::21c:73ff:fe00:1`. Pings and HTTP fetches to off-link destinations fail with "Connection timed out."

## Reproduction

1. Boot a Dell PowerEdge with an Intel E610 NIC via UEFI HTTP Boot into `ipxe.efi`
2. `ifstat` shows the NII driver is in use (no native E610 driver in iPXE)
3. `ifconf -c ipv6 net2` — acquires DHCPv6 address and processes RA
4. `route` — gateway is shown without the `fe80::` prefix
5. `ping <any-off-link-ipv6>` — fails

## Key observations

- **Same `ipxe.efi` binary booted from ISO works correctly** — gateway shows `fe80::21c:73ff:fe00:1`. The difference is that ISO boot uses iPXE's native driver, while UEFI HTTP Boot falls back to the NII driver since iPXE has no native E610 support.
- **Supermicro servers on the same subnet with the same RA/DHCPv6 config work fine** — this is specific to the NII driver path on Dell.
- **The RA source address is correct** — the router's actual link-local is `fe80::21c:73ff:fe00:1`, confirmed via `radvdump` and working clients.

## Workaround

Re-creating the route by setting `ip6` + `gateway6` together on the interface triggers the settings applicator and prepends a correct routing entry:
Re-creating the route by setting `ip6` + `gateway6` together on the interface triggers the settings applicator and prepends a correct routing entry:

     #!ipxe
     ifconf -c ipv6 net2 ||
     set net2/ip6 ${net2/ip6}
     set net2/len6 64
     set net2/gateway6 fe80${net2/gateway6}

This works because `${net2/gateway6}` returns `::21c:73ff:fe00:1`, and prepending `fe80` produces the correct `fe80::21c:73ff:fe00:1`.

## Possible root causes

1. **NII driver is delivering RA packets with truncated/offset source address data** to iPXE's NDP processing code, causing the first 2 bytes (`fe80`) to be lost.
2. **iPXE's NDP code is parsing the RA source address incorrectly** when receiving via the NII path vs. a native driver.

## Related

- The Intel E610 (X550-derived, ixgbe family) has no native iPXE driver — PCI IDs `0x57ae`–`0x57b2` are not in `intel.c`, `intelx.c`, `intelxl.c`, or `ice.c`. Adding E610 support to a native driver would also resolve this issue by bypassing NII entirely.
- Dell PowerEdge servers have a history of NII/SNP driver issues with iPXE: #250, #1424, and various mailing list threads.

mcb30:
The EFI device path settings are currently registered as the "netX.dhcp" settings block, in order that they will be automatically overridden if a real DHCP configuration takes place.  This does not work as expected in an IPv6-only network, since the IPv6 configurator will register "netX.ndp" rather than "netX.dhcp".

Fix by registering the EFI device path settings as either "netX.dhcp" or "netX.ndp" based on the first address family encountered within the device path.

Fixes: #1695

### DIFF
--- a/src/include/ipxe/efi/efi_path.h
+++ b/src/include/ipxe/efi/efi_path.h
@@ -47,6 +47,7 @@ extern size_t efi_path_len ( EFI_DEVICE_PATH_PROTOCOL *path );
 extern int efi_path_check ( EFI_DEVICE_PATH_PROTOCOL *path, size_t max );
 extern void * efi_path_mac ( EFI_DEVICE_PATH_PROTOCOL *path );
 extern unsigned int efi_path_vlan ( EFI_DEVICE_PATH_PROTOCOL *path );
+extern unsigned int efi_path_family ( EFI_DEVICE_PATH_PROTOCOL *path );
 extern int efi_path_guid ( EFI_DEVICE_PATH_PROTOCOL *path, union uuid *uuid );
 extern struct uri * efi_path_uri ( EFI_DEVICE_PATH_PROTOCOL *path );
 extern EFI_DEVICE_PATH_PROTOCOL * efi_paths ( EFI_DEVICE_PATH_PROTOCOL *first,
--- a/src/interface/efi/efi_path.c
+++ b/src/interface/efi/efi_path.c
@@ -37,6 +37,7 @@ FILE_SECBOOT ( PERMITTED );
 #include <ipxe/usb.h>
 #include <ipxe/settings.h>
 #include <ipxe/dhcp.h>
+#include <ipxe/ndp.h>
 #include <ipxe/efi/efi.h>
 #include <ipxe/efi/efi_driver.h>
 #include <ipxe/efi/efi_strings.h>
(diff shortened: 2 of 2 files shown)


goranpejovic:
confirmed - this fixes #1695 

thank you so much for the quick fix!

PR #1702, opened 2026-05-01
[usr] Add fetchvar command to store HTTP response in setting

refrigerator18:
Add a new "fetchvar" command that downloads a URI and stores the
response body as a string in a named iPXE setting.  This provides a
generic mechanism for scripts to issue HTTP requests and capture the
response value for use in subsequent commands.

Usage: fetchvar <setting> <uri>

For example:

  fetchvar myip http://httpbin.org/get
  show myip

The implementation follows the nslookup pattern (struct with refcnt,
job and xfer interfaces, setting storage) combined with the downloader
pattern (xferbuf for HTTP data accumulation with redirect support).

A size limit of 8 kB prevents memory exhaustion from accidentally
fetching large files.

Tested on EC2 instances with HTTP and HTTPS endpoints.

### DIFF
--- a/src/config/cloud/general.h
+++ b/src/config/cloud/general.h
@@ -11,3 +11,8 @@
  * incurring unnecessary costs.
  */
 #define POWEROFF_CMD
+
+/* Allow scripts to fetch a URI and store the response body in a
+ * named setting for use in subsequent commands.
+ */
+#define FETCHVAR_CMD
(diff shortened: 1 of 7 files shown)


NiKiZe:
Similar things has been proposed in the past and rejected.
iPXE is not an API client, if you want this, have the server return an ipxe script that sets the settings value.

PR #1713, opened 2026-05-19
[cloud] Add IMDSv2 support for EC2 cloud boot

refrigerator18:
## Summary

AWS now [defaults new instances to IMDSv2-only](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-new-instances.html), with IMDSv1 disabled. Without this change, iPXE's cloud boot script cannot retrieve user-data on these instances.

This series adds IMDSv2 support to iPXE's cloud boot, allowing the boot script to acquire a session token and use it for subsequent metadata requests.

The implementation uses three small, layered changes:

### 1. `fetchvar` command

A new command that fetches a URI and stores the response body in a named setting. This is analogous to `nslookup` (which stores a DNS response in a setting) — we need it because the metadata server returns a raw token string that must be captured for reuse.

```
fetchvar data http://example.com/value
show data
```

### 2. HTTP method override

Adds a `--method` option to `fetchvar` and defines `http_put` in the method table. IMDSv2 requires a PUT request to acquire the session token — without this, `fetchvar` can only do GET.

```
fetchvar --method PUT token http://169.254.169.254/latest/api/token
```

### 3. EC2 metadata headers (`httpec2.c`)

Mirrors the existing `httpgce.c` pattern. Automatically injects:
- `X-aws-ec2-metadata-token-ttl-seconds: 21600` on PUT requests to `169.254.169.254`
- `X-aws-ec2-metadata-token: <value>` on subsequent requests (using the stored setting)

The updated `aws.ipxe` boot script ties it together:
```
fetchvar --method PUT ec2token http://169.254.169.254/latest/api/token
chain -ar http://169.254.169.254/latest/user-data
```

## Design choices

- **IMDSv2 logic lives in the boot script**, not in custom C code
- **`httpec2.c` mirrors `httpgce.c` exactly** — same `__http_request_header` table pattern
- **`fetchvar` and `--method` are generic features** useful beyond EC2

Tested on EC2 instances. Full boot flow working with IMDSv2.

### DIFF
--- a/src/config/cloud/aws.ipxe
+++ b/src/config/cloud/aws.ipxe
@@ -19,6 +19,7 @@ exit
 
 :dhcp_ok
 route ||
+fetchvar --method PUT ec2token http://169.254.169.254/latest/api/token ||
 chain -ar http://169.254.169.254/latest/user-data ||
 ifstat ||
 exit
--- a/src/config/cloud/general.h
+++ b/src/config/cloud/general.h
@@ -11,3 +11,8 @@
  * incurring unnecessary costs.
  */
 #define POWEROFF_CMD
+
+/* Allow scripts to fetch a URI and store the response body in a
+ * named setting for use in subsequent commands.
+ */
+#define FETCHVAR_CMD
--- a/src/config/cloud/general.h
+++ b/src/config/cloud/general.h
@@ -7,6 +7,11 @@
  */
 #define HTTP_HACK_GCE
 
+/* Allow retrieval of metadata from Amazon EC2 Instance Metadata
+ * Service using IMDSv2 session tokens.
+ */
+#define HTTP_HACK_EC2
+
 /* Allow scripts to handle errors by powering down the VM to avoid
  * incurring unnecessary costs.
  */
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -270,6 +270,9 @@ REQUIRE_OBJECT ( shell );
 #ifdef NSLOOKUP_CMD
 REQUIRE_OBJECT ( nslookup_cmd );
 #endif
+#ifdef FETCHVAR_CMD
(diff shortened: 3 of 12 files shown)


NiKiZe:
https://github.com/ipxe/ipxe/pull/1702#issuecomment-4362059450

refrigerator18:
How would you approach this? IMDSv2 requires a PUT to acquire a session token, then that token must be attached as a header to subsequent requests. Having the EC2 metadata server (169.254.169.254) return an iPXE script won't be possible.

mcb30:
This is a fairly large change, and impacts some core parts of iPXE.  As I did explain to you when you first contacted me about this (and as now documented at https://github.com/ipxe/ipxe/blob/master/CONTRIBUTING.md#major-feature-pull-requests), reviewing code is **more** work than writing it in the first place.  In the absence of any funding from Amazon to support this work, I cannot provide any solid timeline on getting this reviewed.

refrigerator18:
Understood, no rush. Just want to flag that this is becoming more urgent for the cloud boot scripts. AWS now defaults new instances to IMDSv2-only, so the current aws.ipxe won't be able to fetch user-data on those instances. Alibaba Cloud has the same token-based metadata pattern (they call it "security hardening mode") and will likely follow the
same path.

mcb30:
I went with a much simpler and more general-purpose design instead:
* The [`params --method`](https://ipxe.org/cmd/params) option can be used to specify PUT or POST for any HTTP request.
* The new [`imgset`](https://ipxe.org/cmd/imgset) command can be used to set a setting value from any downloaded image (not just HTTP)
* The HTTP GCE hack is removed, and the custom headers for GCE, AWS, and AliCloud are constructed within the scripting language instead.

IMDSv2 is now supported by the official images for both AWS and Alibaba Cloud.

Issue #1703, opened 2026-05-02
autoexec.ipxe with IPv6


NiKiZe:
### Discussed in https://github.com/ipxe/ipxe/discussions/1355

<div type='discussions-op-text'>

<sup>Originally posted by **NiKiZe** December  5, 2024</sup>
Observed behaviour 
When booting in UEFI with IPv6 no IP is seen in the boot path, no cwurl can be created and no autoexec.ipxe attempt is done.

Is this just a boot path thing in the firmware I have used, or is it something that maybe haven't been tested?

Could the firmware issues somehow be mitigated if dhcpcache supported dhcpv6?

I would like to have given more details and tested more, but my environment just doesn't want to play nice today.

This was seen while documenting IPv6 boot, notes at
https://gist.github.com/NiKiZe/3da6bbc9bd54c79645d833b9ed271ef9</div>

Other example from the discussion:
dhcp4:

<img width="1920" height="1080" alt="mpv-shot0009" src="https://github.com/user-attachments/assets/8c8b82d2-205b-4f62-b4a8-e2fdd0cb01cd" />

dhcp6:
<img width="1920" height="1080" alt="mpv-shot0010" src="https://github.com/user-attachments/assets/efe89210-9f91-4acc-b76f-3ce8fc8983cf" />
<img width="1920" height="1080" alt="mpv-shot0011" src="https://github.com/user-attachments/assets/be106704-920c-4e69-977c-80b3a8ed89d7" />
<img width="1920" height="1080" alt="mpv-shot0012" src="https://github.com/user-attachments/assets/07e110cb-f330-45f8-a725-cae566184774" />

dhcp6, with following script embedded:
```
#!ipxe

dhcp -c ipv6
chain --replace --autofree tftp://[2001:912:1480:310:2400::1]/autoexec.ipxe
```

<img width="1920" height="1080" alt="mpv-shot0015" src="https://github.com/user-attachments/assets/b7754da1-ebea-4cf8-abf8-725623f6cec6" />

Note: this is real hardware, not a VM.

dseven:
[...]

PR #1718, opened 2026-05-26
[efi] Install duplicate SNP protocols onto dedicated device handle

### LINKED ISSUE #1716: Chaining from iPXE 2.0.0 to WDS does not pass along ${next-server}
~~~
set my-wds-server:ipv4 1.2.3.4
set my-wds-nbp:string smsboot/blabla/x64/wdsmgfw.efi

set netX.dhcp/next-server ${my-wds-server}
set netX.dhcp/filename ${my-wds-nbp}
set proxydhcp/next-server ${my-wds-server}
set proxydhcp/filename ${my-wds-nbp}

chain tftp://${my-wds-server}/${my-wds-nbp}
~~~

In iPXE 1.21.1, the `wdsmgfw.efi` stage of WDS displays "Server IP: 1.2.3.4" and successfully loads its subsequent stages.

In iPXE 2.0.0, the `wdsmgfw.efi` stage of WDS displays "Server IP: 0.0.0.0" and fails to load its next stage (error `0xc000000e`).

Has anyone successfully chained iPXE 2.0.0 to WDS?

mcb30:
Commit 5a17d8d ("[efi] Install protocols onto a dedicated device handle") changed the behaviour for image execution to install the per-image protocols (EFI_SIMPLE_FILE_SYSTEM_PROTOCOL etc) onto a dedicated child device handle, instead of directly onto the parent SNP device handle.  The commit message mentions that if UEFI executables are found that rely upon the SNP protocols being present on the loaded image's device handle (rather than on a parent of that handle), then it may be necessary to install copies of EFI_SIMPLE_NETWORK_PROTOCOL and others onto the child handle.

Testing shows that WDS apparently does require these protocols to be present on the loaded image's device handle.  Install copies of the SNP and NII protocols onto the child handle to support this.

Fixes: #1716 

Debugged-by: Alexander Perlis <alexanderperlis@gmail.com>

### DIFF
--- a/src/image/efi_image.c
+++ b/src/image/efi_image.c
@@ -261,6 +261,20 @@ static int efi_image_exec ( struct image *image ) {
 	DBGC ( image, "EFIIMAGE %s device is %s\n",
 	       image->name, efi_handle_name ( device ) );
 
+	/* Install duplicate SNP protocols, if applicable */
+	if ( ( snpdev != NULL ) &&
+	     ( ( efirc = bs->InstallMultipleProtocolInterfaces (
+			&device,
+			&efi_simple_network_protocol_guid, &snpdev->snp,
+			&efi_nii_protocol_guid, &snpdev->nii,
+			&efi_nii31_protocol_guid, &snpdev->nii,
+			NULL ) ) != 0 ) ) {
+		rc = -EEFI ( efirc );
+		DBGC ( image, "EFIIMAGE %s could not duplicate %s: %s\n",
+		       image->name, netdev->name, strerror ( rc ) );
+		goto err_snp_install;
+	}
+
 	/* Add as a child of the parent device */
 	if ( ( parent != NULL ) &&
 	     ( rc = efi_child_add ( parent, device ) ) != 0 ) {
@@ -461,6 +475,15 @@ static int efi_image_exec ( struct image *image ) {
 	if ( parent )
 		efi_child_del ( parent, device );
  err_child_add:
+	if ( snpdev ) {
+		bs->UninstallMultipleProtocolInterfaces (
+			device,
+			&efi_simple_network_protocol_guid, &snpdev->snp,
+			&efi_nii_protocol_guid, &snpdev->nii,
+			&efi_nii31_protocol_guid, &snpdev->nii,
+			NULL );
+	}
+ err_snp_install:
 	bs->UninstallMultipleProtocolInterfaces (
 			device,
 			&efi_device_path_protocol_guid, devpath,


alexanderperlis:
Alas the first WDS stage is still showing 0.0.0.0. I tried `make` with some `DEBUG` parameters but haven't obtained anything useful worth sharing. Any advice on how to debug this? From the discussion at https://ipxe.org/settings#pxe_nbps I assume WDS is calling PXENV_GET_CACHED_INFO but iPXE is no longer correctly populating the fake proxyDHCP ACK response. Is there a way to confirm what WDS is calling? It would then subsequently be nice to eliminate the opaque WDS stage and replace it with something open from which one can perhaps invoke that same call and see the result. I wonder whether `shellx64.efi` can be told to invoke PXENV_GET_CACHED_INFO?

mcb30:
Thank you for testing.  You can build with various debug options to see what WDS is doing.  The most informative for this purpose is probably to build with `DEBUG=efi_wrap`: this will effectively put an `strace`-like wrapper around WDS to show which boot services API calls it is making.  This includes all calls that WDS makes to locate and open UEFI protocols.

Comparing the traces from the last known-working commit with the traces from current master (and/or this PR) should reveal the problem.  There will be a large volume of debug messages, and WDS will most likely corrupt the display, so you will need to find some way to reliably capture the output (e.g. via `CONSOLE_SERIAL` if you have a real serial port, or `CONSOLE_SERIAL`/`CONSOLE_DEBUGCON` if you can run inside a VM, or `CONSOLE_SYSLOG` as a last resort).

alexanderperlis:
Thanks for the tips. Here's a comparison of a working trace and a broken trace (manually constructed, not a true diff, disregards minor memory location variations):
~~~
---:working trace
+++:broken trace

    WRAP wdsmgfw at base 0x10000000 has protocols:
    HANDLE wdsmgfw LoadedImageDevicePath at 0x45e15d18
    HANDLE wdsmgfw LoadedImage at 0x45e16d40
    WRAP wdsmgfw parent snponly.efi
--- WRAP wdsmgfw device PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1)
+++ WRAP wdsmgfw device PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1)/Uri(http://1.2.3.4/wdsmgfw.efi)
    WRAP wdsmgfw file wdsmgfw.efi
    WRAP installed local wrappers
    GetVariable ( GlobalVar:SetupMode, 0x1 ) = 0 ( 0x1, 48df41d8 : 00 : .) -> 0x1004bb15
    GetVariable ( GlobalVar:SecureBoot, 0x1 ) = 0 ( 0x1, 48df41d0 : 00 : .) -> 0x1004bb53
    HandleProtocol ( wdsmgfw, LoadedImage ) = 0 ( 0x45e16d40 ) -> 0x1001cfdf
--- HandleProtocol ( PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1), DevicePath ) = 0 ( 0x45885a08 ) -> 0x1001d026
+++ HandleProtocol ( PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1)/Uri(http://1.2.3.4/wdsmgfw.efi), DevicePath ) = 0 ( 0x45884c08 ) -> 0x1001d026
    AllocatePages ( Address, LoaderData, 0x1, 0x102000 ) = 0 ( 0x102000 ) -> 0x1001d050
--- LocateDevicePath ( PxeBaseCode, PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1) ) = 0 ( , PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1) ) -> 0x1001d4a9
--- OpenProtocol ( PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1), PxeBaseCode, <NULL>, <NULL>, G ) = 0 ( 0x4588b6a0 ) -> 0x1001d4e8
    GetMemoryMap ( 0x0, 0x0 ) = BUFFER_TOO_SMALL ( 0x2a00, 0x10dc5, 0x30, v1 ) -> 0x1001f7e2
    AllocatePages ( AnyPages, LoaderData, 0x4, 0x0 ) = 0 ( 0x45cd6000 ) -> 0x1001f880
    FreePages ( 0x45cd6000, 0x4 ) = 0 -> 0x1001f903
    AllocatePages ( Address, LoaderData, 0x3, 0x45cd6000 ) = 0 ( 0x45cd6000 ) -> 0x1001f880
    GetMemoryMap ( 0x3000, 0x45cd6000 ) = 0 ( 0x2580, 0x1103c, 0x30, v1, ... )
    FreePages ( 0x45cd6000, 0x3 ) = 0 -> 0x1001f903
    AllocatePages ( Address, LoaderCode, 0x10, 0x103000 ) = 0 ( 0x103000 ) -> 0x1001f880
    SetWatchdogTimer ( 0s, 0x0, 0x0, 0x0 ) = 0 -> 0x1001e30c
    OpenProtocol ( ConIn, SimpleTextInputEx, wdsmgfw, <NULL>, G ) = 0 ( 0x4c9e9230 ) -> 0x1001fde1
    AllocatePages ( Address, LoaderCode, 0x80, 0x113000 ) = 0 ( 0x113000 ) -> 0x1001f880
    LocateHandle ( ByProtocol, Tcg2, 0x0, 0 ) = BUFFER_TOO_SMALL ( 8 ) -> 0x1001fc67
    LocateHandle ( ByProtocol, Tcg2, 0x0, 8 ) = 0 ( 8, { UNKNOWN<0x460d7a18,Tcg2> } ) -> 0x1001fd01
    OpenProtocol ( UNKNOWN<0x460d7a18,Tcg2>, Tcg2, wdsmgfw, <NULL>, G ) = 0 ( 0x4c5f3390 ) -> 0x1001fde1
    GetVariable ( e20939be-32d4-41be-a150-897f85d49829:MemoryOverwriteRequestControl, 0x1 ) = 0 ( 0x1, 48df3930 : 10 : . ) -> 0x1001f2e6
    SetVariable ( e20939be-32d4-41be-a150-897f85d49829:MemoryOverwriteRequestControl, 0x07, 48df3938 : 11 : . ) = 0 -> 0x1001f3ea
--- OpenProtocol ( wdsmgfw, LoadedImage, wdsmgfw, <NULL>, G ) = 0 ( 0x45e16d40 ) -> 0x1001fde1
--- OpenProtocol ( PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1), DevicePath, wdsmgfw, <NULL>, G ) = 0 ( 0x45885a08 ) -> 0x1001fde1
--- LocateDevicePath ( PxeBaseCode, PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1) ) = 0 ( , PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1) ) -> 0x1001fb8c
--- OpenProtocol ( PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1), PxeBaseCode, wdsmgfw, <NULL>, G ) = 0 ( 0x4588b6a0 ) -> 0x1001fde1
--- OpenProtocol ( PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1), SimpleNetwork, wdsmgfw, <NULL>, G ) = 0 ( 0x45886e50 ) -> 0x1001fde1
--- CloseProtocol ( PciRoot(0x1)/Pci(0x1F,0x6)/MAC(D4A2CDCDB494,0x0)/MAC(D4A2CDCDB494,0x1), DevicePath, wdsmgfw, <NULL> ) = 0 -> 0x1001fe8b
--- CloseProtocol ( wdsmgfw, LoadedImage, wdsmgfw, <NULL> ) = 0 -> 0x1001fe8b
    GetVariable ( GlobalVar:SetupMode, 0x1 ) = 0 ( 0x1, 48df39f8 : 00 : . ) -> 0x1001f2e6
    GetVariable ( GlobalVar:SecureBoot, 0x1 ) = 0 ( 0x1, 48df39f0 : 00 : . ) -> 0x1001f2e6
    AllocatePages ( Address, LoaderCode, 0x1, 0x100000 ) = 0 ( 0x100000 ) -> 0x1001f880
    AllocatePages ( Address, LoaderCode, 0x10, 0x193000 ) = 0 ( 0x193000 ) -> 0x1001f880
    AllocatePages ( Address, LoaderCode, 0x10, 0x1a3000 ) = 0 ( 0x1a3000 ) -> 0x1001f880
    AllocatePages ( Address, LoaderCode, 0x10, 0x1b3000 ) = 0 ( 0x1b3000 ) -> 0x1001f880
    AllocatePages ( Address, LoaderCode, 0x10, 0x1c3000 ) = 0 ( 0x1c3000 ) -> 0x1001f880
--- CreateEvent ( 0x80000000, Callback, 0x0, 0x0 ) = 0 ( 0x45d1be98 ) -> 0x1001f55d
--- SetTimer ( 0x45d1be98, Relative, 2.000000000s ) = 0 -> 0x1001f5ce
--- CloseEvent ( 0x45d1be98 ) = 0 -> 0x1001f4aa
--- [... prior 3 lines repeat while WDS loads the next stage ...]
~~~

In the working trace we see WDS makes several additional calls. Considering just the first two of those additional calls (`LocateDevicePath` and `OpenProtocol` near the top), my guess is something about the nature of the child device path is causing WDS to not bother making those calls. Could it be WDS is examining the device path leaf entry and concluding (falsely) from the Type/SubType there's no reason to bother searching for `PxeBaseCode`? Indeed, I tried replacing the `Uri(...)` leaf with a fake `MAC(...)` leaf, and lo and behold, WDS then makes the calls (!), although later on (without your proposed patch) it fails to access `SimpleNetwork`:
~~~
OpenProtocol ( PciRoot()/Pci()/MAC()/MAC()/MAC(), SimpleNetwork, wdsmgfw, ... ) = UNSUPPORTED
~~~
The possibility stated in 5a17d8d is indeed occurring: WDS does not use `LocateDevicePath` to find `SimpleNetwork`, instead expects to find it on the leaf. (I didn't test whether it needs `NII` or `NII31` on the leaf.)

Cursory examination of WDS disassembly reveals it indeed walks the device path and actually stops just before reaching the first File Path node, if one is found, otherwise stops just before the end of the device path. The code branches based on the type of the prior node. When the prior node is `MAC()`, `IPv4()`, or `IPv6()`, WDS takes a PXE branch which in turn calls `LocateDevicePath` for `PxeBaseCode`. A different code branch is taken when the prior node is `Uri()`, and a different code branch yet again for everything else.

Conclusions thus far:
- To work with WDS, the device path must end in a network node, or more precisely end in a network node prior to the first File Path if the latter exists.
- The eventual call to `OpenProtocol(..., SimpleNetwork)` does not call `LocateDevicePath()`. Thus `SimpleNetwork` must be attached to the device path leaf.

This now brings up the question: what is 5a17d8d trying to accomplish (I don't understand the concern stated in the commit message)?, and might there be a different way to do so?

I do understand the desire to communicate `cwduri` to a chainloaded iPXE. Is the device path itself the only way (or best way) to do that? Does UEFI support custom protocols so that the chainloaded iPXE could do something like `LocateDevicePath(iPxeBase, ...)`?

If the child device path approach from 5a17d8d remains the way to go, one way to leverage that approach while also satisfying strangely behaved EFIs is to make a *grandchild* device path:
~~~
PciRoot()/Pci()/MAC()/MAC()/Uri()/MAC()
~~~
or even
~~~
PciRoot()/Pci()/MAC()/MAC()/FilePath('')/Uri()
~~~
I didn't test these, but I believe either would work with WDS. Unclear whether it would confuse other EFIs. Anyhow you would still need to install various protocols onto the leaf. Rather than doing that piecemeal in patches over time as users discover issues with EFIs in the wild, I suggest to systematically loop over *all* protocols on the grandparent MAC() and install each one onto the leaf. In case that's an issue with speed or memory, I ask again: are child device nodes the best solution?

Yet another alternative is to make the child device node construction optional, either as an opt-in:
~~~
chain --pass-cwduri-as-child-device-path ipxe.efi
~~~
or as an opt-out:
~~~
chain --do-not-construct-child-device-path http://1.2.3.4/wdsmgfw.efi
~~~

There's a lot to think about here. I have no opinions on the right or wrong way to go about things (this excursion is my first dive into iPXE, UEFI, WDS, ...). I'm just reporting what does or does not work with WDS. I also recognize WDS might be considered broken and the iPXE project might not have a goal of being able to chain to broken things. My own motivation is that I will likely only get permission to make iPXE the primary PXE on my subnet if I can have a menu entry that successfully chains to the central PXE (which happens to currently be WDS, but really could be *anything* and I'd like to chain to it whatever it might be).

Happy to carry out additional testing after you have had time to consider all this.

NiKiZe:
This is not only for iPXE, but other loaders as well that might want an url when booted over http, Should we conclude that WDSs http boot support is broken? (there is now closed issues about iPXE loading iPXE not downloading autoexec.ipxe)

I must again ask why not to use wimboot to get into your WDS environment? even if I agree that this is a regression, I think your environment would get an upgrade by using wimboot, I understand that it is not acceptable with the current requirement, but if you  test it and it improves boot time into WDS, maybe that requirement is allowed to be adjusted ;)

mcb30:
> Cursory examination of WDS disassembly reveals it indeed walks the device path and actually stops just before reaching the first File Path node, if one is found, otherwise stops just before the end of the device path. The code branches based on the type of the prior node. When the prior node is `MAC()`, `IPv4()`, or `IPv6()`, WDS takes a PXE branch which in turn calls `LocateDevicePath` for `PxeBaseCode`. A different code branch is taken when the prior node is `Uri()`, and a different code branch yet again for everything else.
> 
> Conclusions thus far:
> 
>  * To work with WDS, the device path must end in a network node, or more precisely end in a network node prior to the first File Path if the latter exists.
> 
>  * The eventual call to `OpenProtocol(..., SimpleNetwork)` does not call `LocateDevicePath()`. Thus `SimpleNetwork` must be attached to the device path leaf.

Wonderful; thank you so much for going this far in the debugging and analysis.

> This now brings up the question: what is [5a17d8d](https://github.com/ipxe/ipxe/commit/5a17d8daf3c5247a95df88bcd651b61675058592) trying to accomplish (I don't understand the concern stated in the commit message)?, and might there be a different way to do so?

That commit is following on from the change introduced in the previous commit https://github.com/ipxe/ipxe/commit/910d9c5a9:

> ```
> [efi] Allow executing images even with no open network devices
>
> We need a device handle from which to nominally load an EFI image.  We
> currently rely on using the most recently opened network device's SNP
> device handle, in the same way that we use the most recently opened
> network device when loading a BIOS PXE NBP image.  If there is no most
> recently opened network device, then we cannot execute an EFI image.
> ```

The purpose of that commit is to allow iPXE to execute an EFI image even if there are no currently opened network devices.  For example, the user may be loading iPXE from a USB key and have an `autoexec.ipxe` script that attempts to boot from the network, falling back to executing a copy of the UEFI shell loaded from the same USB key via e.g. `chain file:/UefiShell.efi` if no network device is found.  Before commit https://github.com/ipxe/ipxe/commit/910d9c5a9, this cannot be done.

That commit then introduces an additional failure possibility, which is the scenario described in commit https://github.com/ipxe/ipxe/commit/5a17d8daf: there may already be other protocols present on the fallback handle that we use (which we did not construct ourself).  Creating a child device handle avoids this problem.

> I do understand the desire to communicate `cwduri` to a chainloaded iPXE. Is the device path itself the only way (or best way) to do that? Does UEFI support custom protocols so that the chainloaded iPXE could do something like `LocateDevicePath(iPxeBase, ...)`?

The `Uri()` device path component is already used in some other scenarios (e.g. using UEFI HTTP boot to load iPXE), and so this is the most natural way to communicate the URI to a subsequent iPXE.

> If the child device path approach from [5a17d8d](https://github.com/ipxe/ipxe/commit/5a17d8daf3c5247a95df88bcd651b61675058592) remains the way to go, one way to leverage that approach while also satisfying strangely behaved EFIs is to make a _grandchild_ device path:
> 
> ```
> PciRoot()/Pci()/MAC()/MAC()/Uri()/MAC()
> ```
> 
> or even
> 
> ```
> PciRoot()/Pci()/MAC()/MAC()/FilePath('')/Uri()
> ```

This gets increasingly messy, and I would prefer to find a cleaner solution.

> I didn't test these, but I believe either would work with WDS. Unclear whether it would confuse other EFIs. Anyhow you would still need to install various protocols onto the leaf. Rather than doing that piecemeal in patches over time as users discover issues with EFIs in the wild, I suggest to systematically loop over _all_ protocols on the grandparent MAC() and install each one onto the leaf. In case that's an issue with speed or memory, I ask again: are child device nodes the best solution?

Looping over all protocols on the parent device node and installing them all onto the child would negate the whole point of having the child (which is to avoid clashes between protocols that already exist on the parent and protocols that we want to install ourselves).

> Yet another alternative is to make the child device node construction optional, either as an opt-in:
> 
> ```
> chain --pass-cwduri-as-child-device-path ipxe.efi
> ```
> 
> or as an opt-out:
> 
> ```
> chain --do-not-construct-child-device-path http://1.2.3.4/wdsmgfw.efi
> ```

Definitely not this approach.  We put a lot of effort into making sure that end users don't have to make this kind of confusing and apparently arbitrary choice.

> There's a lot to think about here. I have no opinions on the right or wrong way to go about things (this excursion is my first dive into iPXE, UEFI, WDS, ...). I'm just reporting what does or does not work with WDS. I also recognize WDS might be considered broken and the iPXE project might not have a goal of being able to chain to broken things. My own motivation is that I will likely only get permission to make iPXE the primary PXE on my subnet if I can have a menu entry that successfully chains to the central PXE (which happens to currently be WDS, but really could be _anything_ and I'd like to chain to it whatever it might be).

We definitely should be able to chain into WDS: that is a totally valid use case.  More thought is probably required to plan out the best way to accommodate all of these requirements.

As a first thought, it might be sensible to choose the child device path component type based on the image URI scheme, with a TFTP URI resulting in a `.../IPv4()` or `.../IPv6()` component, and all other protocols resulting in a `.../Uri()` component.  That would more closely match what UEFI firmware tends to provide for TFTP and HTTP boot scenarios.

In the TFTP case, a chainloaded iPXE will already have another way to obtain the effective boot URI (using the `EFI_PXE_BASE_CODE_PROTOCOL` to obtain the cached DHCP packets), and so this should not be a regression for that use case.

I will think about this a bit longer before implementing anything, and would be happy to take further input.

mcb30:
One additional thought: if we do attempt to roughly match EDK2 by using `IPv4()`/`IPv6()` device nodes, then we should check what EDK2 does in terms of which protocols it copies onto its own child nodes.  (I haven't checked yet to see if it copies Nii/Nii31, or only Snp.)

alexanderperlis:
Update on WDS (at least the particular WDS version I'm talking to):
- WDS only needs `SimpleNetwork`. After removing the `Nii`/`Nii31` portion of this proposed patch, I successfully completed an end-to-end Windows installation with WDS.
- For this test I also switched from appending a fake `MAC()` node (37 bytes) to instead inserting an empty File Path node (just 4 bytes) immediately before the `Uri()` leaf. I like that more, and WDS remained happy.

@NiKiZe: I appreciate the tip to explore `wimboot` and will certainly do so — you're right, offering a super fast Windows installation procedure could make this iPXE server useful not just to me for Linux but also attractive to the WDS folks here. However, offering just `wimboot` alone feels fragile because it will likely need maintenance whenever WDS is modified. I might not be available at that moment and that could doom this iPXE. Thus it behooves me to offer a pass-through PXE-to-WDS menu option as a fall-back. They also mentioned the existence of certain multi-stage installations that involve a post-Windows client reboot-to-PXE and a *modified* follow-up behavior (unclear whether this means a modified proxyDHCP response or perhaps `wdsmgfw.efi` gathers a modified WDS response) that will send that client directly to the next stage of the installation. From what I understand of `wimboot` it will be helpful for the initial Windows installation stage but any PXE-based follow-up stages will need some kind of PXE pass-through.

@mcb30: I'm still trying to wrap my head around the child device path situation. I take it the “protocols we want to install” are somehow *different* from the identically-named protocols that might already be in place (and thus are now an obstacle)? I furthermore take it one should not delete-and-replace the obstacle protocol from the existing device path? I take it the existing device path might belong to someone else (handed to us by a calling EFI?) and should not be trampled? But then how was it ever okay for iPXE to install additional protocols on the original device path (prior to the commit that introduced the child device path)?

I'm curious: can one somehow make a copy of a device path (either the network one, otherwise the fallback one), modify the original (add new protocols, replace obstacle protocols), later restore it? [I'm not proposing to do this—it wouldn't pass along the URI—just wondering whether this type of messing around is okay.]

> Looping over all protocols on the parent device node and installing them all onto the child would negate the whole point of having the child (which is to avoid clashes between protocols that already exist on the parent and protocols that we want to install ourselves).

How about: first install the protocols you want to install, then loop over the parent to install anything else that's still missing...

Or what about this crazy idea: install solely what you want to install, do not preemptively install anything additional (not even `SimpleNetwork`), instead do *just-in-time* installation inside `efi_open_protocol_wrapper`: based on the result of `bs->OpenProtocol` it could install the missing protocol and again try `bs->OpenProtocol`. (This approach would be bad if there exist EFIs that first try to open an older protocol and only try a better protocol if the older one fails to open. But I imagine EFIs would typically first go for what they want, falling back to something less desirable only when necessary; in that case there's no problem.)

> As a first thought, it might be sensible to choose the child device path component type based on the image URI scheme, with a TFTP URI resulting in a .../IPv4() or .../IPv6() component, and all other protocols resulting in a .../Uri() component. That would more closely match what UEFI firmware tends to provide for TFTP and HTTP boot scenarios.

Might work fine for many yet doesn't *feel right* because a file transfer method indicates nothing about the capabilities of the file being transferred. Someone can add HTTP support to their TFTP server merely to provide faster transfer speed, no change to the files being transferred. I'd prefer a configuration option that is more closely tied to the hidden behavior.

For example, perhaps you could trigger on whether `next-server` has been set by the user. Wouldn't that indicate a network situation, furthermore that the user wants to communicate a piece of information via cached DHCP packets? That's a situation in which you could always stick with the network device path (error if it doesn't exist), no looking at a fall back device path, no creating a child device path, and no need to install `SimpleNetwork` (whether preemptively or just-in-time via the wrappers)...

> In the TFTP case, a chainloaded iPXE will already have another way to obtain the effective boot URI (using the `EFI_PXE_BASE_CODE_PROTOCOL` to obtain the cached DHCP packets), and so this should not be a regression for that use case.

Wait, are you using the `Uri()` leaf to communicate the original boot URI, or to communicate the chain load URI (the starting `cwuri` for the newly loaded image)? It seems the cached DHCP packet would hold the wrong URI unless the `chain` was preceded by `set filename`. Someone chaining from iPXE to iPXE shouldn't have to do anything additional to obtain or maintain desirable behavior.

mcb30:
> Update on WDS (at least the particular WDS version I'm talking to):
> 
>     * WDS only needs `SimpleNetwork`. After removing the `Nii`/`Nii31` portion of this proposed patch, I successfully completed an end-to-end Windows installation with WDS.

Great, thank you for testing that.

>     * For this test I also switched from appending a fake `MAC()` node (37 bytes) to instead inserting an empty File Path node (just 4 bytes) immediately before the `Uri()` leaf. I like that more, and WDS remained happy.

This is definitely not the solution we're going for, sorry.  It creates a very unnatural and surprising device path, and there's a high chance that it's going to confuse something, even if WDS is happy with it.

> @mcb30: I'm still trying to wrap my head around the child device path situation. I take it the “protocols we want to install” are somehow _different_ from the identically-named protocols that might already be in place (and thus are now an obstacle)?

Yes.  For example: we install an EFI_SIMPLE_FILE_SYSTEM_PROTOCOL to provide a virtual filesystem interface to all of the images downloaded by iPXE.  In the case where `last_opened_snpdev()` returns NULL and we fall back to using the handle of the device that iPXE was loaded from (instead of one of our SNP device handles), there may already be a completely unrelated EFI_SIMPLE_FILE_SYSTEM_PROTOCOL instance on that handle (e.g. if iPXE was loaded from a USB key).

> I furthermore take it one should not delete-and-replace the obstacle protocol from the existing device path?

Correct.  That way lies complexity and madness.

> I take it the existing device path might belong to someone else (handed to us by a calling EFI?) and should not be trampled? But then how was it ever okay for iPXE to install additional protocols on the original device path (prior to the commit that introduced the child device path)?

It wasn't previously possible.  Those two commits were pushed as a pair.  Prior to commit https://github.com/ipxe/ipxe/commit/910d9c5a9 we could only ever have used the SNP device handle (which is one that we created ourselves, and so can do whatever we want to).  That commit is the one that opened up the possibility of falling back to a handle that might have already had clashing protocols installed on it.  (Our clashing protocols would have cleanly failed to install, so this commit didn't introduce any kind of regression, it just changed the error message when attempting to boot without any open network devices.)  The next commit https://github.com/ipxe/ipxe/commit/5a17d8daf completed the feature of being able to boot without an open network device (by creating the child handle, which can never have clashing protocols).

> I'm curious: can one somehow make a copy of a device path (either the network one, otherwise the fallback one), modify the original (add new protocols, replace obstacle protocols), later restore it? [I'm not proposing to do this—it wouldn't pass along the URI—just wondering whether this type of messing around is okay.]

Not a meaningful question, sorry.  The device path is itself just a protocol installed on a handle.  It's not meaningful to talk about adding protocols to a device path, only to a handle.

> > Looping over all protocols on the parent device node and installing them all onto the child would negate the whole point of having the child (which is to avoid clashes between protocols that already exist on the parent and protocols that we want to install ourselves).
> 
> How about: first install the protocols you want to install, then loop over the parent to install anything else that's still missing...

No.  I'd much rather have something deterministic that I can reason about and that will behave consistently between different systems.

> Or what about this crazy idea: install solely what you want to install, do not preemptively install anything additional (not even `SimpleNetwork`), instead do _just-in-time_ installation inside `efi_open_protocol_wrapper`: based on the result of `bs->OpenProtocol` it could install the missing protocol and again try `bs->OpenProtocol`. (This approach would be bad if there exist EFIs that first try to open an older protocol and only try a better protocol if the older one fails to open. But I imagine EFIs would typically first go for what they want, falling back to something less desirable only when necessary; in that case there's no problem.)

Completely impractical, sorry.  UEFI is not designed to operate this way, and it would break huge numbers of use cases (e.g. using `LocateHandleBuffer()` to get a list of all handles supporting a given protocol).

> > As a first thought, it might be sensible to choose the child device path component type based on the image URI scheme, with a TFTP URI resulting in a .../IPv4() or .../IPv6() component, and all other protocols resulting in a .../Uri() component. That would more closely match what UEFI firmware tends to provide for TFTP and HTTP boot scenarios.
> 
> Might work fine for many yet doesn't _feel right_ because a file transfer method indicates nothing about the capabilities of the file being transferred. Someone can add HTTP support to their TFTP server merely to provide faster transfer speed, no change to the files being transferred. I'd prefer a configuration option that is more closely tied to the hidden behavior.

It feels ugly because it's designed to match what UEFI firmware itself does, and the UEFI design is itself very ugly.  But it's almost certainly the least surprising, which is what matters in terms of interoperability.

> For example, perhaps you could trigger on whether `next-server` has been set by the user. Wouldn't that indicate a network situation, furthermore that the user wants to communicate a piece of information via cached DHCP packets?

No.  Some DHCP servers will always set `next-server`.  iPXE will happily accept a situation in which `next-server` exists but `filename` contains a full URL (with a scheme).

> > In the TFTP case, a chainloaded iPXE will already have another way to obtain the effective boot URI (using the `EFI_PXE_BASE_CODE_PROTOCOL` to obtain the cached DHCP packets), and so this should not be a regression for that use case.
> 
> Wait, are you using the `Uri()` leaf to communicate the original boot URI, or to communicate the chain load URI (the starting `cwuri` for the newly loaded image)? It seems the cached DHCP packet would hold the wrong URI unless the `chain` was preceded by `set filename`. Someone chaining from iPXE to iPXE shouldn't have to do anything additional to obtain or maintain desirable behavior.

It's currently used to communicate the URI of the image being executed (which will become the `cwuri` if the thing being executed is iPXE).

In the common DHCP/TFTP autoboot case, this will be the same thing as the filename in the cached DHCP packet.

Prior to commit https://github.com/ipxe/ipxe/commit/5a17d8daf, using `set netX/filename` or similar would have been the only way for someone to pass a modified URI through to the executed image.  This proposed change will effectively restore that behaviour for anyone using TFTP.

It would be nice if UEFI hadn't decided to wholesale import all the design limitations of the original PXE and had instead settled on consistently using URIs to describe network locations, but that's not a choice we get to make.  Matching the behaviour of EDK2 (i.e. providing `IPv4()` or `IPv6()` instead of `Uri()` for TFTP boot flows) is the least bad option, as far as I can see.

alexanderperlis:
I appreciate the discussion and your indulgence as I struggle with the influx of unfamiliar information. Here's the ground truth as I currently understand it:
- You own the snp handle. You create it when there is a network device. You can do whatever you want with it.
- You do *not* own the "fallback handle", which is instead owned by whoever invoked you. They passed it to you. You can access it, but you can't make changes to it.
- You can create additional handles.
- You need to pass a handle to the next stage, but you want to install your own file system on the handle being passed on.
- You would like to also somehow communicate to the next stage the URI that invoked them.
- Newer UEFI HTTP boot methods use a Uri() leaf node to communicate the URI.
- For maximum future compatibility iPXE should similarly use a Uri() leaf node to communicate the URI.
- Backwards compatibility issues: there exist "legacy PXE-style" next stages that expect snp-style handles (and will examine the cached DHCP packet) but will choke on URI-style handles.
- (Added by me:) One should be able to load a legacy PXE-style next stage via HTTP or from local USB file system.

Broadly speaking, the various nuanced ways of chaining (endless combinations of vendor hardware behavior and DHCP behavior) fall into two essentially distinct (and incompatible) types:

**Legacy PXE-style chaining**: `next-server` and `filename` (just file name, not a complete URI) are embedded in cached DHCP packet; typically (but not always) limited to network-based chaining; typically (but not always) TFTP is the transfer method; device handle given to next image is the snp device handle when available (no Uri() leaf appended), otherwise is either the fallback handle (without custom protocols) or a child handle (with custom protocols but no Uri() leaf appended).

**Modern URI chaining**: can be network or local; unlikely to use TFTP (but could); no need for cached DHCP packet (all needed info appears already in URI); the URI is communicated in Uri() leaf node.

Your proposed *heuristic* uses the chaining scheme (whether or not it is `tftp`) to automatically choose the chaining type. It makes sense to have this heuristic so that the user is not *required* to specify the chaining type. However, I contend you should *support* an optional explicit choice for advanced users in peculiar situations.

- `chain tftp:...` could secretly imply `chain --type=legacy-pxe tftp:...`
- `chain http:...` could secretly imply `chain --type=uri http:...`
- `chain file:...` could secretly imply `chain --type=uri file:...`

With support for an optional explicit choice, iPXE could participate in unusual magic tricks, such as having a large legacy stage available on one's own iPXE server or on a USB stick, load it locally and quickly, while still giving it the legacy-PXE environment (snp handle, cached DHCP) it might expect for something.

~~~
set netX/next-server 1.2.3.4   # goes into fake DHCP, needed by legacy stage
chain --type=legacy-pxe http://${ipxe-server}/legacystage.efi   # quickly load legacy stage via HTTP
chain --type=legacy-pxe file:legacystage.efi   # quickly load legacy stage from USB
~~~

Without support for `--type=legacy-pxe`, the heuristic would have gone for modern URI-style chaining, meaning a Uri() leaf node would be created and the legacy stage would choke. One would be limited to pulling the legacy stage over the network and only via slow TFTP. This would be a new limitation: in the past (before the introduction of the modern URI chaining type), legacy stages could be loaded quickly over HTTP.

**Notice:** You can be both modern and fully backwards-compatible. You will not have to create `IPv4()` nodes or try to mimic EDK2 behavior or even install `SimpleNetwork` anywhere. Legacy images can access legacy PXE-style behavior. Everyone else enjoys modern URI behavior.

mcb30:
> Your proposed _heuristic_ uses the chaining scheme (whether or not it is `tftp`) to automatically choose the chaining type. It makes sense to have this heuristic so that the user is not _required_ to specify the chaining type. However, I contend you should _support_ an optional explicit choice for advanced users in peculiar situations.
> 
>     * `chain tftp:...` could secretly imply `chain --type=legacy-pxe tftp:...`
> 
>     * `chain http:...` could secretly imply `chain --type=uri http:...`
> 
>     * `chain file:...` could secretly imply `chain --type=uri file:...`
> 
> With support for an optional explicit choice, iPXE could participate in unusual magic tricks, such as having a large legacy stage available on one's own iPXE server or on a USB stick, load it locally and quickly, while still giving it the legacy-PXE environment (snp handle, cached DHCP) it might expect for something.

It is superficially attractive, but the problem with this is that it goes against literal decades of trying **extremely hard** to make iPXE always just do the right thing by default, and not force users to make explicit choices where those choices are not actually necessary.  Other bootloaders tend to force unnecessary choices on the user.  For example:

* UEFI forces the user to decide between TFTP vs. HTTP and IPv4 vs. IPv6 in advance, and these choices must match how the network that you are booting from is configured.  iPXE just boots whatever the network tells it to boot, and doesn't force the user into making (or understanding, or even caring about the existence of) this unnecessary choice.

* U-Boot may require the user to specify the kernel load address, and this address must not clash with an address that U-Boot is already using.  iPXE just determines and uses the correct address automatically, and doesn't force the user into making (or understanding, or even caring about the existence of) this unnecessary choice.

* Several bootloaders force the user to state the type of executable file that they want to boot, or to use a particular naming convention.  iPXE just determines the file type based on its content, and doesn't force the user into making (or understanding, or even caring about the existence of) this unnecessary choice.

I appreciate that adding just one single option is relatively simple, and provides an easy way out of this problem.  Taking the easy way out is always easy the first time.  I have to think about what would happen after taking the easy way out 20 or 30 times.

If we really can't find any other way to support this, then it may have to be done.  But the design you are suggesting would be the last resort, to be used only as an admission of failure to come up with a clean solution.  It would irreversibly add complexity into a core part of the code and documentation.

> **Notice:** You can be both modern and fully backwards-compatible. You will not have to create `IPv4()` nodes or try to mimic EDK2 behavior or even install `SimpleNetwork` anywhere. Legacy images can access legacy PXE-style behavior. Everyone else enjoys modern URI behavior.

We should mimic the EDK2 device paths, because those are the least surprising for any other UEFI software to encounter.  Being minimally surprising is a desirable goal.

alexanderperlis:
> [...] make iPXE always just do the right thing by default, and not force users to make explicit choices [...]

We might be saying the same thing. I was in agreement with your heuristic, which almost always does the right thing by default, thereby indeed does not generally force users to make explicit choices. I was just asking to also have support for an *optional* config parameter to handle rare edge cases.

But if you're saying you'd like to find a solution that *absolute always* does the right thing automatically for *absolutely all* situations, while *also* satisfying the list of “ground truths” I stated, I believe my prior message is essentially a proof of impossibility. *However, if you're willing to ever-so-slightly relax those ground truth requirements,* here follows an approach that seems fully backwards-compatible with prior iPXE behavior while *also* supporting chain-to-USB and *also* providing `Uri()` device paths in *many* circumstances.

Important aside: Is it even correct to use `Uri()` for chain-to-USB? I accept your claim that UEFI HTTPs boot [which I currently cannot test in my environment] passes in a handle ending in `Uri()`. But that type of handle might be *specifically* and *solely* intended for just HTTPs boot mode. If the user does UEFI USB boot or UEFI HD boot, isn't the device path something like `PciRoot() / Pci() / USB()|Sata() / HD() / Path(\EFI\Boot\...)` ? Thus, when iPXE chains to `file:...`, instead of making a child handle ending in `Uri()`, shouldn't it make a handle that identifies the drive device and ends in `Path()`?

iPXE is wonderful because `chain` just takes a URI whose scheme could be `file` or `http` or `tftp`. Everything is simply a URI! It's so simple and elegant! But iPXE must operate within the UEFI universe wherein the `Uri()` device path should be used for only one of several UEFI boot methods.

Here then is the proposed way for iPXE to process a `chain ...` command:
- Set a legacy-mode flag to true.
- *Load* the next stage into memory (whether over network or from local filesystem). If it cannot be loaded that's an error.
- Do a libmagic-style fingerprinting to determine whether the next stage is itself another iPXE (and of a sufficiently new vintage to understand `Uri()` or `Path()` leafs) — in this case, turn off the legacy-mode flag.
- If legacy-mode flag is on but snp device handle does not exist [prior iPXE would be stuck at this point?], turn off legacy-mode flag.
- If legacy-mode flag is still on (which at this point implies snp device handle exists), regardless of whether chaining to network location or local file, use the snp device handle as the handle to pass along to the next stage (no creating a child handle, and no appending `Uri()` to the handle). [This preserves prior behavior for iPXE which was *always* in “legacy PXE-style” mode; legacy stages can be loaded via HTTP or even from local USB.]
- If legacy-mode flag is off (we might also get here from some of the conditions above that turned off that mode):
  - If chaining to a network-location scheme, create a `Uri()` device path.
  - If chaining to `file:...`, create an appropriate device path handle (with `USB()` or `Sata()` or ...) ending in `Path()`.
- Prepare the handle for the next stage (add custom file system protocol if we have one).
- Transfer control to the loaded stage.

In pseudo-code:
~~~
next_handle = snphandle || 0;
image = load_into_memory || error;
image_is_ipxe = fingerprint(image);
if (image_is_ipxe): next_handle = 0;
if (!next_handle):
  if (scheme == "file"):
    next_handle = create_handle(`USB()|Sata()`, `HD()`, `Path()`);
  else:
    next_handle = create_handle(`Uri()`);
if (have_own_filesystem): append(next_handle, EFI_SIMPLE_FILE_SYSTEM_PROTOCOL);
exec(image, next_handle);
~~~

The intent is to correctly handle past situations as well as chain-to-USB and chain-to-iPXE-anywhere. The user can also use modern UEFI HTTPs boot to quickly get to an iPXE menu system and/or use UEFI USB boot (if their network DHCP does not have correct values for network booting) to chain to that same HTTPs iPXE menu system, and from there can have menu entries that chain to legacy remote PXE servers (even ones that support HTTP but do not support `Uri()`).

Alas this approach is suboptimal for chaining to modern non-iPXE stages:
- `chain http://some-server/some-non-ipxe-modern-uri-aware-stage.efi`
- `chain file:some-non-ipxe-modern-filepath-aware-stage.efi`
- (The modern non-iPXE stage will receive a `MAC()` handle instead of a `Path()` or `Uri()` handle.)

That edge case isn't necessarily broken — the next stage *could* be okay with the `MAC()` and *could* work perfectly well or might have some feature limitations. This edge case is *unavoidable* because: (1) iPXE cannot auto-detect whether a non-iPXE stage is modern or legacy; (2) we're seeking an approach that does not introduce additional things for the iPXE user to specify. We either break backwards compatibility or we give out a legacy `MAC()` handle even when it could have been okay to give out `Uri()`.

mcb30:
> > [...] make iPXE always just do the right thing by default, and not force users to make explicit choices [...]
> 
> We might be saying the same thing. I was in agreement with your heuristic, which almost always does the right thing by default, thereby indeed does not generally force users to make explicit choices. I was just asking to also have support for an _optional_ config parameter to handle rare edge cases.

Understood.  But optional parameters have a nasty habit of becoming de facto mandatory for handling known cases.

> Important aside: Is it even correct to use `Uri()` for chain-to-USB? I accept your claim that UEFI HTTPs boot [which I currently cannot test in my environment] passes in a handle ending in `Uri()`. But that type of handle might be _specifically_ and _solely_ intended for just HTTPs boot mode. If the user does UEFI USB boot or UEFI HD boot, isn't the device path something like `PciRoot() / Pci() / USB()|Sata() / HD() / Path(\EFI\Boot\...)` ? Thus, when iPXE chains to `file:...`, instead of making a child handle ending in `Uri()`, shouldn't it make a handle that identifies the drive device and ends in `Path()`?

Yes, we arguably should be using a file-based device path when executing something that we loaded from a filesystem.  This gets messy since the image stores only the originating URI, and the `file:` URI alone doesn't give us enough information to reconstruct a suitable UEFI device path.

> Here then is the proposed way for iPXE to process a `chain ...` command:
> 
>     * Set a legacy-mode flag to true.
> 
>     * _Load_ the next stage into memory (whether over network or from local filesystem). If it cannot be loaded that's an error.
> 
>     * Do a libmagic-style fingerprinting to determine whether the next stage is itself another iPXE (and of a sufficiently new vintage to understand `Uri()` or `Path()` leafs) — in this case, turn off the legacy-mode flag.

...this is around the point at which I reject this proposed design for having too much complexity.  If it's hard to describe, then it's likely to be messy to implement and very hard to document.  But thank you for investing your time and effort to come up with ideas and talk them through - I really do appreciate it, and the discussion helps move the design forward.

> But if you're saying you'd like to find a solution that _absolute always_ does the right thing automatically for _absolutely all_ situations, while _also_ satisfying the list of “ground truths” I stated, I believe my prior message is essentially a proof of impossibility.

I think I have a violation of the proof of impossibility.  :upside_down_face: 

In the UEFI world, the least surprising approach is along the lines of the heuristic that I described: provide a `Uri()` device path for an HTTP boot, or an `IPv4()`/`IPv6()` device path for a TFTP boot.  Deriving the device path type from the image's URI is therefore the starting point.

We already have an `imgargs` command that can be used to modify the argument list for an existing image.  This command ends up calling `image_set_cmdline()`.

It would be relatively easy to add an `imguri` command that can similarly be used to modify the URI for an existing image (by calling `image_set_uri()`).

Having an `imguri` command provides a very natural way to handle the exception cases that you describe, such as loading the WDS binary via HTTPS but then causing it to believe that it was loaded via TFTP:
```
imgload https://server.name/boot/wdsmgfw.efi
imguri wdsmgfw.efi tftp://${next-server}/wdsmgfw.efi
boot
```

So, I think the plan would be:
1. Add the `imguri` command.
2. Check EDK2 behaviour for the `Uri()` and `IPv4()`/`IPv6()` child nodes, and modify `efi_image.c` to replicate the EDK2 behaviour in terms of copying `EFI_SIMPLE_NETWORK_PROTOCOL` etc to the child handle.
3. Modify `efi_image.c` to construct a device path based on the image's URI.

Step 3 is non-trivial.  In particular, handling the IPv4/IPv6 decision will require some replumbing, since nothing in the URI necessarily indicates which should be used.  (Even TFTP URIs can use DNS names rather than literal IPv4/IPv6 addresses.)  We can't perform resolution of the hostname portion of the URI in `efi_image.c` since it would be possible for all network interfaces to be closed by that point.  It's not impossible, but it will require some careful design work.  Whatever design is chosen here should also allow for a clean way to handle `file:`-based URIs.

I believe that this `imguri` approach handles all of your listed situations, and does so by providing a small, generic, well-defined, and easily documented new feature that also provides a way to handle other corner cases (e.g. loading a payload via HTTPS and then pretending that you loaded it via HTTP, which could be necessary if the payload itself doesn't support HTTPS).

Thoughts?

NiKiZe:
suggestion, reuse `imgargs --uri` instead of the new command? (not needing separate documentation pages)
DHCPv6 does not use separate tftp and filename, instead it takes an URI, but many EFI implementations still only supports tftp:// protocol.
Also my tests have shown that EDK2 etc does not give enough information for autoexec.ipxe to work in v6, #1703 I wonder if there is anything available to solve this, or if it is an EDK2 bug that needs to be solved first. (and even worse for http boot?) I wonder how WDS behaves in a v6 start, but regardless, for v6 you might just want to go with Uri anyway (the fix that should have happened when moving to EFI but didn't happen :D)

mcb30:
> suggestion, reuse `imgargs --uri` instead of the new command? (not needing separate documentation pages)

Good point.  The side effect of that would be that it would not be possible to change just the URI without also providing the argument list.  I will let that sit for a while and think about whether or not it would be a problem.

> DHCPv6 does not use separate tftp and filename, instead it takes an URI, but many EFI implementations still only supports tftp:// protocol. Also my tests have shown that EDK2 etc does not give enough information for autoexec.ipxe to work in v6, #1703 I wonder if there is anything available to solve this, or if it is an EDK2 bug that needs to be solved first. (and even worse for http boot?) I wonder how WDS behaves in a v6 start, but regardless, for v6 you might just want to go with Uri anyway (the fix that should have happened when moving to EFI but didn't happen :D)

I'm pretty sure that IPv6 TFTP will provide an `IPv6()`-based device path, so we should probably match that.  Although I'm not sure without looking which packets iPXE will provide via `EFI_PXE_BASE_CODE_PROTOCOL` for IPv6.

alexanderperlis:
I like where this is going. I believe you are explicitly separating two things:

~~~
imgload ...  # how next stage is ACTUALLY loaded
imguri  ...  # how next stage will THINK it was loaded
~~~

This will provide a lot of flexibility. A missing `imguri` would presumably mean the same thing as if `imguri` had been specified with the *same* URI as was given to `imgload`. In other words, I believe you are proposing:
~~~
chain URI
~~~
is shorthand for
~~~
imgload URI
boot
~~~
which is shorthand for
~~~
imgload URI
imguri URI
boot
~~~

And if URI started with `tftp` in the above, then this is the *special case* that means “provide a PXE-like environment [snp handle, fake DHCP info]” to the next stage. It would be nice then if `imguri tftp:...` would automatically populate
~~~
netX/next-server
netX/filename
proxydhcp/next-server
proxydhcp/filename
~~~
(I can still follow `imguri` with `set` commands if I need to have non-identical `netX` and `proxydhcp` assignments.)

Otherwise what is the purpose/meaning of the `...` portion of `imguri tftp:...`?

> We can't perform resolution of the hostname [...] all network interfaces closed [...]

I imagine you're thinking of the case
~~~
imgload file:next-stage.efi
imguri tftp://some-dns-name/base/boot.0
~~~
specifically when iPXE has no difficulty loading the next stage from USB but has no network environment to resolve `some-dns-name` to an IP. Why not just error out with "Cannot resolve DNS name to an IP address"? Or do you mean a network environment exists but `imgload file:...` caused it to get closed prematurely? I don't understand enough to know why it needs to be closed at all (ever) or why that can't wait until `boot` determines which handle to pass along. (In the example at hand, the snp handle anyhow is the one that needs to be passed along.)

mcb30:
> This will provide a lot of flexibility. A missing `imguri` would presumably mean the same thing as if `imguri` had been specified with the _same_ URI as was given to `imgload`. In other words, I believe you are proposing:
> 
> ```
> chain URI
> ```
> 
> is shorthand for
> 
> ```
> imgload URI
> boot
> ```

Not really "proposing": that's already how `chain` works.

> And if URI started with `tftp` in the above, then this is the _special case_ that means “provide a PXE-like environment [snp handle, fake DHCP info]” to the next stage. 

Pretty close, yes.  In all cases we will provide an SNP handle and a PXE base code protocol (which provides the fake DHCP info).  The difference will be just in the device path that we provide.

> It would be nice then if `imguri tftp:...` would automatically populate
> 
> ```
> netX/next-server
> netX/filename
> proxydhcp/next-server
> proxydhcp/filename
> ```

It will not populate those automatically, since those are covered by the settings mechanism, and I don't want to conflate those mechanisms because it adds complexity and makes it hard to reason about what is supposed to happen.

> Otherwise what is the purpose/meaning of the `...` portion of `imguri tftp:...`?

Very little.  For some purposes, the whole URI provided to `imguri` would be meaningful.  For example, if the image is an iPXE script, then using `imguri` would change the base URI applied while the script is executing.  If it is a non-TFTP URI, then using `imguri` will affect the `Uri()` component of the device path.

For a TFTP URI, it will just iPXE to provide a "legacy" TFTP device path, in which the actual URI is communicated only indirectly via the DHCP packet contents (and uses the value defined via the settings mechanism, as was the case prior to commit https://github.com/ipxe/ipxe/commit/5a17d8daf).
 
> > We can't perform resolution of the hostname [...] all network interfaces closed [...]
> 
> I imagine you're thinking of the case
> 
> ```
> imgload file:next-stage.efi
> imguri tftp://some-dns-name/base/boot.0
> ```
> 
> specifically when iPXE has no difficulty loading the next stage from USB but has no network environment to resolve `some-dns-name` to an IP. Why not just error out with "Cannot resolve DNS name to an IP address"? Or do you mean a network environment exists but `imgload file:...` caused it to get closed prematurely? I don't understand enough to know why it needs to be closed at all (ever) or why that can't wait until `boot` determines which handle to pass along. (In the example at hand, the snp handle anyhow is the one that needs to be passed along.)

There is nothing in iPXE that currently requires network interfaces to be open at the point that an image is executed.  A script could choose to close them.  A script could choose never to have opened them (as in your example).  There may be no network interfaces present on the system.

We could probably add a restriction that performing a PXE-style boot (with an `IPv4`/`IPv6` device path) **requires** that the network is open at the point that the image is executed.  This technically adds a limitation that isn't currently there, but it's unlikely to affect anyone in practice.

With this in place, we could do something like:
* check if the image URI scheme is `tftp`
* if so, open the URI
* when name resolution completes (i.e. when we receive an `xfer_vredirect` event with `type == LOCATION_SOCKET`), then record the socket family (`AF_INET` or `AF_INET6`) and close the interface
* Use the socket family to decide which device path family to provide

Steinliiippp:
I just got stuck in setting up ipxe with WDS to do Linux-Boot and Windows-Deployment. I came across this problem and it seams to be exactly the problem I have. Are there already further plans to merge the patch and do a release (because of SecureBoot)?

mcb30:
Yes, fixing this is still in my queue.  No promises on a timeline at the moment, though,

PR #1725, opened 2026-05-28
[bnxt] Prevent out-of-bounds memory access

jw14812:
Add boundary checks to prevent out-of-bounds memory accesses in RX and HWRM paths.

### DIFF
--- a/src/drivers/net/bnxt/bnxt.c
+++ b/src/drivers/net/bnxt/bnxt.c
@@ -467,12 +467,15 @@ void bnxt_rx_process ( struct net_device *dev, struct bnxt *bp,
 		       struct rx_pkt_cmpl *rx_cmp,
 		       struct rx_pkt_cmpl_hi *rx_cmp_hi )
 {
-	u32 desc_idx = rx_cmp->opaque;
+	u32 desc_idx = rx_cmp->opaque & ( NUM_RX_BUFFERS - 1 );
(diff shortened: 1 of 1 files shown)


mcb30:
What is the motivation for these changes?

To keep code size small, iPXE doesn't generally attempt to defend against broken hardware.  The working assumption is that hardware with DMA access to host memory can already trash the machine if it wants to, and so there is very limited value in trying to defend against it.  We assume that network hardware will behave correctly in the same way that we assume the CPU will behave correctly.

We treat malicious hardware as out of scope for Secure Boot, since if an attacker has physical access to replace a PCI device, then it's already game over for Secure Boot protections.  (Commit https://github.com/ipxe/ipxe/commit/7b60a4875 was an exception to this, and was done just to placate the Secure Boot review process since it was the only "defect" they managed to identify in that round of review.)

The best approach for iPXE in terms of verifying expected hardware behaviour is generally to use assertions, since these will show up immediately during development or debugging and will have zero impact in production builds.

The exception would be if there are known (or suspected) hardware or firmware bugs that could plausibly cause the hardware to do the wrong thing.  Is this the case for bnxt?

jw14812:
Ack. Will withdraw this PR

jw14812:
@mcb30 - would this PR be acceptable if the assert() was restored back, but the other checks were retained?

PR #1735, opened 2026-06-15
[bnxt] Prevent out-of-bounds memory access

jw14812:
Add boundary checks to prevent out-of-bounds memory accesses in RX and HWRM paths.

### DIFF
--- a/src/drivers/net/bnxt/bnxt.c
+++ b/src/drivers/net/bnxt/bnxt.c
@@ -467,7 +467,7 @@ void bnxt_rx_process ( struct net_device *dev, struct bnxt *bp,
 		       struct rx_pkt_cmpl *rx_cmp,
 		       struct rx_pkt_cmpl_hi *rx_cmp_hi )
 {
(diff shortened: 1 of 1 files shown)


NiKiZe:
Is it worth mentioning in the commit message why this is needed?

mcb30:
Same comment as for #1725 - do we have a reason to expect the hardware to DMA invalid values into memory?

jw14812:
We don't have a specific reason to expect the hardware to DMA invalid values, but we are in the process of hardening the driver and this enforces standard software bounds checking before accessing host memory arrays. Validating resp_len against RESP_BUFFER_SIZE ensures we never overflow the allocated DMA buffer, which protects the driver if the buffer sizes change in future HWRM specifications. Similarly, masking rx_cmp->opaque explicitly enforces our power-of-two ring boundaries in software, ensuring safe array access even if we dynamically resize rings or pack additional metadata into the opaque field in the future.

mcb30:
Thank you for the explanation.  Rebasing and merging now.

PR #1743, opened 2026-06-22
[aqc1xx] Fix operation behind an IOMMU and RX buffer leak on close

animeshbm:
## Summary

Two fixes for the aqc1xx driver, found while bringing up aqc1xx
adapters on a host with the IOMMU (Intel VT-d) enabled.

## Changes

- **Set netdev->dma for operation with an IOMMU** — the driver never set
  `netdev->dma`, so transmit buffers were left unmapped for DMA. Without an
  IOMMU this happened to work (physical address == device address), but with
  an IOMMU the unmapped DMA access faults and stalls the receive path: no
  packets are received and DHCP fails. Setting `netdev->dma` and a 64-bit DMA
  mask maps the buffers through the firmware IOMMU, matching the other iPXE
  drivers.

- **Free outstanding receive I/O buffers on close** — `atl_close()` freed the
  descriptor rings but left the posted receive I/O buffers allocated, leaking
  them and tripping an assertion on the next open. These are now freed in
  `atl_close()`.

## Testing

Verified on AQC113 with VT-d enabled: the adapter now receives packets
and completes DHCP, and repeated open/close cycles no longer leak buffers or
assert.

### DIFF
--- a/src/drivers/net/marvell/aqc1xx.c
+++ b/src/drivers/net/marvell/aqc1xx.c
@@ -528,8 +528,10 @@ static int atl_probe ( struct pci_device *pci ) {
 		goto err_ioremap;
 	}
 
-	 /* Configure DMA */
+	/* Configure DMA */
 	nic->dma = &pci->dma;
+	dma_set_mask_64bit ( nic->dma );
+	netdev->dma = nic->dma;
 
 	/* Reset the NIC */
 	if ( ( rc = nic->hw_ops->reset ( nic ) ) != 0 )
--- a/src/drivers/net/marvell/aqc1xx.c
+++ b/src/drivers/net/marvell/aqc1xx.c
@@ -262,6 +262,7 @@ err_tx_alloc:
 */
 static void atl_close ( struct net_device *netdev ) {
 	struct atl_nic *nic = netdev->priv;
+	unsigned int i;
 
 	nic->hw_ops->stop ( nic );
 	/* rpb global ctrl */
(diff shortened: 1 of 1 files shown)


mcb30:
Is this intended to still be a draft PR, or would you like this reviewed?

animeshbm:
Hi @mcb30,

This was initially opened as a draft because we were still validating the fix. One of our customers reported this bug, and we have now implemented a fix and verified that it is working correctly in our testing.

The PR is ready for review now.

mcb30:
Thank you for these fixes.  They are very well structured and documented, and it is a pleasure to review them.  I am rebasing and merging now.

While reviewing, I noticed that the driver seems to set descriptor fields without forcing little-endianness (via `cpu_to_le64()` etc).  At the moment, all PCI systems that iPXE runs on are little-endian, but this may change in future.  (I just added partial support for the s390x architecture, which is big-endian.)

If you have the time to go through and ensure that all descriptor fields enforce the correct endianness, that would be amazing.

animeshbm:
@mcb30 Thanks for the review and merge.

Good catch on the endianness, and I agree: the descriptor field writes should go through `cpu_to_leXX()` so they stay correct on big-endian hosts. The read side already does the right thing (`le16_to_cpu()` / `le32_to_cpu()`), so it's really the write side that's inconsistent right now.

I'll send a follow-up patch making the TX/RX descriptor writes endianness-safe, bringing the write side in line with the already-correct read side. That should keep the driver correct on s390x and any future big-endian PCI platform.

Thanks :)

PR #1745, opened 2026-06-26
Change default keyboard map to dynamic

sfonteneau:
Hello,

Until now, I have been building iPXE myself mainly to change the KEYBOARD_MAP setting to dynamic.

With the new automated UEFI Secure Boot signing and release process, I would prefer to use the official signed builds instead of maintaining a custom build.

Would you consider changing the default KEYBOARD_MAP value to dynamic, or providing an official build option that enables it?

Thanks.

### DIFF
--- a/src/config/console.h
+++ b/src/config/console.h
@@ -83,6 +83,12 @@ FILE_SECBOOT ( PERMITTED );
 #define KEYBOARD_MAP	us	/* Default US keyboard map */
 //#define KEYBOARD_MAP	dynamic	/* Runtime selectable keyboard map */
 
+/* Use dynamic keyboard by default on EFI platforms */
+#if defined ( PLATFORM_efi )
+  #undef KEYBOARD_MAP
+  #define KEYBOARD_MAP	dynamic
+#endif
+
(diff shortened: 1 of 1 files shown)


NiKiZe:
If changed, it should only be done for efi builds, as is conditional for several other features as well.

sfonteneau:
Why not. What could cause a problem with non-UEFI systems?

NiKiZe:
I'm not going to reiterate on the size constraints a pcbios bootloader has.
And also on the basis that you can build them yourself if you need the feature, while as you say that is more difficult when you want signed binaries usable with secure Boot.

sfonteneau:
Would you like me to modify my pull request accordingly?

mcb30:
Don't worry about updating the PR: I'll treat this as being an issue report and fix it up myself.

It's definitely worth doing for Secure Boot builds.  For consistency, we should do it for all EFI builds by default.  I agree with @NiKiZe that it should probably be left as `us` by default for non-EFI builds.

sfonteneau:
Great, thanks a lot!

PR #1781, opened 2026-08-03
[golan] add native Ethernet support for mlx5Gen VF

fish-openai:
## Summary

Add a native Ethernet netdev path to iPXE's existing golan driver for the
Mellanox/NVIDIA 15b3:101e mlx5Gen Virtual Function exposed by Crusoe c1a.2x
VMs.

The existing golan probe recognizes Mellanox hardware but only constructs an
InfiniBand/IPoIB device. On this VF that leaves iPXE without a usable Ethernet
interface even though Linux drives the same PCI function as a normal mlx5e
netdev. This series adds the minimum mlx5e-style command, queue, steering,
TX, and RX support needed for iPXE to DHCP and chain an HTTP boot menu before
any guest OS runs.

## What changed

- Register PCI ID 15b3:101e and select a VF-specific Ethernet path.
- Negotiate the ISSI and HCA capability sequence accepted by the VF.
- Read the NIC vport MAC and register an Ethernet netdevice.
- Create the required CQ, RQ, SQ, TIS, TIR, transport-domain, queue-counter,
  and receive-flow-steering objects.
- Post receive WQEs, transmit ordinary Ethernet WQEs, process mlx5 CQEs, and
  deliver successful receive buffers to netdev_rx().
- Use EFI DMA mappings for the VF queue memory and Linux-compatible MKey/RQ
  protection-domain context.
- Preserve the existing golan behavior for every other PCI ID: the larger
  mailbox arena, EFI-DMA CQ allocation, and mlx5e MKey fields are gated to
  15b3:101e.

## Real-hardware result

Tested on a Crusoe c1a.2x VM whose frontend device is:

    15b3:101e Mellanox ConnectX mlx5Gen Virtual Function

The VM booted directly into the baked iPXE EFI, obtained DHCP through the
native VF path, fetched a remote iPXE menu over HTTP, and executed it:

    net0: 10.0.0.21/255.255.255.240 gw 10.0.0.17
    Chaining http://10.0.0.30:18080/boot for Crusoe
    http://10.0.0.30:18080/boot... ok
    REMOTE_BOOTSERVER_CHAIN_OK

This was a pre-OS boot: Ubuntu was used only once to place the EFI binary on
a disk before snapshotting it. The canary itself started in iPXE.

## Validation

- Built the complete series as x86_64 EFI with the cloud config, embedded
  autoexec script, and PARAM_CMD/IMDSV2_CMD flags used by the canary.
- Repeated real Crusoe canaries while bringing up capability negotiation,
  queue creation, TX CQEs, RX CQEs, DHCP, and HTTP chaining.
- Re-applied the complete series to a clean upstream checkout and compiled the
  final gated tree.
- Audited the final compatibility patch: non-101e golan devices retain the
  original mailbox size, malloc_phys CQ allocation, bus-address encoding, and
  MKey context.

## Notes for review

This is intentionally a draft. The implementation is proven on the Crusoe
15b3:101e VF, but the command layouts are still minimal and several diagnostic
messages remain useful while upstream reviewers decide how much of the path
should be generalized into a reusable mlx5 Ethernet layer.

The series is kept as separate commits matching the bring-up steps so each
hardware finding can be reviewed or squashed independently.


NiKiZe:
https://github.com/ipxe/ipxe/blob/master/CONTRIBUTING.md#abuse-of-ai

Even if the pr in full isn't that bad, 77 commits to review is unacceptable, this follows the pattern that LLMs to often waste human time and energy.

A human needs to fix and rebase this PR.

NiKiZe:
Reading the patch names, there is a lot of "to match Linux" driver development is done from data sheets, not based on Linux kernel or other existing source code.

fish-openai:
Added follow-up commit `b313f277e` on `fish-openai:dev/fish/mlx5e-vf` to fix the Crusoe SQ software-ring geometry.

`crusoe_create_sq()` already programs `log_wq_sz=10`, i.e. 1024 WQEBBs backed by the existing 16 pages, but `GOLAN_CRUSOE_SQ_WQES` was still 64. The producer and CQE-consumer masks therefore recycled software slots after 64 entries while firmware could continue using the larger queue. The commit changes the software ring to 1024 and adds an adjacent comment tying it to the programmed SQ size.

This is the same narrowly scoped correction folded into fleet-infra's consolidated `0007-crusoe-mlx5e-vf.patch`; it is driver correctness, not part of the later FOU/GRE work.

mcb30:
@fish-openai This PR is absurd in its current form.  Please restructure it sensibly.

Issue #381, opened 2021-06-08
Freeze: iPXE initialising devices...


nathanshearer:
When booting iPXE and it always freezes at "iPXE initialising devices" for several minutes then the system reboots.

This bug is reproducible with the latest master copy of iPXE and with iPXE 1.21.1 on a MacBook Pro 9,1.
This does not happen with iPXE 1.20.1.

I would normally just downgrade iPXE and continue, however 1.20.1 has other bugs that are fixed in the newer versions.

nathanshearer:
How can I enable debugging for everything when I build iPXE?

The documentation indicates I need to specify each item that should have debugging enabled, however I do not know which component is freezing. It may be verbose, but at least I'll get some kind of output that may narrow down the problem.

NiKiZe:
Yes you need to specify every item, enabling everything does not make sense.

Since you have known working commit and a non working commit the best thing is probably git bisect https://ipxe.org/howto/bisect

nathanshearer:
I have started the bisect, however older versions of iPXE cannot be built:

```
make bin-x86_64-efi/ipxe.efi
...
  [BUILD] bin-x86_64-efi/flexboot_nodnic.o
drivers/infiniband/flexboot_nodnic.c: In function ‘flexboot_nodnic_create_qp’:
drivers/infiniband/flexboot_nodnic.c:368:53: error: implicit conversion from ‘enum ib_queue_pair_type’ to ‘nodnic_queue_pair_type’ [-Werror=enum-conversion]
  368 |  status = nodnic_port_create_qp(&port->port_priv, qp->type,
      |                                                   ~~^~~~~~
drivers/infiniband/flexboot_nodnic.c: In function ‘flexboot_nodnic_destroy_qp’:
drivers/infiniband/flexboot_nodnic.c:409:45: error: implicit conversion from ‘enum ib_queue_pair_type’ to ‘nodnic_queue_pair_type’ [-Werror=enum-conversion]
  409 |  nodnic_port_destroy_qp(&port->port_priv, qp->type,
      |                                           ~~^~~~~~
cc1: all warnings being treated as errors
make: *** [Makefile.housekeeping:946: bin-x86_64-efi/flexboot_nodnic.o] Error 1
```

mcb30:
You can force older versions to build by using `make bin-x86_64-efi/ipxe.efi NO_WERROR=1`

stappersg:
Recently on _IRC_
```text
[...]

PR #1815, opened 2026-08-18
[efi] Support Apple EFI: TPL hang workaround, HII-less font fallback, high-DPI console scaling

lonlundgren:
Apple EFI (2012–2013 Intel Macs) hangs hard in current iPXE: once startup
completes, `bs->RestoreTPL ( TPL_APPLICATION )` never returns on this
firmware, freezing the machine at the first timed prompt with a dead
keyboard (issue #381 reports the same family on a MacBookPro9,1; we
reproduced and root-caused on a MacBookPro10,1 with on-screen marker
instrumentation — timer events create, arm and fire correctly, and console
CheckEvent is never reached; the wedge is the drop itself).

Three commits:

1. `[efi]` Record the external TPL as at most `TPL_APPLICATION`, floor all
   internal TPL drops at `TPL_CALLBACK` (the fatal call is never made), and
   move the timer tick event to `TPL_NOTIFY` so the clock advances while
   running at `TPL_CALLBACK`. Consequence: the tick-jitter entropy source is
   unusable under the floor, so TLS needs another entropy source (RDRAND
   covers every 2012+ Ivy Bridge Mac).

2. `[efifb]` Apple EFI provides GOP but no HII font protocol, so the
   framebuffer console cannot start on these machines. Fall back to a
   built-in 8x16 font (Linux `lib/fonts/font_8x16.c`, GPL-2.0) for
   single-byte characters, consulting HII only for glyphs beyond that
   range. Behavior is unchanged where HII exists.

3. `[fbcon]` These machines expose only the panel's native GOP mode; at
   2880x1800 an 8x16 font is unreadable. Draw each font pixel as an NxN
   block chosen from the actual mode (largest N keeping >=100 columns,
   demoted under 25 rows). Panels narrower than 1800px always get N=1,
   bit-identical to the previous rendering (verified by QEMU screendump
   comparison).

Testing: dozens of boots on a MacBookPro10,1 (USB and GRUB-chainloaded from
the ESP), including full Ubuntu 24.04 netboot installs over TLS
(RDRAND-backed); QEMU/OVMF regression runs show identical rendering and
flow to unpatched behavior at ordinary resolutions.

Addresses #381.


NiKiZe:
https://github.com/ipxe/ipxe/blob/master/CONTRIBUTING.md#major-feature-pull-requests

Where is the font from and what is the copyright?
> Glyph data is the classic VGA 8x16 font as distributed in the Linux
 kernel (lib/fonts/font_8x16.c, GPL-2.0, generated by cpi2fnt).

mcb30:
The question that needs to be answered before even considering this approach is: why does restoring to `TPL_APPLICATION` fail? I'm not merging a complicated workaround that has no justification for its own necessity.

It's impossible that `RestoreTPL` is just completely broken on this firmware since it's such a foundational boot services call. 

Start by finding out why it fails. The best way to do this will be to find the earliest point in iPXE's execution at which restoring to `TPL_APPLICATION` stops working.

lonlundgren:
> Where is the font from and what is the copyright?

The glyph table is generated from the Linux kernel's `lib/fonts/font_8x16.c` (GPL-2.0) — the classic IBM VGA 8×16 face every kernel console ships. If GPL-2.0-only versus UBDL relicensing is the concern, the table regenerates mechanically from any public-domain 8×16 bitmap source.

@mcb30 — I'd like the underlying answer too. What marker instrumentation established on this firmware (MacBookPro10,1, no source access): timer events create, arm, and fire normally; once startup completes, `RestoreTPL ( TPL_APPLICATION )` never returns; floored at `TPL_CALLBACK`, everything runs. From outside I can't distinguish RestoreTPL itself wedging from a queued notification handler wedging during the dispatch it performs — the observable is identical, and probing for the earliest failing drop is an instrumented-boot campaign on physical Apple hardware I can't commit to. If a ~6-line workaround without the underlying mechanism doesn't meet the bar, understood — the branch remains available for anyone stranded by #381.

Issue #1231, opened 2024-06-02
Feature request: substring syntax


alkisg:
In bash, the `${var:start:count}` syntax allows extracting a substring. It would be useful to have something similar in ipxe, e.g. to apply different settings based on a hostname or mac prefix.

In the meantime, here's a trick to extract the vendor (3 bytes) from the mac address:

> set x ${mac:base64} && set x:ipv4 {x:ipv4} && set x:base64 ${x:string} && set x:hex ${x:hex} && show x

And here's a similar trick to extract the first four characters of the hostname, padded with dashes if it's smaller:

> set x ${hostname}____ && set x:ipv4 ${x:ipv4} && set x ${x:string} && show x

Of course feel free to close the issue if it's not deemed important, compared to the expected binary size increase! 😊

NiKiZe:
Unfortunately... closing as duplicate of #1061 https://github.com/ipxe/ipxe/discussions/326#discussioncomment-629914

Issue #1061, opened 2023-10-20
String manipulation methods


kingsumos:
Hi,

Looks like we don't have any possibility for string manipulation inside iPXE scripts. I found this pull request which was rejected for some reason:
https://github.com/ipxe/ipxe/pull/95
The proposal above will be very helpful, e.g.:
show smbios/product:string:0:8
show smbios/uuid::0:24
echo ${ifname::3:4}

Other user looking for this feature:
https://forum.ipxe.org/showthread.php?tid=18573&pid=36076
(tried adding a post in forum 1st, but new users aren't allowed for registration)

Another string manipulation attempts:
https://forum.ipxe.org/showthread.php?tid=7979
https://forum.ipxe.org/showthread.php?tid=18573
https://forum.ipxe.org/showthread.php?tid=18559

kingsumos:
It would be nice if iPXE can mimic bash-scripting parameter substitution functionality, reference link:
https://tldp.org/LDP/abs/html/parameter-substitution.html

NiKiZe:
String manipulations will not be added to iPXE, even if there is good usecases you are almost always better of sending strings to a server and let more competent languages handle these things.

kingsumos:
Ok no problem Mr. Lennart Poettering
I'm indeed very competent, good luck with your project

NiKiZe:
Better description then mine https://github.com/ipxe/ipxe/discussions/326#discussioncomment-629914

PR #1835, opened 2026-09-04
[cmdline] Add numeric comparison commands "numlt", "numeq" and "numgt"

cbamelis:
The "iseq" command compares strings for equality, but there is no way to compare values numerically.  A script therefore cannot branch on numeric quantities such as ${memsize}, e.g. to guard a menu entry on the amount of available memory:

    #!ipxe
    isge ${memsize} 4096 && item bigimage Large image (needs 4GB)

Add "islt", "isgt", "isle" and "isge" as the numeric counterparts to the string comparison performed by "iseq", covering "<", ">", "<=" and ">=" respectively.  Both arguments are parsed with parse_integer(), so a non-numeric argument is reported as an error rather than being silently treated as zero.

### DIFF
--- a/src/core/exec.c
+++ b/src/core/exec.c
@@ -538,6 +538,80 @@ static int iseq_exec ( int argc, char **argv ) {
 /** "iseq" command */
 COMMAND ( iseq, iseq_exec );
 
+/** First value is less than second value */
+#define NUM_LT 0x01
+
+/** Values are equal */
+#define NUM_EQ 0x02
+
+/** First value is greater than second value */
+#define NUM_GT 0x04
+
+/** Numeric comparison options */
+struct num_options {};
+
+/** Numeric comparison option list */
+static struct option_descriptor num_opts[] = {};
+
+/** Numeric comparison command descriptor */
+static struct command_descriptor num_cmd =
+	COMMAND_DESC ( struct num_options, num_opts, 2, 2,
+		       "<value1> <value2>" );
+
+/**
+ * Numeric comparison commands
+ *
+ * @v argc		Argument count
+ * @v argv		Argument list
+ * @v permitted		Permitted comparison outcomes
+ * @ret rc		Return status code
+ */
+static int num_exec ( int argc, char **argv, unsigned int permitted ) {
+	struct num_options opts;
+	unsigned int first;
+	unsigned int second;
+	unsigned int outcome;
+	int rc;
+
+	/* Parse options */
+	if ( ( rc = parse_options ( argc, argv, &num_cmd, &opts ) ) != 0 )
+		return rc;
+
+	/* Parse values */
+	if ( ( rc = parse_integer ( argv[optind], &first ) ) != 0 )
+		return rc;
+	if ( ( rc = parse_integer ( argv[ optind + 1 ], &second ) ) != 0 )
+		return rc;
+
+	/* Compare values */
+	outcome = ( ( first < second ) ? NUM_LT :
+		    ( ( first > second ) ? NUM_GT : NUM_EQ ) );
+
+	/* Return success iff outcome is permitted */
+	return ( ( outcome & permitted ) ? 0 : -ERANGE );
+}
+
+/** "numlt" command */
+static int numlt_exec ( int argc, char **argv ) {
+	return num_exec ( argc, argv, NUM_LT );
+}
+
+COMMAND ( numlt, numlt_exec );
+
+/** "numeq" command */
+static int numeq_exec ( int argc, char **argv ) {
+	return num_exec ( argc, argv, NUM_EQ );
+}
+
+COMMAND ( numeq, numeq_exec );
+
+/** "numgt" command */
+static int numgt_exec ( int argc, char **argv ) {
+	return num_exec ( argc, argv, NUM_GT );
+}
+
+COMMAND ( numgt, numgt_exec );
+
 /** "sleep" options */
 struct sleep_options {};
 
--- a/src/core/exec.c
+++ b/src/core/exec.c
@@ -530,7 +530,7 @@ static int iseq_exec ( int argc, char **argv ) {
 	if ( ( rc = parse_options ( argc, argv, &iseq_cmd, &opts ) ) != 0 )
 		return rc;
 
-	/* Return success iff arguments are equal */
+	/* Return success iff string values are equal */
 	return ( ( strcmp ( argv[optind], argv[ optind + 1 ] ) == 0 ) ?
 		 0 : -ERANGE );
 }


NiKiZe:
This is indeed nice to have features, but can it be solved in other ways?
Similarities to https://github.com/ipxe/ipxe/discussions/326#discussioncomment-629914 #95 #1061 #1231 #1330

The use-case described sounds like something that cane be done by sending to http server and let a server script make the decision?

If we consider the code, would it make sense code size wise to rely on the int setting types, and that theses functions can only take setting names, and those settings must be of identical int type, that reduces parsing etc. (ignoring the usability aspect for a moment)

mcb30:
I'm not against this in principle (though I would probably adjust the code a bit).  The issue for me is that the existence of `islt`, `isgt`, etc sets up an incorrect expectation that `iseq` will also compare numeric values.

I think any commands for numeric comparison need to exist with names that do not visibly live alongside `iseq`, in order to avoid creating this false expectation.

cbamelis:
I have read the linked discussions and reworked the patch. My friend Claude Code helped: it found the existing patterns in the tree that the code now follows, made the measurements below, and tested the commands and the derived relations in qemu. The design choices are mine.

@mcb30 The set now has an explicit `numeq`, which is aimed at exactly that expectation. The three commands are `numlt`, `numeq` and `numgt`. No name has an `is` prefix, and the numeric equality is now a command of its own: `numeq 0x10 16` succeeds, and `iseq 0x10 16` fails.

A separate commit addresses the same expectation in the code: the comment on `iseq` now says that it compares string values.

Three commands are a complete set, because a numeric comparison has three possible outcomes. The other relations are an `||` of two positive tests:

| relation | script |
| --- | --- |
| a `<=` b | `numlt a b \|\| numeq a b` |
| a `>=` b | `numgt a b \|\| numeq a b` |
| a `!=` b | `numlt a b \|\| numgt a b` |

Both sides of the `||` are a positive test, so a later `&&` still works as expected:

```
numgt ${memsize} 4096 || numeq ${memsize} 4096 && cpuid --ext 29 && item win11 Windows 11 (x64)
```

On the code: the three commands share one comparison body, in the same shape as `digest_exec()` in `hci/commands/digest_cmd.c` with its per-algorithm wrappers in `crypto/mishmash/cmd_*.c`. `num_exec()` parses the two values, maps them to one of three outcomes, and returns success when that outcome is in the bitmask that the command permits. Each command is then a wrapper of one line.

@NiKiZe On the other threads: #95, #1061, #1231 and #1330 all ask for string manipulation. Discussion #326 states the general principle, that the script language stays small and that this kind of processing belongs on the server. These commands stay inside that principle: they do no arithmetic and they build no strings. They compare two values and return a status.

A server can indeed make the decision, and for anything complex it should. That is not always possible: the server must run scripts, and the client-only values (`${memsize}`, `${cpuid/...}`) must travel there as URL parameters first. An embedded script has no server at all until it configures the network, so its flow control and conditional execution have to work locally.

On the idea to accept only integer settings: the commands add no parsing code as they are, because `parse_integer()` is in `parseopt.c` and every build links that file already. A setting-only form needs more code, not less, because it must find the setting by name and read its value as an integer. It also loses literal values: `numgt ${memsize} 4095` would then need a setting that holds 4095.

On size: for `bin/ipxe.pxe` on i386 the uncompressed `.text` grows by 184 bytes, from 972441 to 972625. This is less than the four commands in the earlier form. In a ROM image the compressed payload grows by about 110 bytes. ROM images are padded to 512-byte blocks, so a given ROM either keeps its size or gains one block.

So the open question is whether that is acceptable in a default build. Every build already has the string comparison, and every build already exposes `${memsize}`, `${cpuid/...}` and the other numbers as settings. iPXE therefore produces numbers that its own scripts cannot order, and these 184 bytes close that gap. Whether that is worth the size, or whether the commands belong behind a conditional `NUM_CMD` option like `PING_CMD` and `NTP_CMD`, is your decision. I can move them into `hci/commands/num_cmd.c` if you prefer the option.

cbamelis:
Correction to my previous comment. I wrote that every build exposes `${memsize}` and `${cpuid/...}` as settings. That is wrong: `MEMMAP_SETTINGS` is commented out in `config/settings.h`, and `CPUID_SETTINGS` is undefined again for `PLATFORM_pcbios`, so neither setting exists in the default i386 build that I measured.

Every build does have `${unixtime}` as a built-in setting, and a counter through `set attempt:int32 0` plus `inc`, because `NVO_CMD` is on by default. Both hold numbers that a script can currently only test for equality.

The case I should have given needs no settings source at all:

```
set attempt:int32 0
:retry
inc attempt
numlt ${attempt} 4 && goto quick ||
numlt ${attempt} 12 && goto patient ||
goto exhausted
```

An equality test cannot express those tiers, and the ordering form cannot overshoot the limit.

Whether that is worth 184 bytes in a default build remains your decision.

cbamelis:
I removed my two earlier comments in this thread: they stated that a default build exposes `${memsize}` and `${cpuid/...}` as settings, which is wrong. This comment replaces them, and I have edited it since to fix further mistakes of my own. Your notification mail holds the original wording. Sorry for the noise — I should have checked those claims before I posted them.

My friend Claude Code helped with this patch: it found the existing patterns in the tree that the code follows, made the measurements below, and tested the commands in qemu. The design choices are mine.

@mcb30 The commands are now `numlt`, `numeq` and `numgt`. The set has an explicit `numeq`, so numeric equality is a command of its own and no name carries the `is` prefix: `numeq 0x10 16` succeeds, and `iseq 0x10 16` fails. A separate commit says the same in the code, where the comment on `iseq` now states that it compares string values.

They are for a boot menu that offers different images by RAM size and CPU features. That needs `MEMMAP_SETTINGS` and `CPUID_SETTINGS`, and I enable both in my own builds. `iseq` is enough for a counter, because `inc` steps by one and therefore meets every boundary value. It cannot test an arbitrary quantity: a memory size or a timestamp never steps through the values, so there is nothing to match it against.

Three commands are a complete set, because a numeric comparison has three possible outcomes. The other relations are an `||` of two positive tests, so a later `&&` still works as expected:

| relation | script |
| --- | --- |
| a `<=` b | `numlt a b \|\| numeq a b` |
| a `>=` b | `numgt a b \|\| numeq a b` |
| a `!=` b | `numlt a b \|\| numgt a b` |

The three commands share one comparison body, in the same shape as `digest_exec()` in `hci/commands/digest_cmd.c` with its per-algorithm wrappers in `crypto/mishmash/cmd_*.c`. `num_exec()` parses the two values, maps them to one of three outcomes, and returns success when that outcome is in the bitmask that the command permits. Each command is then a wrapper of one line.

@NiKiZe #95, #1061, #1231 and #1330 all ask for string manipulation, and discussion #326 states the general principle. These commands stay inside it: they do no arithmetic and they build no strings. They compare two values and return a status. A server can make the decision instead, and for anything complex it should, but that needs a server which runs scripts, and it cannot help before the network exists.

On accepting only integer settings: the commands add no parsing code as they are, because `parse_integer()` is in `parseopt.c` and every build links that file already. A setting-only form must look the setting up by name and read it as an integer, which is more code, and a literal threshold would then need a setting of its own.

For `bin/ipxe.pxe` on i386 with the default configuration, the uncompressed `.text` grows by 184 bytes, from 972441 to 972625. In a ROM image the compressed payload grows by about 110 bytes. ROM images are padded to 512-byte blocks, so a given ROM either keeps its size or gains one block.

Every build has `${unixtime}` as a built-in setting, and that is such a quantity: there is no equality equivalent for `numlt ${unixtime} 1735689600`, and `${unixtime}` renders as hex on top of that. The commands therefore fill a gap in flow control that `iseq` leaves open, in any build. Whether that gap is worth 184 bytes in a default build, or whether it belongs behind a `NUM_CMD` option like `PING_CMD` and `NTP_CMD`, is your decision. Both outcomes are fine by me, and I will move the commands into `hci/commands/num_cmd.c` if you prefer the option.

